PR #7, opened 2020-12-01
Adapt to use taskwrapper

gconesab:


MC/run/PWGGAJE/run_dirgamma.sh
```diff
@@ -1,39 +1,86 @@
...
+[ ! -d output ] && mkdir output
+
+copypersistentsimfiles() {
+  simprefix=$1
+  outputdir=$2
+  cp ${simprefix}_Kine.root ${simprefix}_grp.root ${simprefix}*.ini ${outputdir}
+}
+
+# Loop over timeframes
+for tf in `seq 1 ${NTIMEFRAMES}`; do
```
sawenzel, line 51: Hi @gconesab : I think I had in mind a much more minor change: just prepend the elemental stages (o2-sim) by the "taskwrapper" keyword. My intention was that you get familiar with it's features and provide possible feedback.

You went further and made this into full embedding examples. Indeed, it in the end this repetitive code should be handled by a mother sim workflow.

I have no hard feelings about this. We can do the changes for now or you may keep it more simple. What do you think?

gconesab: OK, then let's keep it simple, also the timeframe loop should go out right?

sawenzel: yep

sawenzel (approved):
Looks good to me

PR #8, opened 2020-12-08
Kr-Kr config file

mmazzilli:
Adding Kr-Kr config file


MC/config/common/pythia8/generator/pythia8_KrKr.cfg
```diff
@@ -0,0 +1,16 @@
+### beams
+Beams:idA 1000822080		# Pb
+Beams:idB 1000822080		# Pb
+1000360840:all 84Kr 84Krbar 10 108 0 83.798
+Beams:eCM 6460. 		# GeV
+
+### heavy-ion settings (valid for Pb-Pb 5520 only)
```
preghenella, line 7: I believe this comment should be modified to `(valid for Kr-Kr 6460 only`)

PR #12, opened 2021-01-13
adapt to change in interface in O2

sawenzel:


sawenzel:
@preghenella : This is necessary after your recent change. Maybe we should think about how to avoid going out of sync again. Maybe we should put tests in the full CI. Can you merge a prepare a new tag for the GRID?

preghenella:
hi @sawenzel thanks, my bad.
I will merge and tag.

PR #27, opened 2021-02-25
First commit of Pythia8 gun event generator

mfaggin:
Event generator based on Pythia8 that injects only the desired particles in the event, which can be used with o2-sim.
Files:
1) generator_pythia8_gun.C: it contains the class with the generator  and the custom functions to be called from the *.ini file;
2) config_custom_OmegaC.cfg: custom settings for OmegaC generations.
    In this version: decay forced to Omega-pion; ctau set to 80 micron;
3) configOmegaAnd20Pions_randomCharge.ini: example of config file used to call the gun.
    The function produces events containing a Omega (with random sign) plus 10 pi+ and 10 pi-.
    The function is defined inside the file in point 1).


preghenella:
Thanks @mfaggin 
I merge it as is.
I will look into it, we might change the location of the files as per needs in the future.

PR #33, opened 2021-03-03
Update GH action for new release convention

TimoWilken:
This allows an optional trailing letter on the version, for multiple releases in a day.


sawenzel (approved):
thanks

PR #34, opened 2021-03-03
Introduce demonstrator for background caches in embedding

sawenzel:
This is demonstrating how we can incorporate background caches
into the MC-pipeline.

We can use a sophistication in which we copy background files
in multiple stages, since not all files are needed at the same time.
This allows for hiding the copy latency and we can overlay the copy
of large hit files with signal transport simulation etc.
Moreover, we copy only what is needed by a workflow.

On a bookkeeping level we still need to verify compatibility
of background files with the curent software stack etc.

### DIFF
--- a/MC/bin/o2_dpg_workflow_runner.py
+++ b/MC/bin/o2_dpg_workflow_runner.py
@@ -20,6 +20,7 @@ except ImportError:
 
 formatter = logging.Formatter('%(asctime)s %(levelname)s %(message)s')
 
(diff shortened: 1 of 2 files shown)


sawenzel:
@preghenella , @gconesab.

Let me know your comments. I will incorporate these changes into the main script once I have benchmark numbers.

gconesab:
Hi Sandro,
the idea of the implementation is interesting, I cannot follow the full logic of the changes, but I think it is fine.

PR #39, opened 2021-03-29
Fix Pb PDG and automatic energy setting for Pb and HI pythia configuration

gconesab:
Also, add argument for particle detector acceptance filter.

### DIFF
--- a/MC/config/common/pythia8/utils/mkpy8cfg.py
+++ b/MC/config/common/pythia8/utils/mkpy8cfg.py
@@ -94,6 +94,14 @@ if args.process == 'dirgamma':
     fout.write('PromptPhoton:all = on \n')
 fout.write('\n')
 
+### heavy ion  settings (valid for Pb-Pb 5520 only)
+if args.idA==1000822080 and args.idB==1000822080:
+    fout.write('### heavy-ion settings (valid for Pb-Pb 5520 only) \n')
+    fout.write('HeavyIon:SigFitNGen = 0 \n')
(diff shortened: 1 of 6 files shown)


preghenella:
@gconesab careful with p-Pb and Pb-p settings.
you cannot have a CMS energy there because of asymmetric beam energies.

best would be to redefine everything in terms of the beam energy rather than the CMS energy

MC/config/common/pythia8/utils/mkpy8cfg.py
```diff
@@ -12,13 +12,19 @@
...
-                   help='PDG code of projectile beam')
+                   help='PDG code of projectile beam A')
 
 parser.add_argument('--idB', type=int, default='2212',
-                   help='PDG code of target beam')
+                   help='PDG code of target beam B')
 
-parser.add_argument('--eCM', type=float, default='13000.',
-                    help='Centre-of-mass energy')
+parser.add_argument('--eBeamA', type=float, default='6499.',
```
preghenella, line 20: I suggest you keep the same name as the one used by Pythia, namely `--eA` and `--eB`

MC/config/common/pythia8/utils/mkpy8cfg.py
```diff
@@ -12,13 +12,19 @@
...
+                    help='Energy of beam A')
+
+parser.add_argument('--eBeamB', type=float, default='6499.',
```
preghenella, line 23: same here

gconesab:
I will merge this today unless I get more comments.

PR #41, opened 2021-03-31
Adapt to changed options

sawenzel:


sawenzel:
@gconesab : Is this the new way? Can you merge if ok? (Please use "Rebase and Merge" instead of "Create a merge commit").

gconesab:
Yes, it is the new way, same as PYTHIA

PR #42, opened 2021-04-01
Some changes on workflow generation

sawenzel:
... Mainly geared to being able to process 100PbPb within 16GB.

 Changes to workflow generator
    
    * Split TPC clusterization into 2 stages
      so that we can fit into 16GB for ~100PbPb.
    * Introduce additional cluster merging step
      to prepare one file for TPC reco
    * Add ROOT macro helper for the merge step
      (to be generalized later)
    * Adjust some memory/cpu resource estimates
      (based on ~100PbPb dataframes)
    * Don't use --rate 1 for TOF reco workflow.
      Now terminates much faster.


gconesab:
Fine with me!

PR #43, opened 2021-04-06
TRD TRAP simulation is added to digitization

martenole:
So it does not need to be done separately. This should be merged after [PR5860](https://github.com/AliceO2Group/AliceO2/pull/5860) is merged to O2.

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -425,11 +425,6 @@ for tf in range(1, NTIMEFRAMES + 1):
    ITSTPCMATCHtask['cmd']= 'o2-tpcits-match-workflow ' + getDPL_global_options(bigshm=True, nosmallrate=True) + ' --tpc-track-reader \"tpctracks.root\" --tpc-native-cluster-reader \"--infile tpc-native-clusters.root\"'
    workflow['stages'].append(ITSTPCMATCHtask)
 
-   # this can be combined with TRD digitization if benefical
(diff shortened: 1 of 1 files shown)


davidrohr:
AliceO2Group/AliceO2#5860 is merged, so you can go ahead with this one.

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -425,11 +425,6 @@ def createRestDigiTask(name, det='ALLSMALLER'):
    ITSTPCMATCHtask['cmd']= 'o2-tpcits-match-workflow ' + getDPL_global_options(bigshm=True, nosmallrate=True) + ' --tpc-track-reader \"tpctracks.root\" --tpc-native-cluster-reader \"--infile tpc-native-clusters.root\"'
    workflow['stages'].append(ITSTPCMATCHtask)
 
-   # this can be combined with TRD digitization if benefical
-   TRDTRAPtask = createTask(name='trdtrap_'+str(tf), needs=[TRDDigitask['name']], tf=tf, cwd=timeframeworkdir, lab=["DIGI"])
-   TRDTRAPtask['cmd'] = 'o2-trd-trap-sim'
-   workflow['stages'].append(TRDTRAPtask)
```
sawenzel, line 431: For the record, this change is not enough since it's deleting a "named" graph node which is referenced elsewhere.
I will adapt this PR and merge. Thanks in any case!

martenole:
Sorry about that I should have tried to actually execute the script and then I would have seen it immediately.. Thanks for the fix!

PR #47, opened 2021-04-16
Slightly higher DPL --rate settings

sawenzel:
--rate 1 has exit/termination problems.

With a slightly higher rate (1000) this seems to work much better
while still avoiding busy CPU spinning problems in source
devices.


sawenzel:
New setting tested extensively on alibi. Merging to see if it improves lxplus/GRID problems.

PR #51, opened 2021-04-22
add option to pass interaction rate as parameter and set different de…

gconesab:
…fault parameters depending the collision type


sawenzel (approved):
Looks good. Thanks.

PR #62, opened 2021-06-04
add protections and exit in case of missing parameters, remove defaul…

gconesab:
…t value of generator

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -25,7 +25,7 @@ import array as arr
 parser = argparse.ArgumentParser(description='Create an ALICE (Run3) MC simulation workflow')
 
 parser.add_argument('-ns',help='number of signal events / timeframe', default=20)
-parser.add_argument('-gen',help='generator: pythia8, extgen', default='pythia8')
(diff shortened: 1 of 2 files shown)


gconesab:
Hi, @sawenzel and @preghenella, let me know what you think

sawenzel:
Ok with me. But can you test it with https://github.com/AliceO2Group/O2DPG/blob/master/MC/run/PWGHF/embedding_benchmark.sh please? Probably some scripts need to be adjusted.

gconesab:
Hi @sawenzel, I modified and did a quick test to your benchmark script, have a look. If ok we can merge.
G

gconesab:
Hi @sandro, I see there is a conflict but I have no idea, I think I did rebase before modifying. Do you understand what is wrong?

gconesab:
> /cc @sawenzel

ups, sorry!

sawenzel:
@gconesab : I needed to update your PR since I did some mods at the same time. You should be able to squash-merge without a problem.

gconesab:
> @gconesab : I needed to update your PR since I did some mods at the same time. You should be able to squash-merge without a problem.

It does not allow me unless I resolve the conflict.

gconesab:
@sawenzel, can you check I did not do something wrong?

PR #65, opened 2021-06-14
Scale number of workers

benedikt-voelkel:
* scale number workers in case only a fraction of workers should be used
  used for certain tasks

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -49,7 +49,7 @@ parser.add_argument('-colBkg',help='embedding background collision system', defa
 
 parser.add_argument('-e',help='simengine', default='TGeant4')
 parser.add_argument('-tf',help='number of timeframes', default=2)
-parser.add_argument('-j',help='number of workers (if applicable)', default=8)
+parser.add_argument('-j',help='number of workers (if applicable)', default=8, type=int)
 parser.add_argument('-mod',help='Active modules', default='--skipModules ZDC')
 parser.add_argument('-seed',help='random seed number', default=0)
 parser.add_argument('-o',help='output workflow file', default='workflow.json')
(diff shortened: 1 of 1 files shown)


MC/bin/o2dpg_sim_workflow.py
```diff
@@ -89,8 +89,17 @@
 workflow={}
 workflow['stages'] = []
 
+def relativeCPU(n_rel, n_workers=NWORKERS):
+    # compute number of CPUs from a given number of workers
+    # n_workers and a fraction n_rel
+    # catch cases where n_rel > 1
+    return min(n_workers, n_workers * n_rel)
```
sawenzel, line 96: will this give a float (better) or an int?

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -313,7 +322,7 @@ def getDPL_global_options(bigshm=False,nosmallrate=False):
             signalneeds = signalneeds + [ BKGtask['name'] ]
        else:
             signalneeds = signalneeds + [ BKG_HEADER_task['name'] ]
-   SGNtask=createTask(name='sgnsim_'+str(tf), needs=signalneeds, tf=tf, cwd='tf'+str(tf), lab=["GEANT"], cpu='5.', mem='2000')
+   SGNtask=createTask(name='sgnsim_'+str(tf), needs=signalneeds, tf=tf, cwd='tf'+str(tf), lab=["GEANT"], relative_cpu=5/8, mem='2000')
```
sawenzel, line 325: are you sure 5/8 is interpreted as float (and not as an int in which case it would be 0)? Better to put a dot.

MC/bin/o2dpg_sim_workflow.py
sawenzel: (actually we should probably cut the float at 2 decimals ... so that the JSON is still readable.

MC/bin/o2dpg_sim_workflow.py
benedikt-voelkel: Python 3 does "true divide" by default. So that will always be a float

MC/bin/o2dpg_sim_workflow.py
benedikt-voelkel: Would a float `<1` also be okay? So say `0.15`?

sawenzel: sure. It just means that the task uses around 15% of a full CPU.

sawenzel (approved):

PR #71, opened 2021-06-22
Update workflow handling

benedikt-voelkel:
* revise command line usage of o2dpg-workflow-tools.py

* update README and provide usage of o2dpg-workflow-tools.py there

* implement common functionality in o2dpg_workflow_utils.py
  currently used in o2dpg_sim_workflow.py and o2dpg-workflow-tools.py

* module import handled via relative paths:
  sys.path.append(join(dirname(__file__), '.', 'o2dpg_workflow_utils'))
  from o2dpg_workflow_utils import createTask, dump_workflow


sawenzel:
Tested locally. Seems to work and we can use this as a first basis.

PR #74, opened 2021-07-05
First commit of xicc generation macros

ddobrigk:


preghenella:
hi @ddobrigk 
thanks for the commit.
there is already a `generator_pythia8_gun.C` file in this repository
https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/PWGHF/pythia8_gun/generator_pythia8_gun.C

it would be IMO nice to avoid duplicated code and make sue that the `generator_pythia8_gun.C` can be used universally.

Thanks

ddobrigk:
Hi @preghenella, you are fully correct! I will change the naming of this generator. In fact, this generator is quite different than the gun, so the name is, for starters, not really okay. I will close this PR and open another one very shortly.

PR #79, opened 2021-07-15
Initial integration with QC Tasks

knopers8:
This adds 5 QC Tasks, which can be enabled with --include-qc flag passed to o2dpg_sim_workflow.py. In this iteration they publish results directly to QCDB (at ccdb-test) and results from many TFs and sub-jobs are not merged.

More Tasks to come, also merging will be done in future PRs.

An example to run it can be found in `MC/run/examples/O2DPG_pp_minbias_one_tf_qc.sh`

### DIFF
--- a/MC/run/examples/O2DPG_pp_minbias_one_tf_qc.sh
+++ b/MC/run/examples/O2DPG_pp_minbias_one_tf_qc.sh
@@ -18,12 +18,14 @@ MODULES="--skipModules ZDC"
 SIMENGINE=${SIMENGINE:-TGeant4}
 
 # create workflow
-./o2dpg_sim_workflow.py -eCM 14000  -col pp -gen pythia8 -proc inel -tf 1    \
+${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow.py -eCM 14000  -col pp -gen pythia8 -proc inel -tf 1    \
                                                        -ns 20 -e ${SIMENGINE}                   \
-                                                       -j ${NWORKERS} -interactionRate 500000
+                                                       -j ${NWORKERS} -interactionRate 500000   \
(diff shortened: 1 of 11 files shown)


knopers8:
@sawenzel Does O2DPG require explicit dependencies on O2 and now also QC, or it just requires users to load what is needed manually?

@davidrohr @chiarazampolli  @tomas-herman @tmtetwa FYI, this is some initial effort in integrating QC with the simulation data challenge scripts. I added the MFT Clusters and Tracks QC, TPC Tracking QC, RECO Vertexing QC and TRD Digits QC as an exercise. Please let me know if you have any suggestions.

sawenzel:
@knopers8 : There is some runtime dependence. One should make sure to say
`alienv enter O2,O2DPG,QC`  or put QC into the `O2sim` meta package. So I think nothing special needs to be done.

MC/run/examples/O2DPG_pp_minbias_one_tf_qc.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+#
+# A example workflow MC->RECO->AOD for a simple pp min bias 
+# production
+
+# make sure O2DPG + O2 is loaded
+[ ! "${O2DPG_ROOT}" ] && echo "Error: This needs O2DPG loaded" && exit 1
+[ ! "${O2_ROOT}" ] && echo "Error: This needs O2 loaded" && exit 1
```
sawenzel, line 9: maybe one could add a check for `QC_ROOT`  or similar here

knopers8: Indeed, I will add this, thanks.

PR #84, opened 2021-07-19
configuration files for simulation data challenge (prompt / non-prompt quarkonia)

ffionda:
…t quarkonia)


MC/config/PWGDQ/ini/GeneratorHF_bbbar_fwdy.ini
```diff
@@ -0,0 +1,32 @@
+### The setup uses an external event generator
+### This part sets the path of the file and the function call to retrieve it
+
+#[GeneratorExternal]
```
preghenella, line 4: why is this line commented?

MC/config/PWGDQ/ini/GeneratorHF_bbbar_fwdy.ini
```diff
@@ -0,0 +1,32 @@
...
+
+[GeneratorPythia8]
+config = ${O2DPG_ROOT}/MC/config/common/pythia8/generator/pythia8_hf.cfg
+hooksFileName = ${O2DPG_ROOT}/MC/config/PWGHF/pythia8/hooks/pythia8_userhooks_qqbar.C
+hooksFuncName = pythia8_userhooks_bbbar(-4.3,-2.3)
+
+### The setup uses the base configuration of the decayer which is loaded from the file specified by config[0].
+### On top of the base configuration, two more sets of settings are loaded sequentially from config[1] and [2].
+
+[DecayerPythia8]
```
preghenella, line 19: we need to be careful here (and also with the HF generator example).
because this is the setting for particle decays at GEANT level.

one needs to make sure that settings for decays at generator level are as requested

MC/config/PWGDQ/ini/GeneratorHF_bbbar_midy.ini
```diff
@@ -0,0 +1,32 @@
+### The setup uses an external event generator
+### This part sets the path of the file and the function call to retrieve it
+
+#[GeneratorExternal]
```
preghenella, line 4: why is this line commented?

MC/config/PWGDQ/ini/GeneratorHF_bbbar_fwdy.ini
ffionda: Ciao Roberto, thanks, you are right, I'll fix this in a moment.

MC/config/PWGDQ/ini/GeneratorHF_bbbar_fwdy.ini
ffionda: In my case I'm using EvtGen for decaying B hadrons, so I'm just loading  the default decay configuration file (${O2DPG_ROOT}/MC/config/common/pythia8/decayer/base.cfg) where B-hadron decays are  switched off. At the generation level everything works fine, I checked the output of the simulation and I can see B-hadrons decayed (by EvtGen) according to the settings.

preghenella: uhm... ok. I am puzzled.
I am pretty sure these lines should not matter in your case.

ffionda: Ciao Roberto, sorry for answering with some delay. I've removed the lines corresponding to [DecayerPythia8] as well as [SimUserDecay] and the simulation (at MCtruth level) works as expected. Just let me know if I should change something else. Thanks a lot for your help!

preghenella (approved):

PR #88, opened 2021-07-23
running scripts (PWGDQ)

ffionda:
Running scripts for running PWGDQ simulations (min. bias pythia + prompt charmonium signals, mid / fwd rapidity). 
@sawenzel: these are tested locally at MC truth level. If possible, we would like to include them in one of the nightlies tests for testing the full simulation chain. Thanks a lot in advance for any comment.


gconesab:
Hi @ffionda , I will approve it to not have it lingering, but you should remind us after the vacation period (Sandro is already away for the coming weeks and I will be away now)

PR #91, opened 2021-09-01
update of scripts+macros for non-prompt J/psi simulations

ffionda:
Dear all, the commit includes:

1) running scripts for non-prompt jpsi simulations (full chain, mid-y / fwd-y) 

2) function (selectDaughterFromHFwithinAcc.C) for triggering on the rapidity of the non-prompt J/psi (since in the current version in O2DPG a rapidity cut is applied only for at least one b or bar{b}, so a lot of J/psi from b  go outside the acceptance). We plan to use the same function for Low Mass Dilepton pairs (from HF) simulations. 

3) small fix / update of initialization files -> name of fileName and funcName of external generator are provided in the initialization files (i.e. file.ini) and not from the command line; in addition I’ve added the lines for triggering on the rapidity (i.e. “[TriggerExternal]” field)

4) update of generation macro -> remove the centrality dependent injection for pp case (otherwise since impPar = 0 by default the number of injected signals are those we would have for most central collisions)

Thanks!


gconesab:
It looks good to me, I will approve later if @preghenella or @sawenzel do not have any comment

PR #94, opened 2021-09-17
Put QC results in a dedicated directory in QCDB

knopers8:
@chiarazampolli This will put all the QC results from these simulation scripts to `qc_mc` directory on qcg-test.cern.ch . You might want to include a similar change in your PR, or I will add it here after yours is merged.


chiarazampolli:
Thanks @knopers8 , I will include it in the PR when I am back next week, if it is not merged before. Or you could open a PR on my branch, and I can merge it (sorry, but it will be difficult to work on the code for me till Tue).

PR #96, opened 2021-09-22
Update the parameters for the TOF matching QC

knopers8:
Putting the same parameters as in other tasks, mainly to make the results go to "qc_mc" directory.


knopers8:
FYI @chiarazampolli

PR #100, opened 2021-09-23
Introduce TPC q/pt cutoff to reduce loopers

shahor02:
@sawenzel this is to emulate the same pT cuttof in TPC as we effectively have in the async. reco.

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -108,6 +108,10 @@ SIMENGINE=args.e
 BFIELD=args.field
 RNDSEED=args.seed    # 0 means random seed ! Should we set different seed for Bkg and signal?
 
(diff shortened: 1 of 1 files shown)


MC/bin/o2dpg_sim_workflow.py
```diff
@@ -105,6 +105,10 @@
 BFIELD=args.field
 RNDSEED=args.seed    # 0 means random seed ! Should we set different seed for Bkg and signal?
 
+Q2PTCUTOFF=20 # nominal q/Pt cut-off for TPC
+if float(BFIELD)!=0:
+   Q2PTCUTOFF*=5/abs(float(BFIELD));
+Q2PTCUTOFFSTR='GPU_rec.maxTrackQPt='+str(Q2PTCUTOFF)
```
sawenzel, line 111: the last variable is not needed

sawenzel (approved):

PR #104, opened 2021-10-20
Use O2DPG for TOFmatching JSON, till QC is tagged

chiarazampolli:
And use the MC version.


chiarazampolli:
Merging, @sawenzel  is away, and this should be trivial.
@knopers8 , FYI.

PR #116, opened 2021-11-05
adjust dependencies of event-track-QA task

mario-krueger:


MC/bin/o2dpg_sim_workflow.py
```diff
@@ -878,7 +878,7 @@ def addAnalysisTask(tag, cmd, output, needs=[AOD_merge_task['name']],
 
    # Event and track QA
    addAnalysisTask(tag="EventTrackQA",
-                   cmd="o2-analysis-pp-qa-event-track --etaMin -0.8 --etaMax 0.8 --numberOfContributorsMin 1 --vertexPositionBins 200 --vertexPositionZMin -20 --vertexPositionZMax 20 --vertexPositionXYMin -0.1 --vertexPositionXYMax 0.1 --checkPrimaries 1 --numbersOfContributorsToPVBins 500 --numbersOfContributorsToPVMax 500", output="AnalysisResults.root")
+                   cmd='o2-analysis-timestamp --configuration json://${O2DPG_ROOT}/MC/config/QC/json/event-track-qa.json | o2-analysis-event-selection --configuration json://${O2DPG_ROOT}/MC/config/QC/json/event-track-qa.json | o2-analysis-trackextension --configuration json://${O2DPG_ROOT}/MC/config/QC/json/event-track-qa.json | o2-analysis-trackselection --configuration json://${O2DPG_ROOT}/MC/config/QC/json/event-track-qa.json | o2-analysis-pp-qa-event-track --configuration json://${O2DPG_ROOT}/MC/config/QC/json/event-track-qa.json', output="QAResults.root")
```
sawenzel, line 881: Fine in principle. I am just wondering if we should make the output file name yet more descriptive and unique? `QAResults.root` --> `EventTrackQAResults.root` ?

mario-krueger: I would also be in favour of this. However, the way it is currently implemented, I think the `output` argument of `addAnalysisTask()` refers to the file name that comes out of the analysis framework (which can only be `AnalysisResults.root` or `QAResults.root`). If I understand the code of @njacazio correctly, the final name of the file will then be `QAResults_EventTrackQA.root`. We could of course change this.

njacazio: This is correct, this is how it is treated, I would still keep the information of the output file i.e. `AnalysisResults` -> `AnalysisResults_` and `QAResults` -> `QAResults_` or something similar

mario-krueger: Thanks a lot for clarifying!
Can we merge this? Also, it seems like in the nightly grid productions the analysis output is not enabled (at least I did not find any 'Analysis' folder there). Can we include this step in order to get the QA histograms in the next iteration?

sawenzel: @mario-krueger : The actual options for GRID productions are not maintained here (or in a repo). Catalin takes care of this manually. You could try to enable them first of all for the alibi nightlies here: https://gitlab.cern.ch/aliperf/alibibenchtasks/-/tree/master/, so that we can see how it goes. For example https://gitlab.cern.ch/aliperf/alibibenchtasks/-/blob/master/O2DPG_pp_minbias_pilotbeam_0.2T.sh ??

mario-krueger: Thank you! I will add it to the alibi nightly script as you suggested.

PR #117, opened 2021-11-08
Option to pass run-number from outside

sawenzel:
For now passing directly to AOD conversion.


sawenzel:
This needs: https://github.com/AliceO2Group/AliceO2/pull/7548

PR #118, opened 2021-11-08
Download and use [ITS|MFT]dictionary.bin

sawenzel:
For, let's fetch these necessary files manually.
At some point they should be fetched by the tasks from CCDB.


shahor02:
looks fine as a temporary solution

PR #131, opened 2021-12-08
No need for digi-local policy for TOF matching QC config file

knopers8:


chiarazampolli:
Absolutely correct! thanks.

PR #133, opened 2021-12-09
Speed up startup / simplification

sawenzel:
Speeding up startup time of executables by using
the absolute path of O2 executables:
* less system calls during forking and less
  search in $PATH

Simplification of AOD creation (taking out
workaround no longer necessary)


davidrohr:
wow, how much difference does this make?
Perhaps we should use it for sync processing as well...

sawenzel:
Well. Not sure you will see it. We have lot's of other garbage (like enormous LD_LIBRARY_PATH). But I analysed with `strace` and found that even during the fork process in DPL, the paths are searched again and again. So this is a first cheap step in reducing system calls. Others can be done.

PR #138, opened 2021-12-10
Correct file paths for ITS async QC

knopers8:


davidrohr:
Are we sure these files are always in the current folder?
In the processing workflow, we have the `$FILEWORKDIR` variable for that.

knopers8:
This is what used to work during the first pass, but it would be indeed safer to use `$FILEWORKDIR`. Though I am not sure if it would be resolved within the ITS task. Any idea @JianLIUhep ?

iravasen:
@knopers8 , in ITS tasks we read directly geomPath and clusterDictionaryPath but if we have to modify in order to parse FIRST another variable containing the path to a FILEWORKDIR it is fine with us. Let me know.

PR #140, opened 2021-12-10
We need to also have the matching and vtx QC

chiarazampolli:
@davidrohr , @knopers8 : I don't know if this is the best way to do it, but we need to add the matching to the list of detectors_qc to have also the ITSTPC matching, TOF matching, primary vtx QC.


DATA/common/setenv.sh
```diff
@@ -15,7 +15,10 @@ LIST_OF_GLORECO="ITSTPC,TPCTRD,ITSTPCTRD,TPCTOF,ITSTPCTOF,MFTMCH,PRIMVTX,SECVTX,
 
 # Detectors used in the workflow / enabled parameters
 if [ -z "${WORKFLOW_DETECTORS+x}" ] || [ "0$WORKFLOW_DETECTORS" == "0ALL" ]; then export WORKFLOW_DETECTORS=$LIST_OF_DETECTORS; fi
-if [ -z "${WORKFLOW_DETECTORS_QC+x}" ] || [ "0$WORKFLOW_DETECTORS_QC" == "0ALL" ]; then export WORKFLOW_DETECTORS_QC=$WORKFLOW_DETECTORS; fi
+if [ -z "${WORKFLOW_DETECTORS_QC+x}" ] || [ "0$WORKFLOW_DETECTORS_QC" == "0ALL" ]; then
+    export WORKFLOW_DETECTORS_QC=$WORKFLOW_DETECTORS;
+    WORKFLOW_DETECTORS_QC+=",$LIST_OF_GLORECO"
```
davidrohr, line 20: Why don't you just do:
```
if [ -z "${WORKFLOW_DETECTORS_QC+x}" ] || [ "0$WORKFLOW_DETECTORS_QC" == "0ALL" ]; then export WORKFLOW_DETECTORS_QC="$WORKFLOW_DETECTORS,$LIST_OF_GLORECO"; fi
```

davidrohr: actually, the way you have it right now is actually wrong, since you export `WORKFLOW_DETECTORS_QC=$WORKFLOW_DETECTORS` and then you modify it only for the local shell.
If this is used in a subshell, it will loose the `$LIST_OF_GLORECO`

davidrohr (changes requested):
In general I agree, this was missing, and appending what you do it should work (though I just checked the source code, haven't tested it). Please just address my inline comment. Then consider it approved :)

DATA/common/setenv.sh
chiarazampolli: Done. I thought like this it was clearer in case it was not ok ;-)

davidrohr (approved):

chiarazampolli:
Thanks @davidrohr , I am now merging. 
For now, we still use a QC json from input list, because I need to verify with Piotr some options for some tasks. But soon we will just build teh QC json on the fly.

Chiara

chiarazampolli:
Ah, you already merged :-) thanks!

PR #143, opened 2021-12-16
Apply special ITS-MFT ALPIDE strobe setting for pp processing

sawenzel:
This is asked for the pp production. Setting applied
according to input from @shahor02.


sawenzel:
@shahor02 : Could please take a look and approve?

sawenzel:
@shahor02 : Is is correct to apply this for any pp collision system, or should it depend on interaction rate, other stuff as well?

shahor02 (approved):
Hi @sawenzel 
Looks fine, except, as I wrote, the TOF-matcher does not need these params. 
Concerning the values: there will be a discussion next week about the pp program and corresponding strobe lengths, but since it is important (and difficult) to have short strobes at hight rate, whatever we chose for it will work also at low rate, so I don't think we will have a rate dependence per se. What is possible is that we will start with longer strobes and go to shorter ones once we are sure that the readout is stable.

PR #145, opened 2021-12-17
Adjust pp specific setting for primary vertexer

sawenzel:
Taken as suggested in https://alice.its.cern.ch/jira/browse/O2-2691

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -699,10 +699,44 @@ for tf in range(1, NTIMEFRAMES + 1):
    workflow['stages'].append(MFTMCHMATCHtask)
 
    ## Vertexing
+   PVConfig = {**AlpideConfig} # start with Alpide config which is relevant here
+   if COLTYPEIR == 'pp': 
+      if 301000 <= int(args.run) and int(args.run) <= 301999:
+         # put specific pp tunes for pilot beam
+         # taken from JIRA https://alice.its.cern.ch/jira/browse/O2-2691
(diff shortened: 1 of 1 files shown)


sawenzel:
@gconesab @shahor02

shahor02 (commented):
Hi @sawenzel 
The settings are ok but they should be used only for the pilot beam MC as many cuts/tolerances are relaxed to account for uncalibrated data. 
If the aim is to have general pp settings, then it should be 
```
{"pvertexer.acceptableScale2" : 9,
                       "pvertexer.dbscanMaxDist2" : 36,
                       "pvertexer.dbscanDeltaT" : 24,
                       "pvertexer.maxChi2TZDebris" : 100,
                       "pvertexer.maxMultRatDebris" : 1.,
                       "pvertexer.dbscanAdaptCoef" : 20.}
```

sawenzel:
The aim was to satisfy the request made in the ticket https://alice.its.cern.ch/jira/browse/O2-2691. I didn't know this was for pilot beam only. In principle, I would prefer to put generic settings. The best would probably be to put into place CCDB objects for this as it will otherwise be difficult to maintain the code.
I can for now cheat and make the customization dependent on the run-number? What do you think?

shahor02:
@sawenzel yes, eventually it will be set from the CCDB, but at the moment there is no custom query for such objects and I would not introduce them given that (hopefully) soon we will have DPL support for that.

sawenzel:
reworked to contain generic pp settings as well exceptional ones for pilot beam MC run ranges 301000 - 301199.

PR #148, opened 2021-12-21
HF MC config

mmazzilli:


gconesab:
Ciao Marianna,
the commit is fine with me, but I see in run_pp_PWGHF_testbeam_ccbar_trigger.sh  that you have some local path instead of the proper path in o2dpg to trigger_hf.ini.
Can you change that?
Thanks

mmazzilli:
> 

Ciao Gustavo, 

thanks a lot for noticing it! I've just fixed it.

PR #152, opened 2022-01-05
[MFT] Switch to MFT ASync QC task

rpezzi:
MFT Async QC task is used in central reconstruction and should also be used in official productions.
Also changed `MFTAlpideParam.roFrameLengthInBC` to our lucky number.


knopers8 (changes requested):
Thanks, please also update the task name and config file here:
https://github.com/AliceO2Group/O2DPG/blob/edd876998524613003c4b05008a3686bae54f24f/MC/bin/o2dpg_qc_finalization_workflow.py#L51

Please also consider removing the tracks QC config file if it becomes unused now.

rpezzi:
Thanks @knopers8. Done.

knopers8 (approved):

PR #154, opened 2022-01-05
Update CCDB URL

benedikt-voelkel:


benedikt-voelkel:
Dear @chiarazampolli 
Can we merge this?

chiarazampolli:
Hi @benedikt-voelkel , 

I would prefer that @shahor02  comments. On 15.11.2021, he sent this messgae to WP13:
```
Dear all,

For historical reasons the ITS and MFT cluster dictionary files were shipped in bin format. To be compatible with root files supported by the CCDB API we should switch to root dictionaries which are produced by the sames macros which produce bin files.

With https://github.com/AliceO2Group/AliceO2/pull/7618 the files are loaded via new readFromFile(fname) method which can handle both root and bin files. The old access via bin files is still supported (and if instead of the full file path only directory is provded via o2::itsmft::ClustererParam::.dictFilePath, the default binary filename will be added).

You can substitute your cached ITS/MFTdictionary.bin files by equivalent (same dictionaries) root files from https://alice.its.cern.ch/jira/browse/O2-2288?filter=-2:
https://cernbox.cern.ch/index.php/s/9OykRbGq4WzHeVG ITSdictionary.root
https://cernbox.cern.ch/index.php/s/b4qaFiljqnWREyD MFTdictionary.root
or
/home/epn/odc/files/ITSdictionary.root, /home/epn/odc/files/MFTdictionary.root

Once we switch to direct CCDB access, the access via binary files will be suppressed.

Cheers,
 Ruben 
```
so I am not sure one should use the bin files in CCDB. 

Chiara

shahor02:
Hi @chiarazampolli 

At the moment I have simply extended the validity of the existing objects, they are fetched to local files rather than accessed from the process. I will upload root versions once we start reading them as usual objects.
In any case, this or fixes a general bug: the production server should does not need to use 8080 port.

Cheers
  Ruben

chiarazampolli:
Hi @shahor02 ,
Indeed, I did not pay attention that he was using the production CCDB with the port, I thought he was using the test one. 
I will approve the PR. 
Chiara

PR #161, opened 2022-01-11
Format TOF json files to highlight discrepancies

njacazio:
@chiarazampolli with this I uniform the two json by formatting them

### DIFF
--- a/DATA/production/qc-async/itstpctof.json
+++ b/DATA/production/qc-async/itstpctof.json
@@ -1,57 +1,55 @@
 {
-  "qc": {
-    "config": {
-      "database": {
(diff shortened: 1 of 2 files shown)


chiarazampolli (commented):
For me it is in principle fine, but why did you add the space before the ":" in the various fields? @knopers8 , what do you prefer? The ITS json does not have such spaces for example.

njacazio:
I blindly used clang-format to handle it

knopers8:
I don't mind if there is a space there or not. If we see that it is somehow troublesome, we can make a campaign to unify and reformat all jsons.

PR #162, opened 2022-01-11
Tuning pvertexer setting for high-energy pp

sawenzel:
Following request in production ticket
https://alice.its.cern.ch/jira/browse/O2-2679 (date 10.01.2022)

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -703,6 +703,7 @@ for tf in range(1, NTIMEFRAMES + 1):
    if COLTYPEIR == 'pp': 
       if 301000 <= int(args.run) and int(args.run) <= 301999:
          # put specific pp tunes for pilot beam
+         # (this should at some moment come from CCDB)
          # taken from JIRA https://alice.its.cern.ch/jira/browse/O2-2691
          PVConfig.update({"pvertexer.acceptableScale2" : 9,
                           "pvertexer.minScale2" : 2.,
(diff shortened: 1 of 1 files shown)


sawenzel:
@shahor02

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -724,6 +725,16 @@ def createRestDigiTask(name, det='ALLSMALLER'):
...
+      elif 302000 <= int(args.run) and int(args.run) <= 309999:
+         # specific tunes for high pp
+         # run range taken from https://twiki.cern.ch/twiki/bin/viewauth/ALICE/O2DPGMCSamplingSchema
+         # taken from JIRA https://alice.its.cern.ch/jira/browse/O2-2691
+         PVConfig.update({"pvertexer.dbscanDeltaT" : 7,
+                          "pvertexer.maxChi2TZDebris": 50,
+                          "pvertexer.maxMultRatDebris": 1.,
+                          "pvertexer.dbscanAdaptCoef" : 20,
+                          "pvertexer.maxVerticesPerCluster" : 20,
+                          "pvertexer.dbscanMaxDist2" : 36})
```
shahor02, line 737: this looks fine, but:
https://github.com/AliceO2Group/O2DPG/blob/80a2db7726490319320c4a2c46fdb85406af9dc9/MC/bin/o2dpg_sim_workflow.py#L467 : I think we should have here 128, this is already default for the HBFUtils and this is what was used in the pilot beam.

https://github.com/AliceO2Group/O2DPG/blob/80a2db7726490319320c4a2c46fdb85406af9dc9/MC/bin/o2dpg_sim_workflow.py#L476-L477 these settings are ok for the future pp simulation, but they should not be used for the pilot beam, there we had 891 and 198 respectively.

sawenzel: Ok. I am going to address these other points in a separate commit.

PR #165, opened 2022-01-17
Enable diffraction for pp sim

aalkin:
This enables diffraction in pythia8 for pp simulation, which is, in particular, required for pilot beam simulation. 

See, for example, pythia8 configs at http://mcplots.cern.ch/?query=plots,ppppbar,mb-inelastic,eta#pp900


aalkin:
Hi @preghenella,
It seems to make a difference in the end result, however it is also entangled with other issues (AOD merging). I'll report as soon as I have a final answer.

PR #166, opened 2022-01-17
Move cut on min N TPC cl from 40 to 60

chiarazampolli:
Following async-QC meeting.
@shahor02 , @stheckel


chiarazampolli:
There is a conflict I need to solve, I will open a new one.

PR #168, opened 2022-01-18
Fix comparison

aalkin:


aalkin:
@gconesab sorry for bothering, #165 had a mistake, the comparison should be with integer, not a string.

PR #176, opened 2022-01-25
o2dpg_sim_workflow.py: options to enable forward assessment

rpezzi:
Add options to enable assessment workflows to `o2dpg_sim_workflow.py`

* `--mft-assessment-full` enables MFT standalone assessment workflow, produces `MFTAssessment.root`
*  `--fwdmatching-assessment-full` enables global forward matching assessment workflow, produces `GlobalForwardAssessment.root`

Working example on `MC/run/PWGDQ/runPromptCharmonia_fwdy_pp_assessment.sh`

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -94,6 +94,11 @@ parser.add_argument('--include-analysis', '--include-an', '--analysis',
 
 # MFT reconstruction configuration
 parser.add_argument('--mft-reco-full', action='store_true', help='enables complete mft reco instead of simplified misaligned version')
+parser.add_argument('--mft-assessment-full', action='store_true', help='enables complete assessment of mft reco')
+
+# Global Forward reconstruction configuration
+parser.add_argument('--fwdmatching-assessment-full', action='store_true', help='enables complete assessment of global forward reco')
+
 
 args = parser.parse_args()
 print (args)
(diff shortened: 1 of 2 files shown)


MC/bin/o2dpg_sim_workflow.py
```diff
@@ -94,6 +94,11 @@
 
 # MFT reconstruction configuration
 parser.add_argument('--mft-reco-full', action='store_true', help='enables complete mft reco instead of simplified misaligned version')
+parser.add_argument('--mft-assessment-full', action='store_true', help='enables complete assessment of mft reco')
+
+# Global Forward reconstruction configuration
+parser.add_argument('--fwdmatching-assessment-full', action='store_true', help='enables complete assessment of global forward reco')
```
sawenzel, line 100: Could you please define default values for these arguments? Otherwise, I believe, the value may be `None` and lead to runtime crashes.

rpezzi: Hi @sawenzel , Thanks for checking. 
Please note that other parsed parameters are used in boolean comparisons without default values. For instance: `--embedding`, `--combine-smaller-digi`, `--early-tf-cleanup`, `--mft-reco-full`. Default value here does not seem necessary.  Shall I proceed anyway?

sawenzel: Indeed. My bad. action=`store_true` does it implicitely.

PR #187, opened 2022-01-30
Add configs for MC production needed for HF event filtering for high-energy pp program

fgrosa:


mpuccio:
@gconesab this is good for me! Can we merge it to start some testing jobs on the grid?

gconesab:
Hi, merging, as soon as daily tag is created, tests can be done

mpuccio:
Thanks a lot!

PR #194, opened 2022-02-02
Adding noise scan for MFT.

tomas-herman:


DATA/production/calib/mft-noise-aggregator.sh
```diff
@@ -0,0 +1,32 @@
...
+  ARGS_ALL+=" --shm-throw-bad-alloc 0"
+fi
+if [ $NORATELOG == 1 ]; then
+  ARGS_ALL+=" --fairmq-rate-logging 0"
+fi
+ARGS_ALL_CONFIG="NameConf.mDirGRP=$FILEWORKDIR;NameConf.mDirGeom=$FILEWORKDIR;NameConf.mDirCollContext=$FILEWORKDIR;NameConf.mDirMatLUT=$FILEWORKDIR;keyval.input_dir=$FILEWORKDIR;keyval.output_dir=/dev/null;$ALL_EXTRA_CONFIG"
+
+PROXY_INSPEC=A:MFT/DIGITS/0;B:MFT/DIGITSROF/0"
+
+WORKFLOW="o2-dpl-raw-proxy $ARGS_ALL --proxy-name mft-noise-input-proxy --dataspec \"$PROXY_INSPEC\" --network-interface ib0 --channel-config \"name=readout-proxy,method=bind,type=pull,rateLogging=1,transport=zeromq" | "
```
davidrohr, line 20: Without trying locally, I think this must be wrong. The --proxy-name must match the name in the --channel-config. 
I also think the channel names for the input and the output must match, such that FairMQ creates the connection. Could you check https://github.com/AliceO2Group/O2DPG/tree/master/DATA#calibration-workflows-with-aggregator

davidrohr (changes requested):

DATA/production/calib/mft-noise-aggregator.sh
tomas-herman: Thank you for spotting the error. It should be fixed now in the last commit.

PR #199, opened 2022-02-04
Adding scripts + readAO2D macro for data processing

chiarazampolli:


chiarazampolli:
Hello @catalinristea, 

Here the scripts + macro that we discussed. I will add a comment in JIRA, with the script for the grid in a few hours. 

Chiara

chiarazampolli:
Merged, since they won't break anything, unless explicitly used.

PR #203, opened 2022-02-07
[EMCAL-755, EMCAL-756] Adding EMCAL QC for MC and async reconstruction

mfasDa:
- EMCAL-755: QC workflow for MC
- EMCAL-756: QC workflow for async reconstruction

### DIFF
--- a/DATA/production/qc-async/emc.json
+++ b/DATA/production/qc-async/emc.json
@@ -1,0 +1,41 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "qcdb.cern.ch:8083",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable"
+      },
+      "Activity": {
+        "number": "42",
+        "type": "2"
+      },
+      "monitoring": {
+        "url": "influxdb-unix:///tmp/telegraf.sock"
+      },
+      "consul": {
+        "url": "alio2-cr1-hv-aliecs.cern.ch:8500"
+      },
+      "conditionDB": {
+        "url": "qcdb.cern.ch:8083"
+      }
+    },
+    "tasks": {
+      "DigitsTask": {
+        "active": "true",
+        "className": "o2::quality_control_modules::emcal::DigitsQcTask",
+        "moduleName": "QcEMCAL",
+        "detectorName": "EMC",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "direct",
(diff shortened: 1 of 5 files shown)


chiarazampolli:
Ok with me! If nobody has further comments, I will merge by tonight, to have it for the next MC cycle. 
@mfasDa , to be sure, you tested it with the simulation chain, right?

mfasDa:
Currently running, therefore adding WIP label until tests are done

knopers8 (approved):
It looks correct now, but please do confirm with a test.

mfasDa:
I have tried with the scripts from the current min. bias production LHC21k6, now the emccells.root is there but EMCAL is still excluded from filling the AODs:

```
input sources ITS,MFT,MCH,TPC,ITS-TPC,MFT-MCH,ITS-TPC-TOF,TPC-TOF,FT0,FV0,FDD,CTP,TPC-TRD,ITS-TPC-TRD
```

EMC should be added in order to have the EMC tables filled. 

Using MC/run/examples/O2DPG_pp_minbias_multiple_tf_qc.sh I checked that the EMC QC is running and the output histograms are produced. The following file should be added to the files to be stored: emcDigitsQC.root

chiarazampolli:
Hi @mfasDa ,

Do you mean to change this line:

https://github.com/AliceO2Group/O2DPG/blob/965c5976381fce92da90c77f181f12e64712c591/MC/bin/o2dpg_sim_workflow.py#L853

adding `EMC`? Have you tried it?
I will tell Catalin to include the EMCAL file. 

Cheers,

Chiara

knopers8:
@mfasDa I had a random look at a dozen of the histograms in `qc_mc/EMC/MO/DigitsTask/`. Just to make sure, is it expected that the vast majority is empty? (a few of them have 3 entries, which seems correct)

mfasDa:
@knopers8 `CAL` (Histrograms for the calibration trigger) should be empty, because calibration events are not simulated, however the `PHYS` histograms should be filled. Looking at the output I see that cells are produced however the trigger bit is not set, which will confuse the QC as the trigger class is unknown. Needs to be added in the digitizer.

mfasDa:
Link to fix for Physics Trigger in simulation: https://github.com/AliceO2Group/AliceO2/pull/8089

mfasDa:
After setting the trigger type to PhT in the simulated trigger record the DigitsQC histograms are filled now. I have removed the WIP label in order to have the PR merged. Unfortunately up to now I was not able to test the AOD filling as the workflow crashed in earlier stages.

davidrohr:
The O2 PR was merged, can I merge this now then?

chiarazampolli:
Yes, please (I will do it, since I also can - one iteration less).

PR #204, opened 2022-02-07
TPC-QC: add standard TPC-QC to MC

stheckel:
o Including QC tasks for Clusters, Tracks and PID

### DIFF
--- a/MC/config/QC/json/tpc-qc-standard-direct.json
+++ b/MC/config/QC/json/tpc-qc-standard-direct.json
@@ -1,0 +1,86 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "ccdb-test.cern.ch:8080",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable",
+        "maxObjectSize": "20971520"
+      },
+      "Activity": {
+        "number": "42",
+        "type": "2",
+        "provenance": "qc_mc",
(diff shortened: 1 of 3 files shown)


MC/config/QC/json/tpc-qc-standard-direct.json
```diff
@@ -0,0 +1,86 @@
...
+      }
+    },
+    "tasks": {
+      "Clusters": {
+        "active": "true",
+        "className": "o2::quality_control_modules::tpc::Clusters",
+        "moduleName": "QcTPC",
+        "detectorName": "TPC",
+        "cycleDurationSeconds": "60",
+        "resetAfterCycles": "1",
```
knopers8, line 40: We do not support moving windows in batch mode yet (sorry, I remember you need it). This will have no effect for now, so I would propose to remove the line for clarity.
The same comment applies for other tasks.

knopers8 (approved):
It looks correct. You can consider my one suggestion, but I am fine with merging it as it is.
Could you confirm that `Clusters` work after increasing the MO size limit?

MC/config/QC/json/tpc-qc-standard-direct.json
stheckel: Yes, I agree, for MC its also not needed, at least not in the current setup. Line is removed in all 3 tasks.

stheckel:
I have removed the "resetAfterCycles" for now. I also confirm, that the Clusters work with the increased MO size limit.

PR #208, opened 2022-02-08
optionally add ZDC treatment

sawenzel:


chiarazampolli:
Hi @sawenzel , 

There seems to be a conflict, can you check?

Chiara

cortesep:
I would like to make a quick test on the gain settings that probably are more correct for Pb-Pb rather than p-p. Please could you tell me how can I run a small local production?

sawenzel:
@cortesep:

A typical pp production run would like this:
```shell
#!/bin/bash

#
# A example workflow MC->RECO->AOD for a simple pp min bias 
# production
# ----------- SETUP LOCAL CCDB CACHE --------------------------
export ALICEO2_CCDB_LOCALCACHE=$PWD/.ccdb

# ----------- START ACTUAL JOB  ----------------------------- 
NWORKERS=${NWORKERS:-8}
SIMENGINE=${SIMENGINE:-TGeant4}

# create workflow
# run number 300000 is ok for pp and -0.5 T
${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow.py -eCM 14000  -col pp -gen pythia8 -proc cdiff -tf 5     \
                                                       -ns 200 -e ${SIMENGINE}                   \
                                                       -j ${NWORKERS} -interactionRate 500000     \
                                                       -run 300000                                \
                                                       -confKey "Diamond.width[2]=6" --include-qc --with-ZDC
# run workflow
${O2DPG_ROOT}/MC/bin/o2_dpg_workflow_runner.py -f workflow.json -tt aod
```

This would do 200pp x 5tf, which may be good to start with when ZDC is included. (Without ZDC we do 2000pp per timeframe but this may take some time).

PR #209, opened 2022-02-08
[WIP] ITS threshold scripts modified to work with new workflow scheme

iravasen:
This follows: https://github.com/AliceO2Group/AliceO2/pull/8093


chiarazampolli:
Hi, 

Didn't this one need https://github.com/AliceO2Group/AliceO2/pull/8093?

davidrohr:
indeed, sorry, didn't pay enough attention. But in any case, This is only for the on-the-fly workflow generation, and in there one has to specify an explicit O2DPG version anyway. So with the old O2, new O2DPG shouldn't be used.... Or shall we revert this?

chiarazampolli:
I agree, it is more for consistency. Let's see whay @iravasen says, the PR was also still WIP.

iravasen:
Hi @chiarazampolli, @davidrohr no problem it can stay. The modification that probably will be needed for #8093 will not affect these scripts that will remain anyway valid.

PR #210, opened 2022-02-09
TPC-QC: deactivate Clusters Tasks in MC

stheckel:
We found an issue with the Clusters task, which will likely lead any MC production to crash. This was not spotted in the small-scale tests done before. For now, it has to be deactivated. Will be reactivated, as soon as the issue could be solved.


chiarazampolli:
Thanks, @stheckel !

PR #215, opened 2022-02-10
Enable option for saving parameters at plane for matching training with machine learning

sadekr:


gconesab:
Hi @sawenzel, are you ok with this "hack"?

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -711,6 +713,12 @@ def createRestDigiTask(name, det='ALLSMALLER'):
    MFTMCHMATCHtask['cmd']+= getDPL_global_options()
    workflow['stages'].append(MFTMCHMATCHtask)
 
+   if args.fwdmatching_save_trainingdata == True:
+      MFTMCHMATCHTraintask = createTask(name='mftmchMatchTrain_'+str(tf), needs=[MCHMIDMATCHtask['name'], MFTRECOtask['name']], tf=tf, cwd=timeframeworkdir, lab=["RECO"], mem='1500')
+      MFTMCHMATCHTraintask['cmd'] = '${O2_ROOT}/bin/o2-globalfwd-matcher-workflow ' + putConfigValues({**{"MFTClustererParam.dictFilePath" : "../", "FwdMatching.saveMode" : 2, "FwdMatching.useMIDMatch":"true"} , **AlpideConfig})
+      MFTMCHMATCHTraintask['cmd']+= getDPL_global_options()
+      workflow['stages'].append(MFTMCHMATCHTraintask)
```
sawenzel, line 720: It seems that then there would be 2 different tasks that both run 
`o2-globalfwd-matcher-workflow`?  It would be preferred to just configure the
"MFTClustererParam" based on the new option.

rpezzi: hummm... by removing the task for training data generation one has to choose between generating training data OR generating global muon tracks (and producing AODs). A single execution of the global forward matcher workflow cannot be used for both outputs. With the latest commit [837f5c4](https://github.com/AliceO2Group/O2DPG/pull/215/commits/837f5c4749badc261bcfe2ab42e2f22f9d7d8b23) we loose functionality that was used a recent grid production where both tasks were executed in the same job. With a single task the workflow will crash if one asks for both AOD production and matching training data. 

Please note that we are not sure how often production of training data will be necessary and if demand increases, we will adapt the matcher workflow to generate both in the same execution. 

I advocate for keeping both tasks and while doing so I also ask for suggestions on a better way ensure the new task is executed instead of adding it as a [requirement for AOD production here](https://github.com/AliceO2Group/O2DPG/pull/215/commits/f84d9b0a01d685c5cdad16962d9c83a7a1c34930#diff-d3d7f9e5d4c04202b033167540929ff40471636b8a1bb1ffdbecd1066c051eeaR848-R849).

sawenzel: @rpezzi : You can specify the list of targets to the pipeline runner as a regular expression: `-tt "(aod.*)|(mft.*)"` --> This would build everything needed for AOD as well as all tasks starting with `mft` and their dependencies.

PR #218, opened 2022-02-12
async pass3 zero field runs configuration (O2-2797)

catalinristea:
Ciao @chiarazampolli, 

This is just a clone of standard apass3, needed because we change the tag name for the B=0 runs reco - alien script looks into path from O2DPG, ie DPGRECO=$O2DPG_ROOT/DATA/production/configurations/$ALIEN_JDL_LPMANCHORYEAR/$ALIEN_JDL_LPMPRODUCTIONTAG/$ALIEN_JDL_LPMPASSNAME/async_pass.sh).

We could have bypassed it by using local scripts, but let's do it properly.

Thanks!

Yours,
Catalin.

### DIFF
--- a/DATA/production/configurations/2021/OCT/apass3_zerofield/async_pass.sh
+++ b/DATA/production/configurations/2021/OCT/apass3_zerofield/async_pass.sh
@@ -1,0 +1,206 @@
+#!/bin/bash
+
+# Script to run the async processing
+#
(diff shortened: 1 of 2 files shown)


chiarazampolli:
Hi @catalinristea , 

I agree 100%! Thanks!

Chiara

PR #225, opened 2022-02-16
[EMCAL-524, EMCAL-525] Renaming digits to cell task

mfasDa:
DigitQcTask has been renamed in
QualityControl v1.46 to CellTask for
consistency with its input source.

### DIFF
--- a/DATA/production/qc-async/emc.json
+++ b/DATA/production/qc-async/emc.json
@@ -23,16 +23,16 @@
       }
     },
     "tasks": {
-      "DigitsTask": {
+      "CellTask": {
         "active": "true",
-        "className": "o2::quality_control_modules::emcal::DigitsQcTask",
+        "className": "o2::quality_control_modules::emcal::CellTask",
         "moduleName": "QcEMCAL",
         "detectorName": "EMC",
         "cycleDurationSeconds": "60",
         "maxNumberCycles": "-1",
         "dataSource": {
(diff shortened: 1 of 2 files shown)


mfasDa:
@davidrohr In QC v1.46 we renamed our QC task DigitsQcTask to CellTask and need to adapt the workflows using the renamed task. The DigitsQcTask is not yet changed in order to be able to validate the change. The change was necessary as we are adopting the QC for simulation. Historically Digits and Cells were the same object as in the task the time response of the EMCAL was not simulated. Once the time response was simulated the objects diverged and we modified Digits to contain the single ADC sample and Cells to contain the result of the fit of the time response to the ADC samples. However the name of the QC task on cells stayed with its own name. For the MC production we see a need to also implement a basic QC on digits, for which we need the name. Consequently we performed the renaming now in coordination with the QC experts. The changes in ControlWorkflows are already merged and will be available in the next FLP Suite, so the change needs to be merged in sync that changes come in place on the EPN once the new FLP Suite is installed.

@chiarazampolli @sawenzel I did not perform the corresponding change on the MC side in order not to disturb the ongoing production. We decided to keep the two tasks in parallel for a certain transition period so this is safe. Once indicated by you it is safe to migrate I will adapt the change also in the MC workflow and QC configuration files. Only at that point we will start modifying the DigitsQcTask to monitor MC-related observables.

PR #232, opened 2022-02-19
[R3C-752] Change ITS noise calib to digit inputs with 32 threads

shahor02:
To use clusters input one should select ITS-noise-calibration-clusters from the
`standalone-calibration.desc`

This PR depends on https://github.com/AliceO2Group/AliceO2/pull/8186


chiarazampolli:
Merging, since https://github.com/AliceO2Group/AliceO2/pull/8186 is also merged.

PR #233, opened 2022-02-20
fix for sync calib to update TOF ccdb asap

noferini:


noferini:
This is quite urgent (we want to test it on Tuesday in AliECS).
The reason why we didn't get so far a ccdb entry is that we needed to wait (25 mins instead o f 5 mins to have the entry = 10 slots). And we didn't wait enough.
Now we have to wait just one slot is filled.

PR #245, opened 2022-02-25
added CPV jsons for qc-sync and qc-async workflows

sevdokim:
Hello! Json files for CPV qc-sync and qc-async reco are added. Also json files for PHS are prepared but this will be included into workflow later.

### DIFF
--- a/DATA/production/qc-async/cpv.json
+++ b/DATA/production/qc-async/cpv.json
@@ -1,0 +1,46 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "alio2-cr1-hv-qcdb1.cern.ch:8083",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable"
+      },
+      "Activity": {
+        "number": "42",
+        "type": "2"
+      },
+      "monitoring": {
+        "url": "influxdb-unix:///tmp/telegraf.sock"
+      },
+      "consul": {
+          "url": "alio2-cr1-hv-aliecs:8500"
+      },
+      "conditionDB": {
+        "url": "http://localhost:8084/"
+      }
+    },
+    "tasks": {
+      "PhysicsAsync": {
+        "active": "true",
+        "className": "o2::quality_control_modules::cpv::PhysicsTask",
+        "moduleName": "QcCPV",
+        "detectorName": "CPV",
+        "cycleDurationSeconds": "30",
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "direct",
+          "query": "clusters:CPV/CLUSTERS/0;ctrigrec:CPV/CLUSTERTRIGRECS/0"
+        },
+        "taskParameters": {
+          "": ""
+        }
+      }
+    },
+    "dataSamplingPolicies": [
+        
+    ]
+  }
+}
--- a/DATA/production/qc-sync/phs.json
+++ b/DATA/production/qc-sync/phs.json
@@ -53,7 +53,7 @@
       "id": "phos-clu",
       "active": "true",
       "machines": [
-        "epn"
+        "localhost"
       ],
       "query": "clusters:PHS/CLUSTERS/0;clustertr:PHS/CLUSTERTRIGREC/0",
       "samplingConditions": [
--- a/DATA/production/qc-sync/phs.json
+++ b/DATA/production/qc-sync/phs.json
@@ -1,0 +1,69 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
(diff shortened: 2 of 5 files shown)


DATA/production/qc-async/cpv.json
```diff
@@ -0,0 +1,46 @@
...
+        "detectorName": "CPV",
+        "cycleDurationSeconds": "30",
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "direct",
+          "query": "clusters:CPV/CLUSTERS/0;ctrigrec:CPV/CLUSTERTRIGRECS/0"
+        },
+        "taskParameters": {
+          "": ""
+        }
```
knopers8, line 39: FYI, it should survive without `taskParameters` at all, but there is no harm in being it here.

DATA/production/qc-sync/phs.json
```diff
@@ -0,0 +1,69 @@
...
+      }
+    }
+  },
+  "dataSamplingPolicies": [
+    {
+      "id": "phos-clu",
+      "active": "true",
+      "machines": [
+        "epn"
+      ],
```
knopers8, line 57: Are you sure this works on EPN? in the sampling policy you use the "epn" tag, but in `ClusterTask` it is `localhost`. A local QC workflow can be executed with one `--host` parameter, so either the policy (if `epn`) or the task (if `localhost`) will be spawned, never both.
As far as I remember, EPN QC still uses `--host localhost`. We should probably organize a campaign to move to `epn` which is clearer.

sevdokim: Hi @knopers8 
I have 2 questions regarding that. 
1) What is "localMachines[ ]" ? Is it place where objects for merging are produced?
2) What is "machines" in dataSamplingPolicies?

sevdokim: And one more question: on epns we use "epn" machine to indicate local part of task, or "localhost"? I'm abit confused. Is it just label for machine or real address?

knopers8: I will refer you to the doc:
https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#multi-node-setups
https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#configuration-files-details
if it is still unclear, let me know.

regarding the second question, I believe that `localhost` is still used to label tasks which should run on EPNs. It is probably a remnant of some test setup that propagated to the production.

sevdokim: hi @knopers8 
yes, one more question. If we put localMachines[ "epn" ] then on every epn node same workflow ...| o2-qc --local --host epn ... is spawned, right? Sorry, I don't understand it. What word "enp" stands for? Is it just label or real machine address?

knopers8: It is just a label.
This basically allows you to have one QC config file for both FLPs, EPNs and QC servers (with mergers). Then you can e.g. execute `o2-qc --local --host epn` to spawn only tasks which have the "epn" label, exactly as you say.
You can also have several `localMachines` and then a task will be spawned if `--host` matches any of them. `machines` in data sampling follow the same principle, but if the list is empty, it will be always spawned (for backwards compatibility).

sevdokim: thanks for explanation. Still I have question:
in the qc-workflow.sh qc is added by following command:
`  add_W o2-qc "--config json://$QC_JSON_FROM_OUTSIDE ${QC_CONFIG_PARAM:---local --host ${QC_HOST:-localhost}}" "" 0
`
if I understood correctly, QC_HOST is not defined anywhere so all tasks at epn are now submitted with 'localhost' label so if we want json files to work then all the local parts must be labeled as 'localhost', right?

knopers8: Yes, that is correct.

We will have to organize a campaign to move to "epn". For now, please use "localhost".

sevdokim: Ah, ok! Now I undestood completely.

sevdokim: hi @knopers8 
can you please look again into qc-sync/phs.json? Thanks!

sevdokim: also I din't include phs.json to qc-workflow.sh yet because corresponding adjusted dpl-workflow.sh must be deployed to P2 first in order it to be tested.

knopers8: The file looks correct to me.

Is it ready to merged then?

sevdokim: Hi @knopers8 
I will have chance to test phs.json at epn as soon as https://github.com/AliceO2Group/AliceO2/pull/8224 appears at epn. 
As for cpv, it is tested fine. So I propose to merge it for now.

PR #248, opened 2022-02-25
Adding scripts for ctf_recreation [O2-2767], [O2-2768]

chiarazampolli:


chiarazampolli:
@catalinristea 
These are the scripts to run the ctf_recreation. We still need to define some details for the running on the grid, though. If you agree, I would merge, since they are not breaking anything.

catalinristea:
Ciao Chiara,

Yes, certainly agree.
Thanks!

Yours,
Catalin.

chiarazampolli:
Thanks Catalin, merging.

PR #250, opened 2022-02-28
[QC-746] Custom needs of MC QC finalization should be omitted in standalone

knopers8:


knopers8:
I am running a test on my setup to confirm it does not break anything else, I will merge as soon as it passes, since it affects QC on grid.

PR #269, opened 2022-03-11
MCH: add QC for pedestals (bad channels)

aphecetche:


aphecetche:
superseded by https://github.com/AliceO2Group/O2DPG/pull/291

PR #270, opened 2022-03-11
add TOF digits QC for MC

noferini:
This is the QC digits to be added in MC workflow as for the data (really needed with anchoring!).
The typical workflow is something like

o2-tof-reco-workflow -b --input-type digits --output-type none| o2-qc -b --config json://${O2DPG_ROOT}/MC/config/QC/json/tofdigits.json --local-batch TOFDigitQC.root

### DIFF
--- a/MC/config/QC/json/tofdigits.json
+++ b/MC/config/QC/json/tofdigits.json
@@ -1,0 +1,141 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "ccdb-test.cern.ch:8080",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable"
+      },
+      "Activity": {
+        "number": "42",
+        "type": "2"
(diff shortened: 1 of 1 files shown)


noferini:
@chiarazampolli @knopers8 @njacazio @ercolessi

chiarazampolli:
Ok, I will merge this for now, even if it will have no effect.

Chiara

MC/config/QC/json/tofdigits.json
```diff
@@ -0,0 +1,141 @@
...
+      "query": "tofdigits:TOF/DIGITS/0;readoutwin:TOF/READOUTWINDOW/0;patterns:TOF/PATTERNS",
+      "samplingConditions": [
+        {
+          "condition": "random",
+          "fraction": "0.1",
+          "seed": "1234"
+        }
+      ],
+      "blocking": "false"
+    }
```
knopers8, line 139: this policy seems to be unused, please consider removing it

PR #274, opened 2022-03-14
Fix problem with 'inel' setting in pythia8 configurator

sawenzel:


chiarazampolli:
Good for me too, merged.

PR #275, opened 2022-03-15
add trd qc

bazinski:
Add trd to syn async and in to qc-workflows.sh to get the qc in global runs amongst other things.

### DIFF
--- a/DATA/production/qc-async/trd.json
+++ b/DATA/production/qc-async/trd.json
@@ -1,0 +1,118 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "alio2-cr1-hv-qcdb1.cern.ch:8083",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable"
+      },
+      "Activity": {
+        "type": "2",
+        "number": "42"
+      },
+      "monitoring": {
+        "url": "influxdb-unix:///tmp/telegraf.sock"
+      },
+      "consul": {
+        "url": "alio2-cr1-hv-aliecs:8500"
+      },
+      "conditionDB": {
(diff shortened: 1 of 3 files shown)


bazinski:
ooops I forgot that, did it in a rush before the async qc meeting, fixing now.

bazinski:
Hi @chiarazampolli No the MC only has digits, I can fix that, but there was some python involved, it was a bit of an urgent fix at the time due to it failing for some or other reason. I could not follow and had more pressing things at the time to sort out.
I can come back to if you would like?

DATA/production/qc-async/trd.json
```diff
@@ -0,0 +1,151 @@
...
+  "dataSamplingPolicies": [
+    {
+      "id": "trdall",
+      "active": "true",
+      "machines": [],
+      "query" : "digits:TRD/DIGITS/0;tracklets:TRD/TRACKLETS/0;triggers:TRD/TRKTRGRD/0;rawstats:TRD/RAWSTATS/0",
+      "samplingConditions":[
+        {
+          "condition": "random",
+          "fraction": "0.9",
```
knopers8, line 143: if you need almost all data (90%), you could consider attaching QC directly to the data, as illustrated in here:
https://github.com/AliceO2Group/QualityControl/blob/175baf49916e92552349654a8ce9b3fdb414cd42/Framework/basic-no-sampling.json#L34

knopers8 (commented):
Is there a reason not to add the `qc-async/trd.json` to the corresponding list in `qc-workflow.sh` as well?

bazinski:
Hi @knopers8 I had it on desktop, but not where i was doing the pr from. Fixed now.
qc tasks are direct in the json and I have removed the raw data task as opposed to just disabling it.
Also added it to the qc-workflows.sh

knopers8 (approved):

PR #277, opened 2022-03-16
fix bug in MC config generation script

sawenzel:


sawenzel:
should fix problem in GRID nightlies

PR #278, opened 2022-03-17
Remove TRD from reconstructed detectors for apass4

chiarazampolli:
@noferini , @shahor02 , @sawenzel , @martenole , @tdietel

TRD is now disabled for apass4. 

@sawenzel , can you confirm that you will use this folder for the MC anchored to apass3, to take the correct TPC calibrations, as discussed yesterday?
If so, this is correct, because we want to exclude TRD from MC reco, since in data we saw very low matching efficiency and this cannot now be reproduced in MC.


chiarazampolli:
Agreed via mail with the TRD experts (@martenole ) today - 17.03.2022. 
Merging.

PR #282, opened 2022-03-18
Typo fixed

chiarazampolli:
Comment is possibly used by users, and by MC anchoring script, better to fix it.


chiarazampolli:
@sawenzel

PR #284, opened 2022-03-18
adapt ITS threshold scripts to PR8346 in O2 + added QC

iravasen:
Changes:
- added 6 pipelines for decoder --> compatible with PR8346 in O2
- added QC workflow for calibration scans 
- removed DCS CCDB link from its-aggregator-wf to use the object upload through ccdb-populator-wf
- added --enable-eos to use end-of-stream at P2
- 1 thread in the decoder: multiple thread meaningless in calibration mode because 1 RU data per TF
- 4 threads in o2-its-thr-calib because the improvement is negligible with more threads.

### DIFF
--- a/DATA/production/calib/its-threshold-aggregator.sh
+++ b/DATA/production/calib/its-threshold-aggregator.sh
@@ -18,7 +18,7 @@ ARGS_ALL_CONFIG="NameConf.mDirGRP=$FILEWORKDIR;NameConf.mDirGeom=$FILEWORKDIR;Na
 PROXY_INSPEC="tunestring:ITS/TSTR/0;runtype:ITS/RUNT/0;fittype:ITS/FITT/0;scantype:ITS/SCANT/0;eos:***/INFORMATION"
 
 WORKFLOW="o2-dpl-raw-proxy $ARGS_ALL --proxy-name its-thr-input-proxy --dataspec \"$PROXY_INSPEC\" --network-interface ib0 --channel-config \"name=its-thr-input-proxy,method=bind,type=pull,rateLogging=0,transport=zeromq\" | "
-WORKFLOW+="o2-its-threshold-aggregator-workflow -b --ccdb-url=\"http://alio2-cr1-flp199.cern.ch:8083\" $ARGS_ALL | "
+WORKFLOW+="o2-its-threshold-aggregator-workflow -b --ccdb-url=\"\" $ARGS_ALL | "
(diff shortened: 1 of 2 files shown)


iravasen:
@chiarazampolli this is to work with end-of-stream at P2 and to use multiple-decoder pipelines in the decoder. In addition we added QC.

DATA/production/calib/its-threshold-aggregator.sh
```diff
@@ -18,7 +18,7 @@ ARGS_ALL_CONFIG="NameConf.mDirGRP=$FILEWORKDIR;NameConf.mDirGeom=$FILEWORKDIR;Na
 PROXY_INSPEC="tunestring:ITS/TSTR/0;runtype:ITS/RUNT/0;fittype:ITS/FITT/0;scantype:ITS/SCANT/0;eos:***/INFORMATION"
 
 WORKFLOW="o2-dpl-raw-proxy $ARGS_ALL --proxy-name its-thr-input-proxy --dataspec \"$PROXY_INSPEC\" --network-interface ib0 --channel-config \"name=its-thr-input-proxy,method=bind,type=pull,rateLogging=0,transport=zeromq\" | "
-WORKFLOW+="o2-its-threshold-aggregator-workflow -b --ccdb-url=\"http://alio2-cr1-flp199.cern.ch:8083\" $ARGS_ALL | "
+WORKFLOW+="o2-its-threshold-aggregator-workflow -b --ccdb-url=\"\" $ARGS_ALL | "
```
chiarazampolli, line 21: Hi Ivan, 
Why have you removed the CCDB?

iravasen: to use the o2-ccdb-populator workflow. In o2-its-threshold-aggregator-worfklow, if I leave the ccdb-url empty, it does not use the ccdb API but ships the DCSConfigObject to the ccdb-populator wf.

chiarazampolli: Hi Ivan.
Ok, I see. But honestly, I don't like much that you use an empty argument, wouldn't it be clearer to pass something like "not-used" that you then parse in the task?

iravasen: @chiarazampolli for the task itself it doesn't matter because I have an if condition like that:
```
if (!mCcdbUrl.empty()) {
  //do stuff
}
```
I forgot and actually I can completely remove the --ccdb-url "" from this PR because the default value it's already an empty string. So if you agree I would simply do that.

chiarazampolli: Ciao @iravasen , 

Yes, please, remove it as it is misleading. 

Thanks!

iravasen: done

chiarazampolli:
Thanks!

PR #285, opened 2022-03-18
additional TOF QC, using also TRD tracks

chiarazampolli:
Needs https://github.com/AliceO2Group/QualityControl/pull/1103


chiarazampolli:
@njacazio , @noferini , I am merging this, let's see if I forgot anything and it will break the nightlies (hopefully not!).

chiarazampolli:
Actually we need a protection because if TRD is not there, the inputs will not be found. I will fix this tomorrow, the nighlies should be fine, and hopefully nobody will do anchored MC tonight .

PR #289, opened 2022-03-22
Add TOF QC for all types of tracks only is TRD is active

chiarazampolli:
@noferini , @njacazio , @sawenzel : this check was missing. It should be trivial, but let me know in case you see something wrong.


sawenzel:
Sorry. I fixed this in between without noticing this PR. Can be closed.

PR #291, opened 2022-03-22
MCH: add pedestal QC (and read config from Consul)

aphecetche:


aphecetche:
@chiarazampolli we tested this in standalone runs this morning : we plugged the QC for pedestal runs, and are ready the QC configuration from Consul

PR #295, opened 2022-03-23
Executables

chiarazampolli:
@catalinristea , let me know in case you have comments.


catalinristea:
Ciao @chiarazampolli,

(a)Looks good, but I did checkout your fork and I still see no executable flag for the scripts:
DATA/production/configurations/2021/ctf_recreation/ctf_recreation.sh
DATA/production/configurations/2021/OCT/apass4/async_pass.sh
O2DPG/DATA/production/qc-workflow.sh

The last one is not really used anywhere, but nonetheless...

(b) Also, a couple of setev_extra.sh are being marked as executables (with "#!/bin/bash") without executable flag, but these are **sourced**, not **executed**, so we're also fine. This is then just academic matter, either give-up "#!/bin/bash" or make these executable also.

For examply, ctf_recreation/setenv_extra*.sh are fine with respect to this aspect.

DATA/production/configurations/2021/OCT/apass3/setenv_extra.sh
DATA/production/configurations/2021/OCT/apass3_zerofield/setenv_extra.sh
DATA/production/configurations/2021/OCT/apass4/setenv_extra.sh

Also, I can do it myself after the inner changes to the scripts themselves are merged, no need to worry about all these.

Thanks a lot, Chiara!

Yours,
Catalin.

chiarazampolli:
Hi @catalinristea ,

Thanks for checking! The commit changed the permissions:
```
100644 → 100755 
```
so I thought it was correct. 
The qc-workflow.sh, though, is maintained by @davidrohr , and it is called by dpl-workflow.sh which is in O2. We should leave as it is. 
As for the setenv, I will fix it, better to have them clean.

catalinristea:
Ciao @chiarazampolli,

No, I think you did it perfectly, but I'm not skilled enough in github. I think you have a branch for these executables, but when I'm checking this out I get the master. Sorry for the noise, let me try first to figure this out and report back properly for -x flags.

Thanks!

Yours,
Catalin.

chiarazampolli:
Done, scripts updated. 
But for the permissions, I see them correct locally. 
I will merge this now, then we will see. 

Thanks, @catalinristea !!

PR #297, opened 2022-03-24
MCH: introduce QC_MCH_EXPERT env var for more QC plots

aphecetche:


DATA/production/qc-workflow.sh
```diff
@@ -15,6 +15,7 @@ if [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
     [[ -z "$QC_JSON_FT0" ]] && QC_JSON_FT0=/home/afurs/O2DataProcessing/testing/detectors/FT0/ft0-digits-ds.json
     [[ -z "$QC_JSON_FV0" ]] && QC_JSON_FV0=/home/afurs/O2DataProcessing/testing/detectors/FV0/fv0-digits-ds.json
     [[ -z "$QC_JSON_EMC" ]] && QC_JSON_EMC=consul://o2/components/qc/ANY/any/emc-qcmn-epnall
+    [[ -z "$QC_JSON_MCH" ]] && [[ x"$QC_MCH_EXPERT" != "x" ]] && QC_JSON_MCH=consul://o2/components/qc/ANY/any/mch-qcmn-epn-digits-expert
```
davidrohr, line 18: Hi Laurent, in principle I am fine with this, just 2 comments:
- instead of `[[ x"$QC_MCH_EXPERT" != "x" ]]` you can use `[[ ! -z "$QC_MCH_EXPERT" ]]`
- How are you using this actually? If you enable it in the expert panel extra env variable `QC_MCH_EXPERT=1`, you could as well just set `QC_JSON_MCH=consul://...` in there? Just wondering how often it is needed. Again, in principle it is fine, but I'd try to keep the production files clean of special expert settings that can also be steered by external variables.

aphecetche: @davidrohr concerning your first comment (depending on the outcome your first comment might become irrelevant). Actually yes, we intended to use it from the expert panel. The usage frequency will depend on how frequent issues are with our detector :wink:

But indeed, setting the `QC_JSON_MCH` from there is also an option. A simpler one (must confess it simply did not occur to me, as I blindly jumped to modify the script...). 
It's longer to type for sure and as such more error prone though. But with a good doc for our experts on our side I guess that's doable : @aferrero2707 what's your opinion ?

aferrero2707: For me both options look ok. In any case the extra ENV variable will be "wrapped" into an AliECS configuration, such that the experts will not have to type it each time. Hence the complexity of what needs to be typed is relatively irrelevant.

As far as I understand the solution proposed by David does not require modifications to the `qc-workflow.sh` script, right? Hence we could test it easily at P2?

davidrohr: Yes, but the advantage is that using the expert panel is that it is also more flexible, since you can select different JSON files from consul / epn home every time. If you store it as saved configuration, it is not so complicated anyway. I think EMC is already doing it that way.
And yes, the 2 suggestion would mean no change to `dpl-workflow.sh` is needed. You just put
```
QC_JSON_MCH='consul://o2/components/qc/ANY/any/mch-qcmn-epn-digits-expert'
```
in the expert panel (not 100% about quotation).

aphecetche: @aferrero2707 yes, basically instead of setting `QC_MCH_EXPERT=1` in the extra env var text entry, we would simply set `QC_JSON_MCH=consul://o2/components/qc/ANY/any/mch-qcmn-epn-digits-expert` there

aphecetche: @davidrohr was faster to type :wink: 
I guess we just have to test how/if to quote the variable from the AliECS UI then

aferrero2707: I propose to do a quick test as soon as possible to check that setting `QC_JSON_MCH` works in our case, and then for me it is fine.

I have already modified the `mch-physics-expert` configuration accordingly.

aphecetche:
Using `QC_JSON_MCH=consul://o2/components/qc/ANY/any/mch-qcmn-epn-digits-expert` was tested and is working just fine, so we close this PR as 'no longer needed'.

PR #298, opened 2022-03-24
Removed explicit IR argument, proc set to cdiff

catalinristea:
Ciao @chiarazampolli and @sawenzel,

Removed setting the IR explicitly, plus changed the proc to cdiff, to include UD collisions.
Please review and merge if you think these are appropriate.

Thanks!

Yours,
Catalin.

### DIFF
--- a/MC/run/ANCHOR/2021/OCT/pass4/anchorMC.sh
+++ b/MC/run/ANCHOR/2021/OCT/pass4/anchorMC.sh
@@ -84,8 +84,7 @@ NSIGEVENTS=${NSIGEVENTS:-22}
 baseargs="-tf ${NTIMEFRAMES} --split-id ${ALIEN_JDL_SPLITID:-0} --prod-split ${ALIEN_JDL_PRODSPLIT:-100} --run-number ${RUNNUMBER}"
 
 # THIS NEEDS TO COME FROM OUTSIDE
-remainingargs="-eCM 900 -col pp -gen pythia8 -proc inel -ns ${NSIGEVENTS}                                                                                                 \
-               -interactionRate 2000                                                                                                                                      \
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Fine with me, I let @sawenzel merge in case he has further comments.

sawenzel:
Why do want to take away the interaction rate? (I discussed with @chiarazampolli and we agreed to keep this one as there is no effect in going smaller).

MC/run/ANCHOR/2021/OCT/pass4/anchorMC.sh
```diff
@@ -84,8 +84,7 @@ NSIGEVENTS=${NSIGEVENTS:-22}
 baseargs="-tf ${NTIMEFRAMES} --split-id ${ALIEN_JDL_SPLITID:-0} --prod-split ${ALIEN_JDL_PRODSPLIT:-100} --run-number ${RUNNUMBER}"
 
 # THIS NEEDS TO COME FROM OUTSIDE
-remainingargs="-eCM 900 -col pp -gen pythia8 -proc inel -ns ${NSIGEVENTS}                                                                                                 \
-               -interactionRate 2000                                                                                                                                      \
```
sawenzel, line 88: we cannot just take it way. There would be no default then. Something like `-interactionRate ${IRATE:-2000}` might be ok.

catalinristea: Hi Sandro,

Ok, agreed not to remove the interaction rate (it was intended to be retrieved from CCDB, but we're not yet there). Will open a new PR, this one cannot be continued, as I have not pushed it from a branch of my fork.

Thanks!

Yours,
Catalin.

PR #301, opened 2022-03-25
add TOF cal in global workflow

noferini:
I created a new global production workflow for AliECS envirnoment where I added the first TOF sync-calibration prototype (tested in standalone mode).

The idea is to define here a variable (TOF_PROXY_OUTSPEC) to be used in O2(to be done) to set output proxy in the proper way.

Since there are no other example I did it in this way but if you prefer I do it in a different way let me know.

### DIFF
--- a/DATA/production/configurations/2021/OCT/apass5/async_pass.sh
+++ b/DATA/production/configurations/2021/OCT/apass5/async_pass.sh
@@ -1,0 +1,188 @@
+#!/bin/bash
+
+# Script to run the async processing
+#
+# if run locally, you need to export e.g.:
+#
+# export ALIEN_JDL_LPMRUNNUMBER=505673
+# export ALIEN_JDL_LPMINTERACTIONTYPE=pp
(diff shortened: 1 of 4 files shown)


noferini:
@davidrohr 
I would assume you review this PR (:))

davidrohr:
Hi @noferini , thx, I see the point of this PR that we need to distinguish between the workflow with and without aggregator. I need to think about how to generalize this best, but just returned from vacation today and a lot of things have piled up, I'll reply some time next week. If you need it urgently, we can also merge this and adapt later if needed.

noferini:
Hi David,
it is not urgent. I was trigger by P2 people which asked us when we think to port calibration in global runs.
I agree that we need to find the best general way. No problem for me to wait, I just wanted to start the discussion.

chiarazampolli:
Should this not go here:
https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/calib-workflow.sh
?

noferini:
I close, I'll reopen once ready

PR #303, opened 2022-03-29
Query the run timerange from GRPECS

sawenzel:
It seems to be more reliable than taking this from RCT/Information.
And we can do it following https://github.com/AliceO2Group/AliceO2/pull/8459

Some accompanying features:
* ability to get CCDB object as JSON for direct treatment in python


sawenzel:
I checked that with this, all run numbers of the pilot beam run ok with anchored MC.

PR #305, opened 2022-03-29
ITS calib: added chip-mod-selector and chip-mod-base + set default to 40 EPNs

iravasen:
This goes with O2 PR8471 (https://github.com/AliceO2Group/AliceO2/pull/8471). In addition:

- refinements for o2-qc 
- 20 -> 40 EPNs (based on last tests at P2)


iravasen:
@chiarazampolli this is based on the discussion with Ruben on ITS Threshold JIRA to add chip-mod-selector and chip-mod-base to the ITS threshold calibration workflow.

chiarazampolli:
Ok, thanks! Merging.

PR #306, opened 2022-03-30
MCH: get the onlyAtEndOfStream option be driven by an AliECS env var …

aphecetche:
…(MCH_END_OF_STREAM_ONLY)

This is to workaround the fact that the end-of-stream seems to be broken again.

### DIFF
--- a/DATA/production/calib/mch-badchannel-aggregator.sh
+++ b/DATA/production/calib/mch-badchannel-aggregator.sh
@@ -18,7 +18,8 @@ ARGS_ALL_CONFIG="NameConf.mDirGRP=$FILEWORKDIR;NameConf.mDirGeom=$FILEWORKDIR;Na
 PROXY_INSPEC="A:MCH/PDIGITS/0"
 CONSUL_ENDPOINT="alio2-cr1-hv-aliecs.cern.ch:8500"
 
-BADCHANNEL_CONFIG="${ARGS_ALL_CONFIG};MCHBadChannelCalibratorParam.maxPed=200.0;MCHBadChannelCalibratorParam.maxNoise=2.0;MCHBadChannelCalibratorParam.minRequiredNofEntriesPerChannel=100;MCHBadChannelCalibratorParam.minRequiredCalibratedFraction=0.5;MCHBadChannelCalibratorParam.onlyAtEndOfStream=true"
+MCH_END_OF_STREAM_ONLY=${MCH_END_OF_STREAM_ONLY:-true}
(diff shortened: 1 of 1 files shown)


ktf:
Do you have a printout in readout-proxy for the end of stream being received?

aphecetche:
something like this you mean ? 

(from `ssh epnlog@epn034 cat /tmp/wn_dds/7568f8b8-ff4a-444b-86e5-69d894671a58/wn_online_epn034-ib.internal/readout-proxy_reco2_2022-03-30-10-12-18_16894768857233317650_out.log`) 

```
[10:19:00][ERROR] device state change is requested, dropping 121600 pending message(s)
on channel from_readout-proxy_to_mch-pedestal-decoder
ATTENTION: DATA IS LOST! Could not dispatch data to downstream consumer(s), check if
consumers have been terminated too early
[10:19:00][INFO] Processing timeslice:5503, tfCounter:5503, firstTFOrbit:0, action:0
[10:19:00][ERROR] device state change is requested, dropping 121600 pending message(s)
on channel from_readout-proxy_to_mch-pedestal-decoder
ATTENTION: DATA IS LOST! Could not dispatch data to downstream consumer(s), check if
consumers have been terminated too early
[10:19:00][INFO] Processing timeslice:5504, tfCounter:5504, firstTFOrbit:0, action:0
[10:19:00][ERROR] device state change is requested, dropping 121600 pending message(s)
on channel from_readout-proxy_to_mch-pedestal-decoder
ATTENTION: DATA IS LOST! Could not dispatch data to downstream consumer(s), check if
consumers have been terminated too early
[10:19:00][INFO] Processing timeslice:5505, tfCounter:5505, firstTFOrbit:0, action:0
[10:19:00][ERROR] device state change is requested, dropping 121600 pending message(s)
on channel from_readout-proxy_to_mch-pedestal-decoder
ATTENTION: DATA IS LOST! Could not dispatch data to downstream consumer(s), check if
consumers have been terminated too early
```

chiarazampolli:
We tested the MID calib this morning, and it worked, @dstocco .

ktf:
this actually looks like you have huge backpressure on the channels… Is the
channelconfig configured properly?

Ciao,
Giulio

ktf:
No, it should say "Sending end of stream to"

Ciao,
Giulio

aphecetche:
@ktf @chiarazampolli indeed the end-of-stream is not the culprit here. We were testing with (most of) our FLPs (while previously we used only a couple) and were not requesting enough EPNs (or have enough raw decoders). 

Increasing the number of EPNs (and the raw decoder multiplicity factor) did result in the calibration working as before. 

Sorry for the false alarm.

PR #311, opened 2022-04-01
Remove DDS session ID for fetcher to work correctly

chiarazampolli:
@martenole


davidrohr:
Setting the DDS session id might have side effects, since it is used in some hacks to determine whether something runs on the EPN or not.

chiarazampolli:
Hello @martenole , @davidrohr , 

Thanks for the feedback!
I have no idea why this was added, so now it is removed, and everything should be ok. 
For apass3, I will leave it, since we trace here the script that we used.

Chiara

PR #315, opened 2022-04-05
Make analysis fully modular

benedikt-voelkel:
* o2dpg_analysis_test_workflows wraps the analysis side of things now
  It can be run
  1. as script where workflow file with only the analysis tasks is
     created
  2. as a module to be used inside another workflow generation

  ==> analyses only need to be maintained in ine place

* no analyses directly implemented in o2dpg_sim_workflow

* changes in o2dpg_sim_workflow fully transparent, --include-analysis
  works as before and expected

* change os.mkdir to os.makedirs in o2_dpg_workflow_runner to avoid the
  necessity of the user to create certain directories if the soecified
  cwd of a tasks is deeper than one directory

### DIFF
--- a/MC/analysis_testing/analysis_test.sh
+++ b/MC/analysis_testing/analysis_test.sh
@@ -21,14 +21,8 @@ NTF=$(find ./ -name "tf*" -type d | wc | awk '//{print $1}')
 # RC1=$?
 # echo "EXIT 1: $RC1"
 
-# run on the merged part
-wf_name="workflow_test_analysis.json"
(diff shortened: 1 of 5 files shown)


benedikt-voelkel:
@sawenzel @chiarazampolli can we merge? it would be nice to have the changes in tonight so we can also test them on alibi.
I can do it as well if you agree/give approval

chiarazampolli:
For me yes! Without it, analysis currently does not work. 
@sawenzel ?

PR #325, opened 2022-04-11
add TOF cal

noferini:
TOF calibration included now

o2calib_tof.root generated

tested with 1 CTF  (66 TFs) from pilot beam

o2calib_tof.root size = 0.5% AO2D size


chiarazampolli:
Ciao @noferini , 

Why is this for apass5 and not apass4, which has not run yet?

noferini:
Because it was never tested so far and I didn't want to break anything.
I will ask Catalin to run a test and in case I will propagate the same change also for apass4

chiarazampolli:
Since we are still preparing for apass4, and we have no ETA for apass5, I would do it in apass4.
Note that tests currently cannot have QC enabled, but yes, we could do one with apass4.

noferini:
now done for apass4

PR #327, opened 2022-04-12
Replace config parameter with array of configs

fgrosa:
@preghenella when you have time could you please check that I correctly changed all the occurrences? Thanks a lot!

Please note that the code in this PR needs its counterpart in O2 https://github.com/AliceO2Group/AliceO2/pull/8567 to work.


chiarazampolli:
@fgrosa , should I wait that the PR in O2 is merged? Maybe it is better...

chiarazampolli:
Exactly. 
Please, ping me again in case I don't see the other PR merged.

PR #332, opened 2022-04-17
enable TOF digit histos to avoid crash

noferini:


noferini:
Accordingly to a suggestion of @njacazio I added a parameter to enable the creation of some histos (via json file).
This should avoid the current crash fixed in https://github.com/AliceO2Group/QualityControl/pull/1126 but not yet available when starting a run in AliECS (O2DPG can be immediately ported there, if I am not wrong)
@davidrohr @Barthelemy @knopers8 

The same fix already copied also in
/home/fnoferin/public/tof-qc-globalrun.json
for consistency
Same changes applied in consul for consistency (even if probably not needed there)

davidrohr:
@noferini : I have already updated QC on the EPNs to 1.54.1, which contains the fix. Is this change still needed in this case?

noferini:
No, it is no longer needed. Thanks

PR #345, opened 2022-04-21
Scripts to compute size and compression for a list of AO2D.root trees.

mgheata:
These scripts depend only on root and allow computing the size of the different AO2D tables and their compression on file.
Run as:
perl size_ao2d.pl [file1] [file2] ...
counttracks.sh [file1] [file2] ... utility to count number of collisions, bc's and tracks in a list of AO2D files
file1, file2, ... have to be local or remote AO2D.root files
remote file locations should be specified as: alien:///.../AO2D.root


sawenzel:
I would prefer having this under some more general top-level directory. So far we have "MC", "DATA", "UTILS".
Maybe "UTILS" sounds appropriate or we create a new one depending on the final use case. What do you think?

mgheata:
> I would prefer having this under some more general top-level directory. So far we have "MC", "DATA", "UTILS".
> Maybe "UTILS" sounds appropriate or we create a new one depending on the final use case. What do you think?

@sawenzel I moved the folder under UTILS

chiarazampolli:
Merging, this does not affect the FST. 
Since @sawenzel is away this week,in case he has further comments, they can still be addressed when he is back, but I prefer to have the code in O2DPG already.
Thanks @mgheata !

PR #346, opened 2022-04-24
updated version of the script which will be used for release validation

emeninno:


RelVal/ReleaseValidation.C
```diff
@@ -216,18 +248,56 @@ void ProcessMonitorObjectCollection(
     } else if (clname.BeginsWith("TH")) {
       TObject *o = (TObject *)o2MonObjColl->FindObject(olname.Data());
       WriteHisto(o);
+      Counts++;
```
chiarazampolli, line 251: Can you make this starting with a small "c"? It is a variable. I spotted this only with capital initial, if there are others, can you change them too?

RelVal/ReleaseValidation.C
```diff
@@ -9,28 +9,30 @@ TFile *fileSummaryOutput = nullptr;
...
+  bool e;
 };
 
-// define "a monte" the possible avaiable tests
+// define "a monte" the possible available tests
 enum options {
   Chi2    = 0x01,
   BinCont     = 0x02,
-  Nentries   = 0x04,
+  BinContInt   = 0x04, //BinContInt is similar to BinCont, but the bin content is normalized by the integral
```
benedikt-voelkel, line 27: I would then call that maybe `BinContNorm`?

RelVal/ReleaseValidation.C
```diff
@@ -9,28 +9,30 @@ TFile *fileSummaryOutput = nullptr;
...
   Chi2    = 0x01,
   BinCont     = 0x02,
-  Nentries   = 0x04,
+  BinContInt   = 0x04, //BinContInt is similar to BinCont, but the bin content is normalized by the integral 
   // ...
 };
 
 
-void ProcessFile(TString fname, TString dirToAnalyse); // needs to be declared
+void ProcessFile(TString fname, TString dirToAnalyse);
```
benedikt-voelkel, line 32: of course you can use `TString` here. In case it's only accessed to be read and not modified, I would make it `const&`

RelVal/ReleaseValidation.C
```diff
@@ -216,18 +248,56 @@ void ProcessMonitorObjectCollection(
...
+    printf("****** %s (class %s)   ******\n", olname.Data(), clname.Data());
+    if (clname == "TDirectoryFile") {
+      TDirectoryFile *dirObj =
+	(TDirectoryFile *)dirCollect->Get(olname.Data());
+      dirObj->ls();
+      ProcessDirCollection(dirObj);  
+    } else if (clname.BeginsWith("TH")) {
+      printf("--- Process histograms in %s ---\n", dirname.Data());
+      printf("--- %s ---\n", olname.Data());
+      TObject *o = (TObject *)dirCollect->Get(olname.Data());
```
benedikt-voelkel, line 286: When you do the `->Get(...)` it should be  a `TObject` by default, right? I guess in that case we can omit the cast to `TObject` here

RelVal/ReleaseValidation.C
```diff
@@ -501,18 +599,19 @@ void CompareHistos(TH1 *hA, TH1 *hB, TString monobj, int whichTest, double valCh
   more->SetBorderSize(1);
   more->AddEntry((TObject*)0, Form("#chi^{2} / Nbins = %f", testResult.b),"");
```
benedikt-voelkel, line 600: as the first argument, I would suggest to use `nullptr` at these places

benedikt-voelkel:
Dear @emeninno 
thanks for this update! I also added a few comments to this PR, all in principle minor. Let me know if you have questions.

Another thing which we should remember to take care of in the near future: In some analysis output, there are histograms stored in a `TList` for instance. We need to make sure that we also process those objects. There are some more generic ways to do something like that which we can discuss after this PR has been merged. So this comment is to make sure we keep that in mind.

chiarazampolli:
Hello @emeninno , 

I am merging, so that @benedikt-voelkel can start the integration with python and alibi discussed yesterday. Please, open a new PR with the comments in this one - they are mainly style-related. 

Thanks!

Chiara

PS: this PR does not affect the FST.

PR #349, opened 2022-04-27
Update for apass4

chiarazampolli:
- aligned geometry should not be a link but passed a input (till
CCDB access will be finalized)
- add " --run-assessment" for MFt reco

@catalinristea @shahor02 @rpezzi


chiarazampolli:
Not interfering with FST, merging.

PR #351, opened 2022-04-28
Avoid complete randomness by default

mpuccio:
This is helpful to try to reproduce locally things that happen on the grid

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python3
 
 #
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
Actually, what we might want to do is setting different seeds for different TFs. Because atm the same seed is set for each TF and so event generation is "the same" for all of them. Would that make sense?

mpuccio:
Very good point, I added the TF number to the default seed with the new patch. Benedikt, Chiara, let me know if you have any other comments

chiarazampolli:
Seems fine to me. @benedikt-voelkel , any further comment?

benedikt-voelkel:
It looks good to me @chiarazampolli @mpuccio

PR #354, opened 2022-04-29
Revert "Avoid complete randomness by default"

benedikt-voelkel:
Reverts AliceO2Group/O2DPG#351

The seeds don't work and the simulation crashed with 
```bash
PYTHIA Error: variable recognized, but its value not meaningful:
   Random:seed = 6025226150559117445
  ```

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -20,7 +20,7 @@
 import sys
 import importlib.util
 import argparse
-from os import environ, mkdir, getcwd
+from os import environ, mkdir
 from os.path import join, dirname, isdir
 import json
 import itertools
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
@chiarazampolli @davidrohr @shahor02 could you merge this?

benedikt-voelkel:
Thanks for the quick action!
Some numbers seem to be too large probably or invalid for some other reason. I will have a look.

chiarazampolli:
Maybe we could use the time in ms?

chiarazampolli:
Or truncate this.

benedikt-voelkel:
Yes, but Max' plan to make it semi-random, namely every simulation different but each single one reproducible. But actually the seed would be hard-coded in the `workflow.json`, so if the seed is fixed once, we can always re-use the workflow.
So we could use ms or anything else actually, what do you think @mpuccio @chiarazampolli ?

chiarazampolli:
What matters is to keep track of it, 0 would have not allowed it but otherwise, whatever we do, we can print it.

benedikt-voelkel:
Sounds good!
No, indeed as long as we have the number, it should be fine, To me it would be important to have method that can cover different generators consistently. I will think about it

mpuccio:
I think we should anyway avoid the time, because log files are not easily programmatically accessibile. By the way, with further investigations we saw with @fgrosa that we need to modify deeper to propagate the seed to simulation with custom Pythia8 configurations. So we will come up with a better solution that does not rely on time and that we test a bit more, sorry for causing the issue with the tests!

benedikt-voelkel:
ok, then it sounds like you will take over the implementation?
Thanks for the investigation!

PR #360, opened 2022-05-03
Apply O2 clang-format

benedikt-voelkel:
In preparation of adding clang-format and whitespace/tab check CI as
implemented for O2 to all C-like scripts


benedikt-voelkel:
As discussed with @chiarazampolli @sawenzel and @TimoWilken I applied `clang-format` according to what we have in O2 in preparation to include the formatting of C-like scripts also in O2DPG

davidrohr (approved):

PR #361, opened 2022-05-03
Fix regex on macOS

ktf:


ktf:
`(,|)` is actually invalid. What I guess you want is actually `,?`, which means 0 or 1 matches.

davidrohr:
afaik `|` is an optional character in sed. `?`should do the same, but haven't tried.
Can you add a removal of the created `.bak` files?

ktf:
They are not the same `|` is the "OR" separator, while `?` is really "matches 0 or 1". The problem is that `(,|)` means `,` OR and empty regex, and that's invalid (`RE error: empty (sub)expression`) for bsd coretools.

PR #362, opened 2022-05-03
Fix mishandling of TMPDIR

ktf:
TMPDIR is used on many system for the user's temporary dir. Overriding
it and deleting has unwanted consequence that any subsequent usage might
fail, e.g. on macOS this happens when trying to create FairMQ channels
which are backed by unix domain sockets in $TMPDIR. I suspect ROOT would
also have troubles with it for some operations, even on linux.


sawenzel (approved):

ktf:
Error unrelated. Tested locally. Merging.

ktf:
Error unrelated. Tested locally. Merging.

PR #365, opened 2022-05-04
Rework file reading and object extraction

benedikt-voelkel:
* rename some functions to give a better idea what they are doing

* always compare all found objects

* remove code duplications

* make a user to specify the whole file path, fon't assume or add
  ".root" file extension

* same for files on the GRID: user has to give full GRID path. When path
  begins with ""alien", assume it is on the GRID and initialise

* remove pointers where not really needed

* use same extraction function for any ROOT input file, don't use
  specific ones for QC ROOT files and others

* wrap object extraction and plotting as much as possible

No changes have been made wrt the actual comparison and comparison
summary

* add very small python wrapper for convenience

### DIFF
--- a/RelVal/ReleaseValidation.C
+++ b/RelVal/ReleaseValidation.C
@@ -207,7 +207,9 @@ bool PotentiallySameHistograms(TH1* hA, TH1* hB)
     return false;
   }
 
-  if (!PotentiallySameAxes(hA->GetXaxis(), hB->GetXaxis()) || !PotentiallySameAxes(hA->GetYaxis(), hB->GetYaxis()) || !PotentiallySameAxes(hA->GetZaxis(), hB->GetZaxis())) {
+  if (!PotentiallySameAxes(hA->GetXaxis(), hB->GetXaxis()) ||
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
Dear @emeninno 
I have re-worked the file reading and input processing a bit to unify some functionality etc (see PR description).
Please take a look and let me know if there is anything that looks weird or wrong to you.

As I said I have not touched the code which does the actual comparison.

benedikt-voelkel:
Another comment: I took out the check for "identical histograms".
Imo we should leave it to the user which files should be compared and it should not be assumed that something is equal by the code.
Also, that check didn't seem to be quite well defined to me:
* sure, the histograms could have the same number of bins in x, y and z but technically, they could have different bin edges at the same time.
* the comparison of the bin contents `cA` and `cB` which was `if (TMath::Abs(cA - cB) > 0.001 * TMath::Abs(cA))` should not be relative to either `cA` or `cB` but to one fixed value of resolution parameter
But please correct me here. I think we can discuss that as well

chiarazampolli:
Hi @benedikt-voelkel ,
Thanks! I would merge this one now, and in case implement further features later, with also the comments by @emeninno and Daniel (I don't know his github account, I will ping him by mail).

Chiara

chiarazampolli:
Does not impact FST, merging.

PR #366, opened 2022-05-04
Updating forward Jpsi in pp scripts

mcoquet642:
Updating beam energy in forward scripts to 13.6 TeV. Prompt script updated to generate only a single J/psi per event.


gconesab (approved):
It looks good to me

mcoquet642:
Thanks for the feedback. Is there something preventing this from being merged ?

PR #367, opened 2022-05-04
Seed propagation for injected productions

mpuccio:
Hi @benedikt-voelkel, @chiarazampolli, @sawenzel, @fgrosa,

how it was done before the seed was not propagated to the signal generator if a custom configuration of pythia8 was passed in the ini file. What I propose here is that we get a random number starting from the jobid using the python random number generator and then we increase it by 1 for each TF. The maximum seed is truncated to fit the Pythia8 requirements.
Then this number is given as general seed to o2-sim, that already gives it to Pythia and it will be the same we set from outside once AliceO2Group/AliceO2#8727 will be merged.

In the case of the HF injection we were also specifying the process in the bash script, thus leading to the on-the-fly generation of a pythia configuration that was overwriting the HF defaults (it was then hacked back in via the confKeyValues). Now this should be solved and I added the support for the SPLITID as seed in the HF bash script.

If you have any comments or feedback, it is very welcome!
Cheers,
Max

### DIFF
--- a/MC/run/PWGHF/run_pp_HFtriggers_bbbar_userhook.sh
+++ b/MC/run/PWGHF/run_pp_HFtriggers_bbbar_userhook.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #
-# A example workflow MC->RECO->AOD for a simple pp min bias 
+# A example workflow MC->RECO->AOD for a simple pp min bias
 # production, targetting test beam conditions.
 
 # make sure O2DPG + O2 is loaded
@@ -11,7 +11,7 @@
 # ----------- LOAD UTILITY FUNCTIONS --------------------------
 . ${O2_ROOT}/share/scripts/jobutils.sh
 
-# ----------- START ACTUAL JOB  ----------------------------- 
+# ----------- START ACTUAL JOB  -----------------------------
 
 NWORKERS=${NWORKERS:-8}
 MODULES="--skipModules ZDC"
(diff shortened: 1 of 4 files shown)


fgrosa:
Hi @mpuccio thanks a lot for this implementation, it looks very good for me! 👍

MC/run/PWGHF/run_pp_HFtriggers_bbbar_userhook.sh
```diff
@@ -11,24 +11,21 @@
...
 NSIGEVENTS=${NSIGEVENTS:-1}
 NBKGEVENTS=${NBKGEVENTS:-1}
 NTIMEFRAMES=${NTIMEFRAMES:-1}
-
+SPLITID=${SPLITID:-None}
 # create workflow
 
 #ccbar filter
-${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow.py -eCM 13500 -col pp -gen pythia8 -proc "inel" -j ${NWORKERS} -ns ${NSIGEVENTS} -tf ${NTIMEFRAMES} -interactionRate 500000 -confKey "Diamond.width[2]=6.;GeneratorPythia8.config=${O2DPG_ROOT}/MC/config/PWGHF/pythia8/generator/pythia8_charmtriggers_with_decays.cfg" -e TGeant4 -mod "--skipModules ZDC" \
+${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow.py -eCM 13500 -col pp -gen pythia8 -j ${NWORKERS} -ns ${NSIGEVENTS} -tf ${NTIMEFRAMES} -interactionRate 500000 -confKey "Diamond.width[2]=6.;" -e ${SIMENGINE} -seed ${SPLITID} -mod "--skipModules ZDC" \
```
benedikt-voelkel, line 26: Would that be interpreted as a string by `argparse`? And does the seeding of python's `random` work with that string of `"None"`?
Same in `MC/run/PWGHF/run_pp_HFtriggers_ccbar_userhook.sh`

benedikt-voelkel:
Thanks @mpuccio !
That looks to me as if it should work now. I just have a brief comment/question above.
Eventually, we need to find a way to make it generic as soon as we start to actually use generators other than pythia8. But we can put that on the agenda

MC/run/PWGHF/run_pp_HFtriggers_bbbar_userhook.sh
mpuccio: Very good point! I updated this part of the code to avoid passing the seed if SPLITID is not set

PR #371, opened 2022-05-05
FIX clang-format took apart some include paths

benedikt-voelkel:


benedikt-voelkel:
@chiarazampolli This needs to be merged so that custom evt generator macros work again. `clang-format` was a bit funny there before

chiarazampolli:
Not used in FST, merged.

PR #372, opened 2022-05-05
If we redirect the QC remotes, make sure we also don't use the production QCDB

davidrohr:


costing:
Hi @TimoWilken,

Thanks, that was indeed the case. `service autofs restart` solved it.

Cheers,

.costin

davidrohr:
@TimoWilken @costing : This is from this morning at 3:45:
```
[INFO] Is alien token present?: 0
[INFO] MC-TRUTH 0
```
Still not working

PR #373, opened 2022-05-06
adding calib-workflow

chiarazampolli:
@davidrohr , @shahor02 

@davidrohr : if you want, the function `add_semicolon_separated()` can be incorporated in `add_common_separated()` in a more general one where one passes also the separator.

Needed by https://github.com/AliceO2Group/AliceO2/pull/8736

### DIFF
--- a/DATA/common/setenv.sh
+++ b/DATA/common/setenv.sh
@@ -225,3 +225,22 @@ add_comma_separated()
     fi
   done
 }
+
+add_semicolon_separated()
+{
+  if (( $# < 2 )); then
+    echo "$# parameters received"
+    echo "Function name: ${FUNCNAME} expects at least 2 parameters:"
+    echo "it concatenates the string in 1st parameter by the following"
+    echo "ones, forming semi-colon-separated string. $# parameters received"
+    exit 1
+  fi
+
+  for ((i = 2; i <= $#; i++ )); do
+    if [[ -z ${!1} ]]; then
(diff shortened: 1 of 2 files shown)


davidrohr (approved):

davidrohr:
> Hi @davidrohr .
> 
> This setting work locally. Maybe it is good enough to start with?
> 
> Chiara

Sure, fine with me, please merge it when you think it is ready

chiarazampolli:
I will merge it, as it is not used in the FST (the extra function in setenv.sh is not called there). 
@davidrohr : can I then remove the WIP from the PR in O2, or should one wait that O2DPG is tagged? I mean, is the CI in O2 running with master of O2DPG?

davidrohr:
no need to wait for a tag, just merge here, and then remove the WIP in O2.

chiarazampolli:
Done, thanks for the feedback.

TimoWilken:
> @TimoWilken : The CI with the FST does only show up after a while, not immediately when the PR is created. Do you know why?

This is because the check is not marked as "required". In this case, GitHub doesn't know the check exists, and only marks it as "pending" once the CI builder reports it for the specific commit for the first time. This can take a bit of time if the builder is busy at the time the PR is submitted.

If it is "required", then GitHub marks it as "expected" until the CI builder reports it for the first time, whereupon it changes from "expected" to "pending".

PR #374, opened 2022-05-06
Add track prop, add analysis for AOD check

benedikt-voelkel:


benedikt-voelkel:
@sawenzel @chiarazampolli 
Finally fixing the analyses with track propagation **and** adding the `o2-analysis-check-data-model-mc`

benedikt-voelkel:
I think the analysis implementation should be corrected in the python script. And only when necessary they should be adjusted accordingly by experts.

Imho We should keep the analysis workflow pipes in the script explicit and not tweak them via env variables.

I feel like analyses might still change a bit (some implementation details, additional tasks needed etc). So if we now introduce certain things via env vars, it could become harder to know how the final pipes will look like.

What do you think @chiarazampolli ?

benedikt-voelkel:
@sawenzel @chiarazampolli 
Could we merge this before today's nightlies? Otherwise we will have analysis crashes everywhere again

PR #375, opened 2022-05-06
add TRD calib and fixes

chiarazampolli:
Adding TRD @martenole , @davidrohr


chiarazampolli:
Tested with async reco, and it works (or at least it does not crash). But TRD is not in for the async reco now.

chiarazampolli:
Merging this one, to go on with the one in O2.

PR #376, opened 2022-05-06
[QC-779] Allow to run QC postprocessing at the end of MC, show how

knopers8:
This adds some helpers to run QC postprocessing at the end of MC productions and shows how to integrate a postpostprocessing workflow on the example of `tofTrendingHits`. The task is executed correctly and does not crash the workflow, but it is missing valid configuration (trended object path), which will be fixed by @ercolessi


chiarazampolli:
Hi @knopers8 , 

Should I merge this?

BTW, this is for post-processing at run level, right? 

Chiara

knopers8:
> Should I merge this?

If you do not disagree, then yes please :)

> BTW, this is for post-processing at run level, right?

Both. As I mentioned during the meeting, if we want to trend a production, we basically invoke this at the end of processing of each run, then we trend all the objects again and update the trend. You would use `runSpecific=False, prodSpecific=True` in such case.

chiarazampolli:
Ok, thanks. 
We will need to see how to process all runs from a production, to have the post-processing over different runs.

chiarazampolli:
Does not impact FST, merging.

knopers8:
> Ok, thanks. We will need to see how to process all runs from a production, to have the post-processing over different runs.

Maybe I was not clear. This allows you to do that.

chiarazampolli:
It was clear, I referred to the grid setup.

knopers8:
There is no need to change anything on the grid, this is all we should need :)

chiarazampolli:
You need to loop over the output of the runs, write the script, JDL... Right?

ercolessi:
Hi @knopers8 here is the fix: https://github.com/AliceO2Group/QualityControl/pull/1172 !

knopers8:
No. Since QC postprocessing works on objects in QCDB, all we need is to set a post-processing trigger which will iterate on the latest object for each run in a produciton. This is exactly how this example is configured.

knopers8:
> Hi @knopers8 here is the fix: [AliceO2Group/QualityControl#1172](https://github.com/AliceO2Group/QualityControl/pull/1172) !

Thanks, I will propagate this to this repo.

chiarazampolli:
> No. Since QC postprocessing works on objects in QCDB, all we need is to set a post-processing trigger which will iterate on the latest object for each run in a produciton. This is exactly how this example is configured.

I probably don't understand your strategy: will this run on the grid or not?

knopers8:
[MC/bin/o2dpg_qc_finalization_workflow.py](https://github.com/AliceO2Group/O2DPG/pull/376/files#diff-84ea54685cdc75c5ac913f9e9a92de9bd1f3bee565a2f1e05de3aa0d9f052922) already runs on grid. I added QC postprocessing to that file, so it will run on grid, because it used to run on grid.

PR #377, opened 2022-05-06
Specify output path from python

benedikt-voelkel:
Avoid writing all files to current directory, so the Python wrapper can
be started from anywhere writing to a different place


chiarazampolli:
Not impacting FST

PR #379, opened 2022-05-06
update for apass4

chiarazampolli:
@catalinristea : can you add "analysisQC.log" in the output for apass4?
The changes here reflect some re-working of the production scripts also.


chiarazampolli:
Merging, FST not affected.

catalinristea:
Ciao Chiara,

This log would be already saved by the generic log wildcards, so nothing new is needed, but thanks for pointing it out - "log_archive.zip:std*,*.log,*.json@disk=1"

Yours,
Catalin.

chiarazampolli:
Hi Catalin, 

Great!

Chiara

PR #381, opened 2022-05-09
Fixing ctf_recreation script after modification of dpl-workflow etc

chiarazampolli:
@catalinristea


chiarazampolli:
Does not interfere with FST, merging.

PR #382, opened 2022-05-09
fix logic to run calibrations

chiarazampolli:
@martenole


chiarazampolli:
Not affecting FST, merging.

PR #385, opened 2022-05-10
Move MID bad channels calibration workflow in standalone-calibration.desc

dstocco:


chiarazampolli:
Not used in FST, merging.

PR #386, opened 2022-05-10
protections for sourcing files

chiarazampolli:


davidrohr:
@chiarazampolli : How is it happening that it is sourced twice?
Probably in your local tests where you run the dpl-workflow.sh and the aggregator as one workflow?
I assume in that case we have the same problem also with setenv.sh and with getCommonArgs.sh?
Perhaps we could just change the aggregator script to not source anything if `CALIB_LOCAL_AGGREGATOR=1`? (as suggested in https://github.com/AliceO2Group/AliceO2/pull/8736)
Or we have to add protection to the other scripts as well.
Although, in general I am also not principally against putting protections everywhere.

chiarazampolli:
I added protections everywhere (unless I forgot something), just for safety. 
The only things I did not touch is the dpl-workflow.sh, the setenv.sh and getCommonArgs.sh

PR #390, opened 2022-05-11
Use IPC for local calib connections

davidrohr:
@chiarazampolli : Haven't tested this yet, but this is what I had in mind for the proxy channels:
So for testing locally, but in 2 separate shells, i.e. without CALIB_LOCAL_INTEGRATED_AGGREGATOR, one has to set both CALIB_PROXIES and CALIB_LOCAL_AGGREGATOR. Then it will use SHM transport with a socket file based on the channel name.
If CALIB_PROXIES is set, but CALIB_LOCAL_AGGREGATOR is not, i.e. when we run online on the EPN, we use zeromq transport and leave the address empty, to be filled by FairMQ automatically, in the same way we are doing for the calib workflows at P2 right now already.

PR #392, opened 2022-05-13
Fix option to pass the name of the proxy/channel

chiarazampolli:
@davidrohr


chiarazampolli:
Trivial and tested, merging. See comment in #390 .

PR #393, opened 2022-05-13
MFT: New URL alias for the DCS-CCDB

tomas-herman:


chiarazampolli:
Hi @tomas-herman , 

I am merging but it should work also with the old alias. Was it not?

Chiara

tomas-herman:
Hi @chiarazampolli ,
Thanks, it worked perfectly fine with the old alias. We just got instructions from Mateusz that we should switch to this new one.
Cheers,
Tomas

PR #397, opened 2022-05-16
Speed improvements for MC workflow

sawenzel:
a) ability to combine DPL devices (via option)

b) setup ROOT environment variables to reduce ROOT startup overhead ("multiple short lived child-processes").
    This complements and uses https://github.com/alisw/root/commit/d6ea57cb6226f37e517bba241a546ee576ed236e


MC/bin/o2_dpg_workflow_runner.py
```diff
@@ -51,6 +51,7 @@
 parser.add_argument('--checkpoint-on-failure', help=argparse.SUPPRESS) # debug option making a debug-tarball and sending to specified address
                                                                        # argument is alien-path
 parser.add_argument('--retry-on-failure', help=argparse.SUPPRESS, default=2) # number of times a failing task is retried
+parser.add_argument('--root', help=argparse.SUPPRESS, action='store_true') # number of times a failing task is retried
```
chiarazampolli, line 54: Can you change the comment?

chiarazampolli:
All looks good to me, just the minor comment above.

Thanks @sawenzel !

MC/bin/o2_dpg_workflow_runner.py
sawenzel: Thanks for spotting. Fixed.

PR #401, opened 2022-05-17
Inject estimated resources into WF runner

benedikt-voelkel:
Central to this PR

* o2dpg_sim_metrics.py resources
  produces a resource estimate for mem and CPU based on previously
  derived metrics from one or more runs
  --take argument can be used to define to take
  * average
  * min
  * max
  in case multiple metric JSONs are passed. The estimates are written to
  a JSON again

  Treat every TF the same so the estimate file only contains
  "task_name": {
    "mem": <memory>,
    "cpu": <cpu>
  }
  without TF suffix

* o2_dpg_workflow_runner --update-resources
  takes a JSON produced by previous resource estimation and updates the
  resources on the fly
  Due to above comment, all TFs get the same resource estimate
  
  Other
  
  some consolidation and help for argparse


sawenzel:
Merging this to have the ability to inject fresh resource estimates into the scheduler.
Tested the principle and it works. The only thing is that the resource estimation from the pipeline_metrics might be too crude (since based on a 5s sampling), so I will provide another script that estimates them directly from the jobutils/taskwrapper output which is using /usr/bin/time.

PR #405, opened 2022-05-18
Include geometry file path in ITS json for DATA/MC

iravasen:


martenole:
Hi Ivan,
would it then not make sense to change the path for sync as well? https://github.com/AliceO2Group/O2DPG/blob/ddc539d16a12e62b615c79f1a917047ba7bc02a0/DATA/production/qc-sync/its.json#L42
Cheers,
Ole

DATA/production/qc-async/its.json
```diff
@@ -39,7 +39,8 @@
           "layer": "1111111",
           "nThreads": "1",
 	  "nBCbins" : "103",
-	  "dicttimestamp" : "0"
+	  "dicttimestamp" : "0",
+	  "geomPath" : "/home/epn/odc/files/"
```
chiarazampolli, line 43: Ciao @iravasen , this cannot work, we don't have the epn files. 
If this is o2sim_geometry-aligned.root, then it is local to the job, you can take it from there. See e.g. https://github.com/AliceO2Group/O2DPG/blob/ddc539d16a12e62b615c79f1a917047ba7bc02a0/DATA/production/qc-async/itstpctof.json#L47

MC/config/QC/json/its-clusters-tracks-qc.json
```diff
@@ -42,7 +42,8 @@
           "layer": "1111111",
           "nThreads": "1",
 	  "nBCbins" : "103",
-	  "dicttimestamp" : 0
+	  "dicttimestamp" : 0,
+	  "geomPath" : "/home/epn/odc/files/"
```
chiarazampolli, line 46: Same as above.

davidrohr:
Sorry, but this doesn't make sense in my opinion. The only case where we are guaranteed to run on the EPN is the EPNSYNCMODE=1, and there we should actually fetch the JSON from consul.
The QC JSONs in O2DPG are for the async reco (which might run on the GRID not on the EPN), and the SYNCMODE=1 is for running sync reco, but not on the EPN (Sync reco on the EPN is EPNSYNCMODE=1). I.e. in these cases we cannot access `/home/epn/...`.
I think the geometry should be fetched from the CCDB by default, not from a hardcoded file. For MC perhaps one can take it from the local folder, not sure where the QC runs, for the SYNCMODE the geometry file is usually not in the local folder. It should be fetched from the CCDB.

iravasen:
Hi @chiarazampolli, yes and this works correctly (I tested locally) provided that on the grid the geom file exists in the local folder (as you said)

PR #406, opened 2022-05-18
Add keep-going feature

sawenzel:


MC/bin/o2_dpg_workflow_runner.py
```diff
@@ -774,9 +775,8 @@ def monitor(self, process_list):
             # --> We could use this for corrective actions such as killing jobs currently back-filling
             # (or better hibernating)
 
-    def waitforany(self, process_list, finished):
+    def waitforany(self, process_list, finished, failingpids = []):
```
benedikt-voelkel, line 778: Just note that if you give a list as default argument, it is filled further and further with multiple calls to the same function

sawenzel: I am actually passing an argument so this shouldn't happen.

PR #408, opened 2022-05-18
Fix condition for TOF calibration

chiarazampolli:
@davidrohr 

@noferini : in the current implementation we attach the calibration for TOF (with tracks) also when we have the ITSTPCTOFTRD matching, but probably this case is not needed since, if I am not wrong, we only use ITSTPCTPOF tracks. Can you confirm? If so, I will remove it (even if it does not harm).

### DIFF
--- a/DATA/common/setenv_calib.sh
+++ b/DATA/common/setenv_calib.sh
@@ -22,7 +22,7 @@ if [[ $BEAMTYPE != "cosmic" ]] || [[ $FORCECALIBRATIONS == 1 ]] ; then
 
     # calibrations for TOF
     if has_detector_calib TOF && has_detector_reco TOF; then
-	if has_detector_matching ITSTPCTOF || has_detector_matching ITSTPCTRDTOF; then
+	if ( has_detectors_reco ITS TPC && has_detector_matching ITSTPCTOF ) || ( has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRDTOF ); then
 	    if [[ -z ${CALIB_TOF_LHCPHASE+x} ]]; then CALIB_TOF_LHCPHASE=1; fi
 	    if [[ -z ${CALIB_TOF_CHANNELOFFSETS+x} ]]; then CALIB_TOF_CHANNELOFFSETS=1; fi
 	else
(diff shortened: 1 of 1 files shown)


noferini:
We can keep it as it is, in any case calibs info will be produced

DATA/common/setenv_calib.sh
```diff
@@ -22,7 +22,7 @@ if [[ $BEAMTYPE != "cosmic" ]] || [[ $FORCECALIBRATIONS == 1 ]] ; then
 
     # calibrations for TOF
     if has_detector_calib TOF && has_detector_reco TOF; then
-	if has_detector_matching ITSTPCTOF || has_detector_matching ITSTPCTRDTOF; then
+	if ( has_detectors_reco ITS TPC && has_detector_matching ITSTPCTOF ) || ( has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRDTOF ); then
```
davidrohr, line 25: Hi @chiarazampolli : just one question: how does the TOF calibration know which traks to assume as input?
ITSTPCTOF or ITSTPCTRDTOF or both?

DATA/common/setenv_calib.sh
```diff
@@ -72,6 +72,17 @@ if [[ $BEAMTYPE != "cosmic" ]] || [[ $FORCECALIBRATIONS == 1 ]] ; then
 
 fi
 
+echo "CALIB_PRIMVTX_MEANVTX = $CALIB_PRIMVTX_MEANVTX" 1>&2
```
davidrohr, line 75: All these printouts are ok for now for debugging, but in the long run I'd like to remove them, or only activated them with some env variable like `GEN_TOPO_VERBOSE=1`

DATA/common/setenv_calib.sh
chiarazampolli: The calib information could in principle be prepared with both. I think that currently we use only ITSTPCTOF, this was my question to @noferini , he already commented to keep it as it is in this PR.

DATA/common/setenv_calib.sh
chiarazampolli: Yes, sure.

chiarazampolli:
It seems that both @davidrohr and @noferini  agree, so I am merging.

PR #412, opened 2022-05-20
fetch timestamp based on run number

sawenzel:
This is fixing the timestamp to the run number,
in case no external timestamp was passed.


chiarazampolli:
Hi @sawenzel , 

thanks!
Ideally, we should do something similar also for the B field.

Chiara

sawenzel:
Can do (in next iteration).

chiarazampolli:
Perfect, thanks again!

PR #415, opened 2022-05-20
Small fixes

benedikt-voelkel:
* make cmd args and timestamp of type int consistently

* extend and actually make tuple


benedikt-voelkel:
@sawenzel @chiarazampolli Just a few small updates, without them the sim WF creation fails

chiarazampolli:
Seems fine to me, thanks!

chiarazampolli (approved):

PR #416, opened 2022-05-20
adding TPC

chiarazampolli:
@wiechula , @davidrohr 

@wiechula : can you check that the condition to enable the calibration is correct? For now it is just 
```
if has_detector_calib TPC; then
```

This PR goes with https://github.com/AliceO2Group/AliceO2/pull/8912

### DIFF
--- a/DATA/common/setenv_calib.sh
+++ b/DATA/common/setenv_calib.sh
@@ -35,12 +35,19 @@ if [[ $BEAMTYPE != "cosmic" ]] || [[ $FORCECALIBRATIONS == 1 ]] ; then
     fi
 
     # calibrations for TPC
-    if has_detector_calib TPC && has_detectors ITS TPC TOF TRD; then
-	if has_detectors TPC ITS TRD TOF && has_detector_matching ITSTPCTRDTOF; then
-	    if [[ -z ${CALIB_TPC_SCDCALIB+x} ]]; then CALIB_TPC_SCDCALIB=1; fi
-	else
-	    CALIB_TPC_SCDCALIB=0
-	fi
+    if has_detector_calib TPC; then
+       if has_detectors ITS TPC TOF TRD; then
+	   if has_detectors TPC ITS TRD TOF && has_detector_matching ITSTPCTRDTOF; then
+	       if [[ -z ${CALIB_TPC_SCDCALIB+x} ]]; then CALIB_TPC_SCDCALIB=1; fi
+	   else
+	       CALIB_TPC_SCDCALIB=0
+	   fi
+       fi
+       if [[ -z ${CALIB_TPC_TIMEGAIN+x} ]]; then CALIB_TPC_TIMEGAIN=1; fi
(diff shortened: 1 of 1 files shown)


DATA/common/setenv_calib.sh
```diff
@@ -91,6 +100,10 @@ if [[ -z $CALIBDATASPEC_BARREL ]]; then
     if [[ $CALIB_TOF_LHCPHASE == 1 ]] || [[ $CALIB_TOF_CHANNELOFFSETS == 1 ]]; then add_semicolon_separated CALIBDATASPEC_BARREL "calibTOF:TOF/CALIBDATA/0"; fi
     if [[ $CALIB_TOF_DIAGNOSTICS == 1 ]]; then add_semicolon_separated CALIBDATASPEC_BARREL "diagWords:TOF/DIAFREQ/0"; fi
 
+    # TPC
+    if [[ $CALIB_TPC_TIMEGAIN == 1 ]]; then add_semicolon_separated CALIBDATASPEC_BARREL "tpcmips:TPC/MIPS"; fi
+    if [[ $CALIB_TPC_RESPADGAIN == 1 ]]; then add_semicolon_separated CALIBDATASPEC_BARREL "trackGainHistoTPC:TPC/TRACKGAINHISTOS"; fi
```
wiechula, line 105: If this is used for the `o2-dpl-raw-proxy`, shouldn't it be
```suggestion
    if [[ $CALIB_TPC_TIMEGAIN == 1 ]]; then add_semicolon_separated CALIBDATASPEC_BARREL "tpcmips:TPC/MIPS/0"; fi
    if [[ $CALIB_TPC_RESPADGAIN == 1 ]]; then add_semicolon_separated CALIBDATASPEC_BARREL "trackGainHistoTPC:TPC/TRACKGAINHISTOS/0"; fi
```
Or is it for the `o2-dpl-output-proxy`, then I think it is fine.

wiechula (commented):
@chiarazampolli , I'm not too familiar with the scripts, so it is hard for me to judge if the logic is ok.

DATA/common/setenv_calib.sh
chiarazampolli: Hi @wiechula , 

Indeed, someone told me not long ago that one of the two proxies might have needed the subSpec, but I was not sure (you confirm that this is the case, only for one proxy?). Now I added it. 

Cheers,

Chiara

chiarazampolli: Btw, the spec is used both for the output and raw proxy.

chiarazampolli:
> @chiarazampolli , I'm not too familiar with the scripts, so it is hard for me to judge if the logic is ok.

Sure, but do you know whether the calib wf added from line 67 in the aggregator (https://github.com/AliceO2Group/AliceO2/pull/8912/files) need something from the reconstruction? If so, we need to use `has_detector_reco`.

Unfortunately, I had no time to test it real-life since my O2 needs to recompile a lot of stuff it seems, and I was doing some other debugging for which I could not start it. I am leaving tomorrow early morning, and as discussed with @davidrohr , it was good to open the PR before then.

PR #419, opened 2022-05-23
by default take field from CCDB

sawenzel:


sawenzel:
should not be merged at moment, until we have GRPMagField defined for all MC runs.

chiarazampolli:
Thanks for the implementation!

sawenzel:
Merging as https://alice.its.cern.ch/jira/browse/O2-2958 is fixed.

PR #421, opened 2022-05-23
Add hypertriton gun on top of Pythia8

mpuccio:


mpuccio:
@sawenzel this only works with G3 for the time being as we miss the hypertriton in the particle definition. I tried adding it in the UserDecay file, but it seems to me that we do not load it anywhere (it is loaded in FairRoot in the FairMCApplication, but we do not depend on that anymore). Without the particle definition, even though we can rely on the external decayer (pythia8), G4 gets stuck. Do you have any practical suggestions?

sawenzel:
Normally, we still use FairRoot so this should work. Could you create a JIRA ticket specifying the goal what should work with G4 again in your own words. Better to follow up there.

PR #426, opened 2022-05-24
add TOF matching QC in sync mode

noferini:
@ercolessi @njacazio @shahor02 @chiarazampolli @knopers8 

Matching task to be added also in consul for both TOF configurations


DATA/production/qc-workflow.sh
```diff
@@ -22,6 +22,7 @@ if [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
     [[ -z "$QC_JSON_PHS" ]] && QC_JSON_PHS=consul://o2/components/qc/ANY/any/phos-raw-clusters-epn
     [[ -z "$QC_JSON_PRIMVTX" ]] && QC_JSON_PRIMVTX=/home/shahoian/jsons/vertexing-qc.json
     [[ -z "$QC_JSON_GLOBAL" ]] && QC_JSON_GLOBAL=$O2DPG_ROOT/DATA/production/qc-sync/qc-global.json
+    [[ -z "$QC_JSON_TOF" ]] && [[ ! -z "$TOF_SOURCES" ]] && QC_JSON_TOF_MATCH=$O2DPG_ROOT/DATA/production/qc-sync/itstpctof.json
```
knopers8, line 25:
```suggestion
    [[ -z "$QC_JSON_TOF_MATCH" ]] && [[ ! -z "$TOF_SOURCES" ]] && QC_JSON_TOF_MATCH=$O2DPG_ROOT/DATA/production/qc-sync/itstpctof.json
```
don't you need this instead (similarly in other lines)?

noferini: you are right, I'll fix it. Thanks

noferini:
Tested with FST
It needs 
https://github.com/AliceO2Group/AliceO2/pull/8981
before to be merged

noferini:
There are no longer dependencies with the PR in O2.
It can be merged as soon as you agree

davidrohr:
ok, let's wait for the FullCI, then we can merge

noferini:
I guess that the error is due to a missing token in the check

Error in <TJAlienConnectionManager>: Failed to get any credentials
Error in <TJAlienFile::Open>: No JAliEn Grid connection available!
[INFO] ccdb reads http://alice-ccdb.cern.ch/CPV/Calib/Gains/1627542202422/fdbfdde0-5441-11ec-8357-200114580202 (retrieve, agent_id: alientest06-1653746611-aQ6Xup), 
[INFO] CCDB: Reading o2::cpv::Pedestals from CPV/Calib/Pedestals
Error in <TJAlienFile::Open>: No JAliEn Grid connection available!

davidrohr:
@TimoWilken : Could you restart the FST CI. It has failed due to an alien problem.
Could be that this was due to general alien issues yesterday.

noferini:
I don't have rights to do restart checks.

davidrohr:
You can force push an ammended commit, that will trigger a restart.

noferini:
I see, done

noferini:
:+1:

davidrohr:
@noferini : That seems to be a legitimate error now in the fullCI. Could you check?

noferini:
> @noferini : That seems to be a legitimate error now in the fullCI. Could you check?

Indeed, it is even if I don't understand what is different from my local tests.
I am looking into that

noferini:
It was a my mistake (sorry, too many open branches in my laptop).
I also pushed one more commit to add an additional inputspec in the TOF digit QC
I'll do the same also in QualityControl
@ercolessi @njacazio : it should be added also in the consul json as well
The addition of the inputspec doesn't do anything at the moment but it is required in all places before to modify the DigitTask 
This is the inputspec: diafreq:TOF/DIAFREQ/0

noferini:
Hi check is failing and I actually found the problem.
Not related to O2DPG but QualityControl (TPC tracks not loaded if TPC-TOF matching not required).
I will open a PR in QC and then I will rerun the check here after the needed QC change is merged.
Sorry for boring you with this long story

noferini:
Thanks
I see your comment (probably we can already start to set and test the merger even if data are not yes sent, let's try)

PR #427, opened 2022-05-24
Fix ITS QC JSON

davidrohr:
@iravasen : Please note for next time that the QC JSONs in DATA/production/qc-sync are used for the FST. Please update them as well.


iravasen:
@davidrohr thanks! Sorry I did not know, now I understand why the O2fst tests were failing in my PR in QC. Thanks.

PR #428, opened 2022-05-24
Use full reconstruction QC in async mode or when MID_RECO is defined

dstocco:


DATA/production/qc-async/mid.json
```diff
@@ -31,12 +31,166 @@
         "cycleDurationSeconds": "60",
         "maxNumberCycles": "-1",
         "dataSource": {
-          "type": "direct",
-          "query": "digits:MID/DATA/0;digitrofs:MID/DATAROF/0;calib:MID/DATA/1;calibrofs:MID/DATAROF/1;fet:MID/DATA/2;fetrofs:MID/DATAROF/2"
-        }
+          "type": "dataSamplingPolicy",
+          "name": "mid-digits"
+        },
+        "location": "local",
```
chiarazampolli, line 37: The lines below are not needed in async, please check with @knopers8

dstocco: I tried to remove the lines that you point out, but if I do so I get: 

```
50824:mid-clusters-proxy]: [10:06:36][ERROR] could not resolve hostname 'any', reason: resolve: Host not found (authoritative)
[50824:mid-clusters-proxy]: [10:06:36][ERROR] failed to attach channel mid-clusters[0] (connect)
```

Unfortunately I am not a QC expert: not exactly sure what is wrong.

dstocco:
@ValerieRamillien

DATA/production/qc-async/mid.json
dstocco: More specifically, I cannot remove `remotePort` and `remoteMachine` parameters.

knopers8: when running asynchronously, we use `--local-batch`, not `--local --host <host>`, so the results are stored in a file, not send to a remote QC server. Please try to execute your workflow again with the first argument (`--local-batch`).

dstocco: I'm launching it with the full-system-test:

```
GEN_TOPO_WORKDIR=./ WORKFLOW_PARAMETERS=QC GEN_TOPO_RUN_HOME=1 QC_REDIRECT_MERGER_TO_LOCALHOST=1 WORKFLOW_DETECTORS=MID WORKFLOW_DETECTORS_QC=MID IGNORE_EXISTING_SHMFILES=1 CTF_DICT=$ALIDATA/o2/raw/2021/OCT/505645/ctf_dictionary.root $O2_ROOT/prodtests/full-system-test/run-workflow-on-inputlist.sh CTF o2_ctf_run00505645_orbit0062644798_tf0000487943.root
```

which calls the qc-workflow.sh where I read: `add_W o2-qc "--config json://$QC_JSON_FROM_OUTSIDE ${QC_CONFIG_PARAM:---local --host ${QC_HOST:-localhost}} ${QC_CONFIG}" "" 0`

I see that there are other scripts that run with `--local-batch`. But the point is: what is running on the Continuous Integration machines? I guess I should be careful not to break it.

Thanks!
Cheers,
Diego

knopers8: I might be wrong, but I thought that the CI runs the synchronous processing, thus it includes the QC from `qc-sync`. These ones should indeed keep the key-values related to multinode setups, because they are executed with `--local`

What goes to `qc-async` is executed with `--local-batch`.

dstocco: Ok, I tested that it is working with `--local-batch`: I've just committed the modification.
Thanks!
Cheers,
Diego

davidrohr: For reference The CI runs sync reco with QC and async reco without QC.
JSON files for the sync QC are in DATA/production/qc-sync

DATA/production/qc-workflow.sh
```diff
@@ -32,7 +32,7 @@ if [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
     [[ -z "$QC_JSON_FV0" ]] && QC_JSON_FV0=$O2DPG_ROOT/DATA/production/qc-sync/fv0.json
     [[ -z "$QC_JSON_EMC" ]] && QC_JSON_EMC=$O2DPG_ROOT/DATA/production/qc-sync/emc.json
     [[ -z "$QC_JSON_MCH" ]] && QC_JSON_MCH=$O2DPG_ROOT/DATA/production/qc-sync/mch.json
-    [[ -z "$QC_JSON_MID" ]] && QC_JSON_MID=$O2DPG_ROOT/DATA/production/qc-sync/mid.json
+    [[ -z "$QC_JSON_MID" ]] && has_processing_step MID_RECO && QC_JSON_MID=$O2DPG_ROOT/DATA/production/qc-sync/mid.json || QC_JSON_MID=$O2DPG_ROOT/DATA/production/qc-sync/mid-digits.json
```
davidrohr, line 35: I am not sure if this `&&` and `||` logic is correct. Don't we need 
```
[[ -z "$QC_JSON_MID" ]] && ( has_processing... && a || b)
```

dstocco: In my tests this logic was working as expected. Anyways, I can have a deeper look next week

davidrohr:
@TimoWilken : Also here, could you restart the FST CI, failed due to CCDB problem.

DATA/production/qc-workflow.sh
dstocco: Hi @davidrohr : indeed, without the parenthesis, the syntax was not working as expected when QC_MID_JSON was already set (it re-defined it anyways).
On the other hand, adding a parenthesis does not seem to work very well on Mac. I therefore modified the syntax with a more standard conditional assignment.

dstocco:
@shahor02 and @martenole : is this PR ok for you?

davidrohr (approved):
Fine with me.
But wan't to wait for the CI to pass before merging.

dstocco:
While waiting for the merging, I made some modification to better treat the digits subspecs in QC. The PR was just merged https://github.com/AliceO2Group/QualityControl/pull/1258 . The PR requires a specific alias for the ROFs, which I implemented in the latest commit. Everything should be now consistent with the QC master.

knopers8:
Since the tests are passing, then I assume we can merge it.

PR #448, opened 2022-06-02
Prepare anchorMC script to also run the async_reco part

sawenzel:


sawenzel:
When `export RUN_ASYNCRECO=ON` and `export CTF_TEST_FILE=some_ctf_file` ... this script will also run a short async_reco workflow which we'd like to do on alibi.

So something like this:
```
# get CTF input_file
alien.py cp /alice/data/2021/OCT/505673/raw/0850/o2_ctf_run00505673_orbit0084970764_tf0000663593.root file:./
export CTF_TEST_FILE=o2_ctf_run00505673_orbit0084970764_tf0000663593.root
export RUN_ASYNCRECO=ON

NTIMEFRAMES=2 ${O2DPG_ROOT}/MC/run/ANCHOR/2021/OCT/pass4/anchorMC.sh
```

@chiarazampolli : Please let me know once this works for you. Then we can integrate it.

sawenzel:
Did some small modifications and for me it works now. Will merge it and in principle we should have a first async_pass test this night.

PR #451, opened 2022-06-03
Remove checks from json since they are not used.

chiarazampolli:
The current version was misconfigured, and even if not used, it was making the QC crash.


chiarazampolli:
Tested locally for one of the json, merging.

PR #452, opened 2022-06-03
Use --configKeyValues for QC in FST and online + unrelated fixes

davidrohr:


davidrohr:
@martenole : TRD is crashing with this PR, although I do not see how this could affect the TRD tracking:
```
[41085:trd-globaltracking_TPC_ITS-TPC_]: [19:06:03][INFO] As input seeds are available: 5324 ITS-TPC matched tracks and 20825 TPC tracks
[41085:trd-globaltracking_TPC_ITS-TPC_]: [19:06:03][INFO] 3413 tracks are loaded into the TRD tracker. Out of those 1546 ITS-TPC tracks and 1867 TPC tracks
[ERROR] pid 41085 (trd-globaltracking_TPC_ITS-TPC_) crashed with 128
[INFO] QUIT_REQUESTED
```

martenole:
Seems it was a random error? The current check just failed because of the grid certificate

davidrohr:
@TimoWilken : The CI is failing since it lacks / cannot get an alien certificate:
```
[INFO] CCDB: Reading o2::cpv::CalibParams from CPV/Calib/Gains
Error in <TJAlienConnectionManager>: Failed to get any credentials
Error in <TJAlienFile::Open>: No JAliEn Grid connection available!
```
Could you check?

davidrohr:
> Seems it was a random error? The current check just failed because of the grid certificate

ok, indeed it seems to happen randomly. But still there was a crash in TRD reco, which should never happen. But I agree if we have no way to reproduce it, we cannot do much.

PR #453, opened 2022-06-07
fix for name too long in json

noferini:
@davidrohr 
I think that the name is too long and it may break with the recent changes in QC framework

### DIFF
--- a/DATA/production/qc-sync/itstpctrdtof.json
+++ b/DATA/production/qc-sync/itstpctrdtof.json
@@ -23,7 +23,7 @@
       }
     },
     "tasks" : {
-      "MatchingTOFwithTRD" : {
+      "MatchingTOFwTRD" : {
         "active" : "true",
         "className" : "o2::quality_control_modules::tof::TOFMatchedTracks",
         "moduleName" : "QcTOF",


noferini:
Adding also @knopers8 and @chiarazampolli 

The error is
Error in <TFile::TFile>: file /tmp/wn_dds/4578dee1-0185-49a5-bd97-ca0befec4fba/97ca1dd8-2aa9-488f-aca1-a4695bcdb848/2YcXps1pBPf_151166_epn113/slots/8017573906789007788/o2sim_geometry-aligned.root_geometry-aligned.root does not exist

davidrohr:
@noferini : ok, I can merge this.
For the future, could you switch to using JSONs from consul for `EPNSYNCMODE=1` so you can change them yourself.
the local files in O2DPG/DATA/production/qc-sync are just copies to be used for the FST.

noferini:
Yes we can add it to consul

chiarazampolli:
The error does not have to do with the name of the task - when it is too long it triggers a warning, at least in the past it was not fatal (@knopers8 , can you confirm? we have others that are long, and nothing ever happened). It seems to me that you don't have the geometry file that you need.

noferini:
We have two json one with and one without TRD
If geometry file is fine without TRD it should be fine also with TRD in.
the two json structures are identical

chiarazampolli:
Ok, then I don't know, but as I said, we have other cases of long task names, and they always worked (even if I always wanted to shorten them).

shahor02:
The geometry file was looked in the working directory while should be taken from /home/epn/odc

knopers8:
> 

That is correct, it will trigger a warning (which is good to fix). If you use too long names, you risk colliding e.g. "MatchingTOFwithTRD" with "MatchingTOFwithTPC", because data description is 16 characters.

noferini:
Hi @shahor02,
indeed the geometry and grp path are hardcoded in the spec.
I will comment/remove the setting of a specific path assuming that the default is set in the workflow
  o2::base::GeometryManager::loadGeometry(); //mGeomFileName);
  o2::base::Propagator::initFieldFromGRP(); //mGRPFileName);
Is it a correct assumption?

martenole:
Hi @noferini ,
unfortunately, not yet. In principle this will be taken from the workflow, but only once o2-qc honors the configKeyValues. Until then you can, as for example ITS is doing it, define the geometry path in the json file and take it from there:
https://github.com/AliceO2Group/QualityControl/blob/feca8543d476b5ab1d9a62107bc101b300a0a725/Modules/ITS/src/ITSClusterTask.cxx#L568
https://github.com/AliceO2Group/QualityControl/blob/feca8543d476b5ab1d9a62107bc101b300a0a725/Modules/ITS/src/ITSClusterTask.cxx#L106

noferini:
I see, I will copy it
Cheers
Francesco

davidrohr:
But o2-qc is already respecting the --configKeyValues (at least it is passed to o2-qc), or do I miss something?
Could you try if it works with --configKeyValues?

martenole:
The last time I checked (when we had similar issues with ITS in middle of May) this was not yet the case. But maybe @knopers8 can comment?

davidrohr:
With the latest QC version (actually I think already the previous one), o2-qc accepts `--configKeyVaues` and since 3 days ago (https://github.com/AliceO2Group/O2DPG/commit/9aedeb08ed0e3b1f7308f82f6d65c7ad251974c6) the FST and the sync reco workflow pass the `--configKeyValues` to `o2-qc`.

knopers8:
> The last time I checked (when we had similar issues with ITS in middle of May) this was not yet the case. But maybe @knopers8 can comment?

I confirm what David says. I added the argument and corresponding lines to load the params as instructed. If it does not work, please let me know (a reproducer would be appreciated).

PR #454, opened 2022-06-07
TOF Match: Fetch json config from consul

njacazio:


njacazio:
This is a follow up on: https://github.com/AliceO2Group/O2DPG/pull/453 
The consul JSON are used (keys already created)

PR #456, opened 2022-06-08
reduce task name for safety

chiarazampolli:
@knopers8


knopers8 (approved):
thanks. If you want to keep the longer names in QCDB, you can add e.g.:
```
"taskName" : "MatchedTracksITSTPC"
```
in the body of a config structure.

chiarazampolli:
Thanks for the suggestion @knopers8 , I added the parameter in some cases, when it could be useful to make it clearer. Is it ok? (just to be sure)

chiarazampolli:
I have also removed a json that was wrongly configured (with MC flag true, but no MC inputs), and not used, so it was confusing (@noferini)

knopers8 (approved):

PR #458, opened 2022-06-10
First round of pythonising and args consistency

benedikt-voelkel:
* pythonise if-else

* add required types to cmd args

* use '--' consistently for optional args

### DIFF
--- a/MC/bin/o2dpg_qc_finalization_workflow.py
+++ b/MC/bin/o2dpg_qc_finalization_workflow.py
@@ -19,6 +19,62 @@ sys.path.append(join(dirname(__file__), '.', 'o2dpg_workflow_utils'))
 
 from o2dpg_workflow_utils import createTask, dump_workflow
 
+#################################################################################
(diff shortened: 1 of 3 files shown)


benedikt-voelkel:
@sawenzel @chiarazampolli @catalinristea 
Especially the cmd args are confusing in the sim WF.
At the same time I adjusted some overhead in `if-else` statements for now to enhance the code (readability)

sawenzel:
As much as I like this, I think it will unfortunately be quite disruptive in many places (especially for scripts maintained outside of O2DPG). Should be done together with updating all production scripts.

benedikt-voelkel:
Let me try to have both the `-<long-option>` and the `--<long-option>` and we issue a deprecation warning when people use the first way.
So I will update this PR soon.
At the same time we notify people so that in a few weeks-months from now we can actually remove the `-<long-option>`

benedikt-voelkel:
Deprecation warnings added when using arguments like `-<long-arg>` (only single dash). The functionality shall eventually be removed together with the deprecated args.

benedikt-voelkel:
close for now, can easily be re-opened and updated

PR #467, opened 2022-06-16
Move output of calibration to ali-ccdb

rmunzer:


chiarazampolli:
Hi @rmunzer , @davidrohr , @costing

Shouldn't the output go to localhost:8084?

rmunzer:
> Hi @rmunzer , @davidrohr , @costing
> 
> Shouldn't the output go to localhost:8084?

Indeed i found both looking to scripts of other detectors. 
I tested with the current path, which was working fine. 
But maybe localhost:8084 is the better?

davidrohr:
Not sure if it is o2-ccdb.internal or localhost:8084 for the output. @costing should comment

costing:
`localhost:8084` will redirect to `o2-ccdb.internal` for any changes. So both will work. 

If the CCDB endpoint is configured somewhere globally (to point to `localhost:8084` in general), I suggest using this variable everywhere instead of hardcoding any address.

davidrohr:
ok, if both work, then I'd propose not to configure any endpoint at all, but use the default one, which should be set via `--configKeyValues` if the CCDB manager is used directly, or via the `DPL_CCDB_BACKEND`, if the dpl ccdb fetcher is used.
On the EPN, both should be set correctly automatically.

chiarazampolli:
Hi @davidrohr , 
But here it is not the fetcher that is used, but the ccdb-populator.

davidrohr:
@chiarazampolli : I have actually no idea about the populator, but isn't it using the CCDB Manager?
If so, it would automatically use the correct CCDB server.
If not, we should either make the populator take the default from `o2::base::NameConf::getCCDBServer()` or alternatively have another default in NameConf for CCDB uploads (if we want to allow for different fetch and upload servers).

chiarazampolli:
ccdb-populator now uses as default "http://alice-ccdb.cern.ch/", see:
https://github.com/AliceO2Group/AliceO2/blob/3e965b900a22ff1a00e8e81fc007e5873cbbcb42/Detectors/Calibration/workflow/ccdb-populator-workflow.cxx#L47
which goes to:
https://github.com/AliceO2Group/AliceO2/blob/a2201a106592416bee225a3e48c2fbf856c4dda5/Common/Utils/include/CommonUtils/NameConf.h#L143

right?

rmunzer:
Thus i think it should be best to just remove the path in the ccdb-populator to always use the default one.

chiarazampolli:
It will then push to alice-ccdb.cern.ch, and this is not the best for online, @costing

costing:
For sure the default value of `alice-ccdb.cern.ch` cannot be used in the Online environment. It can stay there as the default for everything else (users, Grid jobs etc), but it should be overwritten on the Online nodes as David suggests.

rmunzer:
Thus it means, it can stay like it is as an option in ccdb-populator.

chiarazampolli:
Yes, but I still think you should use "localhost:8084". Then surely the best is that this is controlled by the online world, and not workflow by workflow. But till this is done, I would use localhost:8084. Unless they (localhost and o2-ccdb.internal that you are using) are totally equivalent (localohost will redirect you to o2-ccdb.internal). @costing

davidrohr:
it doesn't matter which of the 2 you are using, both will work. o2-ccdb.internal is fine

chiarazampolli:
Sure, but I thought the kind of best way was to use localhost.

PR #469, opened 2022-06-16
Scripts for apass1 of MayJune2022

chiarazampolli:
@catalinristea, @noferini 

@wiechula : the way we get the vdrift is based on the macro I wrote before knowing that the laser runs were not taken before any fill. There is a switch to use CCDB with the start time of the run (which will not work, for the reason above), or with a timestamp that is run dependent and hardcoded in the macro. Could you provide these timestamps?

Cheers,

Chiara


catalinristea:
Ciao @chiarazampolli,

Here is how we bind alien with O2DPG:

DPGRECO=$O2DPG_ROOT/DATA/production/configurations/$ALIEN_JDL_LPMANCHORYEAR/$ALIEN_JDL_LPMPRODUCTIONTAG/$ALIEN_JDL_LPMPASSNAME/async_pass.sh

For OCT pilot beam apass4,
LPMAnchorYear=2021
LPMProductionTag=OCT
LPMPassName=apass4

- apass4 is the only one we can set from JDL
- We may override the LPMPRODUCTIONTAG to MayJunePilotBeam, but this could add ambiguity for LPM setup. There are many others LPM variables which contained OCT, so we don't know for sure how these are filled at runtime - no strong opinion, we can try, no problem. 

If we are to stay unchanged, we'll need to duplicate into 2 configurations,
1. DATA/production/configurations/2022/MAY/apass1/async_pass.sh
2. DATA/production/configurations/2022/LHC22c/apass1/async_pass.sh

Please note for the time being we have just MAY runs scheduled,
Runlists:
- MAY: 517222, 517224, 517032 (individual HMP, but still in global)

Thank you!

Yours, Catalin.

chiarazampolli:
Hi @catalinristea , 

There is already a workaround for the period in async_pass.sh, as I would prefer to not duplicate:

```
( [[ $ALIEN_JDL_LPMPRODUCTIONTAG == MAY ]] || [[ $ALIEN_JDL_LPMPRODUCTIONTAG == LHC22c ]] ) && O2DPGPATH="MayJunePilotBeam"
```

The scheduled runs are those here:

https://docs.google.com/spreadsheets/d/14vXFYVx3oVE8wgJKmIBhjvAt6NpGp7D6H4AmBM9E0Cw/edit#gid=512457836

and more will come next week... (sorry that I did not update the JIRA about this).

Cheers,

Chiara

catalinristea:
Ciao Chiara,

Great, then you already took care. Sorry for the noise. Thanks!

Yours, Catalin.

chiarazampolli:
No noise, better to always check 2x... :-) 
Note that I will need to update this PR with some news on the filling scheme. 

Chiara

chiarazampolli:
Filling scheme updated. 
If nobody is against it, I would merge. Then we can try a functional test, with not updated nor correct TPC vdrift, just to see if it works.

Chiara

chiarazampolli:
Merged to allowed testing (i closed it initially by mistake).

PR #471, opened 2022-06-17
Change MID digits query alias to cope with QC modifications now in QC 1.63.0

dstocco:
This commit was kind of reverted because the corresponding commit in QC was not bumped. This resulted in a failure of the fullCI tests, that was still using an old QC version.
Now the QC is bumped, and the tagged version contains the needed information. The json files can be therefore safely modified accordingly.


chiarazampolli:
@davidrohr , ok for you to merge? It touches also sync QC - but I think the json are not taken from here, are they?

PR #473, opened 2022-06-20
Fixes in apass1 for MayJune 2022 data

chiarazampolli:
- TRD: remove the checks for QC: they are badly configured, and anyway
not used;
- VDRIFT: restore correct command
- REMAPPING: fix syntax

@catalinristea: this is though still WIP, I have issues with a PID QC task now. More news asap.


chiarazampolli:
Error in QC from FT0TOF PID task was spurious: reloading the environment, it disappeared (magic of aliBuild? :-) )

chiarazampolli:
Now analysis QC only does not work, due to the fact that the timestamp task is accessing RCT/RunInformation while now we use RCT/Info/RunInformation. Analysis coordinator (David Chinellato) contacted.

chiarazampolli:
@bazinski , just FYI, I changed the TRD json for async reco since the checks were misconfigured. I removed them completely, since they were not used. If you need them, please add them back but with the correct task name etc.

bazinski:
@chiarazampolli yes remove them, no worries. I think I need to give all those json files a once over.

PR #474, opened 2022-06-20
[EMCAL-755] Change DigitsQCTask to CellTask in simulation

mfasDa:
CellTask was branched off several months ago to
monitor cell level information (in AliRoot Digits and Cells
were the same objects for EMCAL as the time response
was not simulated, with the introduction of time sampling
they diverged). A transition period was kept to not affect
existing productions. The workflows should be adapted now
to allow using the DigitsQCTask to monitor digits (time
sample) level quantities.


chiarazampolli:
Looks fine to me. Merging, @knopers8 is away. If he has more comments, we can take them later.

chiarazampolli (approved):

PR #475, opened 2022-06-20
updated macro for TPC vdrift + analysis QC optional + typo

chiarazampolli:
@catalinristea


chiarazampolli:
Don't merge yet.

chiarazampolli:
@catalinristea : for me with this PR it works. We still have issues with Analysis QC, which is why I added an env variable from the jdl (RUNANALYSISQC) which, when set explicitly to false, will not run it. But then we need also to adjust the files that we save, I guess.

PR #478, opened 2022-06-21
tuning tof digit qc params

noferini:
@njacazio @ercolessi 
Updating tof qc json for apass reconstructions


njacazio:
Thanks Francesco!

PR #479, opened 2022-06-21
fix parse.sh and add TOF option

chiarazampolli:
@catalinristea, @noferini , please check, then I will merge.


noferini:
@chiarazampolli 
For me it is fine

PR #493, opened 2022-06-24
Fixes and some changes

benedikt-voelkel:
* remove redundant plotting of simgle histograms, they are contained in
  the ratio plots

* taken out previous ratio plotting and replace by more modular approach
  TODO:
  * Add for 3D
  * write to summary ROOT file so that we can eventually extract
    critical histograms when desired

* update labels shown in ratio plots
  If only non-critical test failed, show "COMPATIBLE" label in ratio
  plots.
  If at least one critical test failed, show "BAD"
  (before that was overwritten by each proceeding test such that it
  could show compatoble even though an earlier critical test failed)

* resolve memory issues
  many mem problems, in particular from previous plotting. Most of the
  resolved by simply moving to new plotting;
  others were fixed


benedikt-voelkel:
@chiarazampolli @sawenzel

sawenzel (approved):

sawenzel:
Merging to test it during nightlies.

PR #496, opened 2022-06-27
Add TPC cluster residual output for CALIB

martenole:


davidrohr:
@chiarazampolli : Shouldn't we wait for the fst to finish? The FST in the CI runs also the calib workflows in the meantime :)

chiarazampolli:
Is the FST running in O2DPG? I did not think so.

davidrohr:
Sure, there is an FST build for O2DPG, which we have added since some O2DPG scripts are used for the FST:
 @alibuild
build/O2DPG/O2fst/o2 Started Mon 20:47 CET @ mesosci/ci_alisw_slc7-builder_latest/0

chiarazampolli:
Ok, then I apologize. But the chamge looks simple.
Not clear to me, though, how it works when o2 changes depend on o2dpg, how to test.

davidrohr:
Sure, but if e.g. the workflow would not produce `unbinnedTPCResiduals:/GLO/UNBINNEDRES/0`, then it would break the FST.
Obviously, when you have non-backward-compatible changes to both repos that must be merged in parallel, at least one must be force-merged without the check passing.

chiarazampolli:
This is just adding the dataspec, i thought it was “easier “ than the change in O2.

martenole:
> This is just adding the dataspec, i thought it was “easier “ than the change in O2.

I think that's true and https://github.com/AliceO2Group/AliceO2/pull/9281 needed this one to be merged first anyways

davidrohr:
sure, it is more or less trivial, and should be safe.
But this one here was backward compatible.
I was just noting that one could have waited for the CI to pass, then merge this one.
Then wait again for the CI for the other one, and then merge the other one.
Anyway, never mind :)

PR #498, opened 2022-06-27
MCH: select QC config based on presence of RECO processing step

aferrero2707:
The full QC configuration is selected when the MCH_RECO is enabled, otherwise the minimal QC configuration for raw data decoding is chosen.


aferrero2707:
@davidrohr @aphecetche could you please check if the proposed selection of the default QC configuration, depending on the presence of the MCH_RECO processing step, is correct?

Thanks!

chiarazampolli:
Fine with me, @davidrohr ?

PR #503, opened 2022-06-29
Use MID full reconstruction QC in EPN when MID_RECO is active

dstocco:


chiarazampolli:
Fine with me, @davidrohr ?

davidrohr:
could you rebase, and I'd prefer the syntax like in the other detectors. What you have there works, but it sets the JSON file name twice, which is kind of confusing.

dstocco:
Hi @davidrohr : done. Notice however that the syntax used by other detectors is not fully correct for me.
The condition can be false either if MID_RECO is not defined or MID is not in detector list. But in this second case, why would one want to define a QC_JSON for MID?
Anyways, it should be now uniformized to what others do.
Cheers,
Diego

davidrohr:
The has_detector MID has in principle no affect at all, since the QC JSON for MID is only added when MID is present. If it is not present, it doesn't matter which JSON is set.
Though for matching QC it is important, i.e. the TOF QC JSON depends on the presense of TRD, but that is a special case. So in principle one could remove these has_detector ...

PR #508, opened 2022-07-01
Update code and README

benedikt-voelkel:
* use flags --with-test-<test> and remove the need to know the correct
  value of the bit mask --> replaces --test|-t option

* add relative path of where single histograms are stored to JSON
  summary and influxDB file

* pipe RelVal macro verbosity to log file to focus on cerntral status
  messages

* do not compare branches that cannot be compared

* add simple pie charts for each compared file pair per test showing the
  fraction of test results

* udpate README accordingly wrt flags (see above) and explain plots that
  are created


chiarazampolli (approved):

chiarazampolli:
Thanks @benedikt-voelkel !

PR #512, opened 2022-07-06
TPC track interpolation optionally sends track data

martenole:


chiarazampolli:
Seems ok to me, but let's wait for the CI this time :-)

martenole:
I absolutely agree :)
Just updated again so that this can be merged without changes in O2 required. And then one can request externally to send the track data to the aggregator together with the unbinned residuals. After https://github.com/AliceO2Group/AliceO2/pull/9325 is merged the default can be changed to 1

PR #514, opened 2022-07-07
MCH: make calibrator parameters configurable via environment variables

aferrero2707:
It allows to define via environment variables the maximum acceptable pedestal and noise values, as well as the minimum statistics required to process the data and send the output to the CCDB.

The default maximum pedestal value is also changed to 500 to reflect the settings in the command-line tools.

### DIFF
--- a/DATA/production/calib/mch-badchannel-aggregator.sh
+++ b/DATA/production/calib/mch-badchannel-aggregator.sh
@@ -9,8 +9,12 @@ source common/getCommonArgs.sh
 PROXY_INSPEC="A:MCH/PDIGITS/0"
 CONSUL_ENDPOINT="alio2-cr1-hv-aliecs.cern.ch:8500"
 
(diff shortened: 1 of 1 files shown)


aferrero2707:
@davidrohr @martenole @shahor02 @chiarazampolli @aphecetche could you please have a look? The changes are relatively simple.
It would be great if we could have this still in the next FLP suite.

Thanks!

davidrohr:
@aferrero2707 : Find for me since this is a detector standalone calibration.

Just FYI: In the global workflow I prefer to not have so many separate env variables. You could also use the `add_W` syntax for adding workflows (you need to source `production/workflow-setup.sh`), which is the way the global dpl-workflow.sh does it. Then, you have `CONFIG_EXTRA_PROCESS_[name]` available as env variable to set configKeyValues for a process.

PR #518, opened 2022-07-07
More runs with MFT stronbe length = 297

chiarazampolli:
@rpezzi , @catalinristea


chiarazampolli:
Merged, we need it for async reco.

PR #519, opened 2022-07-07
Fix strobe length for runs with 396

chiarazampolli:
@rpezzi , @shahor02 , @catalinristea , @JianLIUhep 

396 is now considered as default. In case it won't be anymore, we will need to change.


chiarazampolli:
Needed for processing.

PR #523, opened 2022-07-11
V1

fapfap69:
FIrst version

### DIFF
--- a/DATA/production/calib/hmp-pedestals-processing.sh
+++ b/DATA/production/calib/hmp-pedestals-processing.sh
@@ -18,7 +18,7 @@ source common/setenv.sh
 # env  > /tmp/hmpid_pedestal_env_dump.txt
 
 # Set general arguments
-source getCommonArgs.sh
+source common/getCommonArgs.sh
 
 # Define the Input/Output streams
 PROXY_INSPEC="A:HMP/RAWDATA;dd:FLP/DISTSUBTIMEFRAME/0"
--- a/DATA/production/calib/hmp-pedestals-processing.sh
+++ b/DATA/production/calib/hmp-pedestals-processing.sh
@@ -15,20 +15,10 @@
 # ------------------------------------------------------------------------
 
 source common/setenv.sh
-# env  > /home/fap/dump_env.txt
+# env  > /tmp/hmpid_pedestal_env_dump.txt
 
-# ---------------------------------------------------------------------------------------------------------------------
 # Set general arguments
-ARGS_ALL="--session default --severity $SEVERITY --shm-segment-size $SHMSIZE $ARGS_ALL_EXTRA"
-ARGS_ALL+=" --infologger-severity $INFOLOGGER_SEVERITY"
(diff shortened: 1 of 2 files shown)


fapfap69:
Please could you review the PR, I'm not able to calculate pedestals . Thank you. @chiarazampolli @martenole @shahor02 @davidrohr

chiarazampolli:
Hi @fapfap69 , 

I cannot spot any issue with your PR, but maybe it is better if also @davidrohr takes a look.

Chiara

DATA/production/calib/hmp-pedestals-processing.sh
```diff
@@ -0,0 +1,99 @@
...
+#
+#
+#   Auth. A.Franco  - INFN  Sez.BARI - ITALY
+# ------------------------------------------------------------------------
+
+source common/setenv.sh
+# env  > /home/fap/dump_env.txt
+
+# ---------------------------------------------------------------------------------------------------------------------
+# Set general arguments
```
davidrohr, line 21: I would recommend to source `getCommonArgs.sh` instead of copy&pasting this part

fapfap69: Done ! Thank you

davidrohr: Did you test this? getCommonArgs is in the common folder as is setenv.sh, so I assume you have to do `source common/getCommonArgs.sh`.
Did you test this workflow at P2? You can check out a local clone with this PR to your home folder on the EPN, and then use the workflow configuration method = repository path pointing to your DATA path to test it, before merging to the official repository.

fapfap69: yes, i did all the test in my DATA path before the PR ...

fapfap69:
> Il giorno 22 lug 2022, alle ore 10:16, David Rohr ***@***.***> ha scritto:
> 
> 
> @davidrohr commented on this pull request.
> 
> In DATA/production/calib/hmp-pedestals-processing.sh <https://github.com/AliceO2Group/O2DPG/pull/523#discussion_r927408272>:
> 
> > +#
> +#  Extra Env Variables:
> +#     HMP_SIGMACUT : The value of sigams for the threshold cut [=4]
> +#     HMP_CCDB_REC : True if we want the recording into ALICE CCDB
> +#     HMP_PED_TAG :  A string that tags the pedestals [=Latest]
> +#
> +#
> +#
> +#   Auth. A.Franco  - INFN  Sez.BARI - ITALY
> +# ------------------------------------------------------------------------
> +
> +source common/setenv.sh
> +# env  > /home/fap/dump_env.txt
> +
> +# ---------------------------------------------------------------------------------------------------------------------
> +# Set general arguments
> getCommonArgs is in the common folder as is setenv.sh, so I assume you have to do source common/getCommonArgs.sh.
> Did you test this workflow at P2? You can check out a local clone with this PR to your home folder on the EPN, and then use the workflow configuration method = repository path pointing to your DATA path to test it, before merging to the official repository.
> 

Yes I did some pedestals calculation at P2 in my private folder, it works, clearly I need to tune better the different configurations and the write into the CCDB...

PR #529, opened 2022-07-12
Fix getTPCvdrift.C location in setenv_extra.sh of special TPC apass

ehellbar:
If I am mistaken and this was intentional, please ignore.

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1_TPCcalib/setenv_extra.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1_TPCcalib/setenv_extra.sh
@@ -49,8 +49,7 @@ else
 fi
 
 # TPC vdrift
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Hi,

i cannot check right now but it was working for me (unless I overlooked). Note also that the macro is the same in the two directories, so there is no problem. @catalinristea , did you see any issue with the current setting? Since i cannot test with the connection i have, i would rely on whthere it works or not.

Chiara

catalinristea:
Ciao @chiarazampolli,
The update is fine, as previously it was pointing to your private copy of O2DPG.
Thanks, Catalin.

chiarazampolli:
Argh! Sorry… checking the PR from the phone, i focused on the last part of the change, sorry.

Chiara

PR #531, opened 2022-07-13
Allow external setting of DPL_CONDITION_BACKEND

shahor02:


davidrohr:
Do you want this for online running?
This will currently not work at P2, since it is hard overwritten to localhost:8084 anyway.

martenole:
I think its for offline. But the way I see it this is not needed for offline, as in that case this script is never called.
https://alice.its.cern.ch/jira/browse/O2-3061

shahor02:
Hi, no, it is for local running 
https://alice.its.cern.ch/jira/browse/O2-3061?focusedCommentId=287901&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-287901

shahor02:
ok, removed both

PR #533, opened 2022-07-13
MFT: Fixing the json selection

tomas-herman:


martenole (approved):
Fine with me, but this does not change the logic, right? It only removes the duplicate setting of the QC json in case MFT RECO is disabled?

tomas-herman:
Hi, yes the logic is the same. But I noticed that the selection was not working for us which I assume was because of the extra line. Now the implementation is exactly the same as for TOF/MCH/MID for which the selection seems to be working fine.

DATA/production/qc-workflow.sh
```diff
@@ -11,13 +11,13 @@ elif [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
   if [[ $EPNSYNCMODE == 1 || "0$GEN_TOPO_LOAD_QC_JSON_FROM_CONSUL" == "01" ]]; then
     [[ -z "$QC_JSON_TPC" ]] && QC_JSON_TPC=consul://o2/components/qc/ANY/any/tpc-full-qcmn
     [[ -z "$QC_JSON_ITS" ]] && QC_JSON_ITS=consul://o2/components/qc/ANY/any/its-qcmn-epn-full
-    [[ -z "$QC_JSON_MFT" ]] && QC_JSON_MFT=consul://o2/components/qc/ANY/any/mft-full-qcmn
```
martenole, line 14: Ah right, because when you set `QC_JSON_MFT` here then the check below `[[ -z $QC_JSON_MFT ]]` is false and therefore you don't use the mft-track-full-qcmn even if MFT_RECO is enabled

PR #538, opened 2022-07-15
update setting for lhc22f/../j

noferini:
@catalinristea @chiarazampolli


DATA/production/configurations/2022/MayJunePilotBeam/apass1/selectSettings.sh
```diff
@@ -23,8 +23,85 @@ if [[ $RUNNUMBER -ge 519903 ]] && [[ $RUNNUMBER -le 520099 ]]; then
...
+fi
+#       FILL 7978: 520884 - 520904
+if [[ $RUNNUMBER -ge 520884 ]] && [[ $RUNNUMBER -le 520904 ]]; then
+  COLLISIONCONTEXT=collisioncontext_25ns_315b_302_237_240_48bpi_11inj.root
+fi
+
+
 # B field update
-if [[ $RUNNUMBER -ge 519041 ]]; then
+if [[ $RUNNUMBER -ge 999999 ]]; then
```
chiarazampolli, line 104: I don't understand this change, sorry. 
Also, we need to reflect the correct B field for QC.

noferini: What do you mean "for QC"?

chiarazampolli: QC might need to have the GRP in the local directory. Using a default would lead to wrong results. And there was anyway a default, so I don’t understand the change - unless i am overlooking something.

Anyway, due to QC, we need to keep this file updated with the running conditions, i fear.

noferini: If I put the last run of LHC22j, tomorrow it will be already obsolete before the PR is merged.

chiarazampolli: Then probably you wanted to have "-le" 999999? but then it will override what you have in line 86. You then need some `else`.
Sorry if I am not seeing the obvious. 
"-ge 999999" will never be true (for now at least).

noferini: I removed that part.
My problem was that I didn't understand the role of 

if [[ $RUNNUMBER -ge 519041 ]]; then
  GRPMAG=o2sim_grp_b5p_128HB.root
fi

in the previous version

chiarazampolli: It was leaving as default the latest B that had been set, overriding the default at the top of the script. The latest should always become the default.

DATA/production/configurations/2022/MayJunePilotBeam/apass1/selectSettings.sh
```diff
@@ -23,11 +23,83 @@ if [[ $RUNNUMBER -ge 519903 ]] && [[ $RUNNUMBER -le 520099 ]]; then
   COLLISIONCONTEXT=collisioncontext_Single_4b_2_2_2_noLR.root
 fi
 
-# B field update
-if [[ $RUNNUMBER -ge 519041 ]]; then
+# LHC22f: 520143 - 520473
+if [[ $RUNNUMBER -ge 520143 ]] && [[ $RUNNUMBER -le 520473 ]]; then
```
chiarazampolli, line 27: What about runs [519041, 520099]? The B field here was positive, like this, negative will be taken from line 5 (unless I am overlooking something).

chiarazampolli (changes requested):

noferini:
Dear all,
- I removed collisioncontext in the setting since it is no longer needed and impossible to be maintained here
- as default GRP I set GRPMAG=o2sim_grp_b5m_128HB.root if run is not matched

DATA/production/configurations/2022/MayJunePilotBeam/apass1/selectSettings.sh
```diff
@@ -1,35 +1,55 @@
 #!/bin/bash
 
+# default
+GRPMAG=o2sim_grp_b5m_128HB.root
```
chiarazampolli, line 4: Do we need this? Isn't is implicit in line 8?

chiarazampolli: Ah no, now I see, you want to cover the holes.

chiarazampolli (approved):

chiarazampolli:
Ciao @noferini , 

Also this one needs the same update:

https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/configurations/2022/LHC22f/apass1/selectSettings.sh

22f is processed with the file above, not the one in this PR - sorry that I overlooked.

Cheers,

Chiara

PR #542, opened 2022-07-16
Run only select list of global QC tasks, checking before that the inputs are actually created

davidrohr:


davidrohr:
currently broken, merging

PR #544, opened 2022-07-16
Add FIT,Calos to vertex matching, fix other sources

shahor02:
@sawenzel this is needed following the addition of FIT and Calo's to vertex matching, but I see that the core vertexing / AOD sources were incomplete (ITS-TPC-TRD-TOF was missing added) and are not really following detector composition. I think these sources, when not taken from the anchoring, should be concatenated to the seed by each workflow which produces corresponding output, as it is done in the dpl-workflow.sh.
We can discuss this next week.


sawenzel (approved):

sawenzel:
Tested it.

PR #547, opened 2022-07-22
RelVal major update

benedikt-voelkel:
ReleaseValidation.C

* restructure JSON output and add more fields such as used threshold and
  computed metric value per test and per histogram
* add reading of custom thresholds per histogram and per test
* only check once if comparable
* con-comparable (for all tests) if
  * at least one histogram is empty
  * different axis/binning
  * integral yields inf or nan
  * negative bin counts

o2dpg_release_validation.py

* add functionality to make histogram from log file content
  (e.g. used to extract number of TPC clusters from tpcreco.log)
* unifiy JSON output of overall summary including QC, MC kine etc. and
  single output from ROOT macro
* add summary plots comparing metric values and threshold per test and
  histogram
* add option --use-values-as-thresholds to inject previously computed
  matric values as new thresholds
* add comparison of 2 RelVal runs with command
  compare -i <relval/outpath1> <relval/outpath2>

### DIFF
--- a/RelVal/README.md
+++ b/RelVal/README.md
@@ -34,6 +34,7 @@ python o2dpg_release_validation.py [<sub-command>] --help
 The wrapper includes 3 different sub-commands for now
 1. `rel-val` to steer the RelVal,
 1. `inspect` to print histograms of specified severity (if any),
+1. `compare` to compare the results of 2 RelVal runs,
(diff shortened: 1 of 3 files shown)


chiarazampolli (approved):

chiarazampolli:
Error unrelated, merging.

PR #549, opened 2022-07-22
[QC-778] Centralized way to run QC post-processing with async reco

knopers8:
We use the workflow_runner framework, because:
- it allows us to execute the workflows in certain order, which is a required for some QC post-processing.
- we do so for post-processing with MC


knopers8:
@davidrohr , @chiarazampolli , @sawenzel do you have something against?

chiarazampolli (commented):
My only comment is that "pp" in the name of the python script sounds like "proton-proton", could this be changed? E.g. postproc, even if it becomes long, I know. 
The rest is fine with me.

knopers8:
that's a fair point, i will change it

chiarazampolli (approved):

PR #550, opened 2022-07-25
Fix sorting histograms for RelVal comparison

benedikt-voelkel:


benedikt-voelkel:
@chiarazampolli This is a small fix. The bug did not affect the actual RelVal but the comparison of 2 RelVals which I have also implemented to see how 2 RelVals compare to one another.
That is useful for our current investigation of whether we can bump G4 and related packages.

chiarazampolli (approved):

chiarazampolli:
Thanks @benedikt-voelkel !

PR #551, opened 2022-07-25
MFT: Sync QC json update

tomas-herman:
A PR to the QC repository is possibly failing due  the sync qc json file not being updated: https://github.com/AliceO2Group/QualityControl/pull/1359

### DIFF
--- a/DATA/production/qc-sync/mft.json
+++ b/DATA/production/qc-sync/mft.json
@@ -10,20 +10,20 @@
       },
       "Activity": {
         "number": "42",
-        "type": "2"
+        "type": "0"
       },
       "monitoring": {
         "url": "influxdb-unix:///tmp/telegraf.sock"
       },
       "consul": {
-        "url": "alio2-cr1-hv-head01:8500"
+        "url": "http://ali-consul.cern.ch:8500"
       },
       "conditionDB": {
         "url": "http://localhost:8084"
       }
     },
     "tasks": {
-      "QcMFTReadoutTask": {
+      "MFTReadoutTask": {
         "active": "true",
         "className": "o2::quality_control_modules::mft::QcMFTReadoutTask",
         "moduleName": "QcMFT",
@@ -41,20 +41,20 @@
         "remoteMachine": "any",
         "remotePort": "47798"
       },
-      "QcMFTDigitTask": {
+      "MFTDigitTask": {
         "active": "true",
         "className": "o2::quality_control_modules::mft::QcMFTDigitTask",
         "moduleName": "QcMFT",
         "detectorName": "MFT",
-        "cycleDurationSeconds": "60",
+        "cycleDurationSeconds": "300",
         "maxNumberCycles": "-1",
         "dataSource": {
           "type": "direct",
-          "query": "randomdigit:MFT/DIGITS/0"
+          "query": "randomdigit:MFT/DIGITS/0;digitsrof:MFT/DIGITSROF/0"
         },
         "taskParameters": {
           "FLP": "0",
-          "TaskLevel": "1"
+          "NoiseScan": "0"
         },
         "location": "local",
         "localMachines": [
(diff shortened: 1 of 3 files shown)


knopers8 (approved):

DATA/production/qc-sync/mft.json
```diff
@@ -82,26 +82,52 @@
         "remoteMachine": "alio2-cr1-qme05.cern.ch",
         "remotePort": "47797",
         "localControl": "odc"
+      },
+      "MFTAsyncTask": {
```
chiarazampolli, line 86: This task is called "Async", are you sure you want it in the "sync" json?

DATA/production/qc-sync/mft.json
```diff
@@ -82,26 +82,52 @@
...
+        "active": "true",
+        "className" : "o2::quality_control_modules::mft::QcMFTAsyncTask",
+        "moduleName": "QcMFT",
+        "detectorName": "MFT",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource" : {
+          "type" : "direct",
```
chiarazampolli, line 94: Is this correct, a dataSource of type "direct" in sync json?

DATA/production/qc-sync/mft.json
tomas-herman: Hi @chiarazampolli, the naming here is for historical reasons. Originally there were no tracks reconstructed online for muons (MFT, MCH, MID) but with the latest developments there is part of the MUON tracks reconstructed for ED and QC. For this reason we now include the task in sync QC but it is still called async task. We will come up with a new naming scheme but for now we kept the old name. 

Also I have been told during past discussions that the json file in the sync folder which I am modifying in this PR is used only for local tests and nothing else. Please let me know if this is not true. 

The file which is being actually used at P2 is located in Consul: https://ali-consul-ui.cern.ch/ui/alice-o2-cluster/kv/o2/components/qc/ANY/any/mft-track-full-qcmn/edit and we have been using that for many weeks now. 
This PR is just copy paste of the file from consul.

davidrohr: correct, the file here is only for the full system test in the CI and for local tests.
Operation at P2 takes the file from consul as defined here: https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/qc-workflow.sh#L14 (compare `EPNSYNCMODE=1` (P2) v.s. `SYNCMODE=1` (Sync FST))

DATA/production/qc-sync/mft.json
tomas-herman: Hi @chiarazampolli yes, we have been using a direct data source for this async task (it is essentially track task) in online QC at P2. The reason is that there is only a given fraction of the MUON tracks reconstructed for the purpose of ED and QC. Therefore if we want to reduce the data flow it makes sense to reduce the reconstruction fraction. 

In the future if it is considered convenient to reduce the QC sampling rather than the reconstruction ration we can also introduce sampling in QC.

chiarazampolli: Thanks, fine with me then.

DATA/production/qc-sync/mft.json
tomas-herman: Great, thank you for confirming.

chiarazampolli: Sure, my point was more that if this is the copy of the one used in sync, and we had there something for async, it would have been wrong. But after the explanation by @tomas-herman , all is clear.

chiarazampolli:
CI reports no token, and this PR cannot then be tested (but it should/could according to @davidrohr 's comment). @TimoWilken , do you have any idea?

TimoWilken:
Hi @chiarazampolli, my mistake (the key file had incorrect permissions). Fixed now, and I've restarted the check.

davidrohr:
@tomas-herman : The FST fails with
```
[ERROR] invalid workflow in o2-qc: No matching output found for <matcher query: (and origin:MFT (and description:MFTTrackROF (and subSpec:0 (just startTime:$0 ))))> as requested by data processor "qc-task-MFT-MFTAsyncTask". Candidates:
```

PR #554, opened 2022-07-28
FT0 calibration: standalone time offset calib

afurs:


chiarazampolli:
Seems ok to me, @davidrohr , can you also take a look?

PR #556, opened 2022-07-28
Workflow parser, parallelize subtopology generation

davidrohr:


davidrohr:
not tested in FST anyway, merging

PR #563, opened 2022-08-03
Adding TPC IDCs specs

chiarazampolli:
@davidrohr @ehellbar @rmunzer @matthias-kleiner


DATA/common/setenv_calib.sh
```diff
@@ -41,6 +41,7 @@ if [[ $BEAMTYPE != "cosmic" ]] || [[ $FORCECALIBRATIONS == 1 ]] ; then
       if [[ -z ${CALIB_TPC_TIMEGAIN+x} ]]; then CALIB_TPC_TIMEGAIN=1; fi
       if [[ -z ${CALIB_TPC_RESPADGAIN+x} ]]; then CALIB_TPC_RESPADGAIN=1; fi
     fi
+    if [[ -z ${CALIB_TPC_IDC+x} ]]; then CALIB_TPC_IDC=1; fi
```
davidrohr, line 44: perhaps we should leave the default at 0 for now, until it has been tested?

chiarazampolli: OK

PR #564, opened 2022-08-03
Fix missing input/output proxy spec for TPC VD TGL calib

davidrohr:


shahor02:
Thanks! Was testing a local aggregator and missed the proxy...

PR #565, opened 2022-08-04
Align to recent changes and extension

njacazio:
@chiarazampolli @strogolo this for me worked on LHC22e data, I also added some improvement to the analysis script

### DIFF
--- a/MC/analysis_testing/analysis_test.sh
+++ b/MC/analysis_testing/analysis_test.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env bash
 
 # This script performs an analysis task (given as first argument)
 # in a directory (produced) where a merged AO2D.root was produced
--- a/MC/analysis_testing/o2dpg_analysis_test_workflow.py
+++ b/MC/analysis_testing/o2dpg_analysis_test_workflow.py
@@ -31,7 +31,7 @@
 #                         pass name
 #   --period-name PERIOD_NAME
 #                         prodcution tag
-#   --config CONFIG       overwrite thee default config JSON. Pass as json://<path/to/file>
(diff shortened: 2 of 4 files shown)


njacazio:
I also tested the multiplicity table and for me it works

njacazio:
pinging also @benedikt-voelkel to check that this is to your taste

benedikt-voelkel:
Hej @njacazio ,
shall we change all WFs to lists as you did for EventTrackQA? I will have a closer look tomorrow also

njacazio:
> Hej @njacazio , shall we change all WFs to lists as you did for EventTrackQA? I will have a closer look tomorrow also

Now it works with both syntax I don't have opinions

njacazio:
@strogolo if you could also crosscheck that it works it would be great

chiarazampolli:
Thanks @njacazio , it looks ok to me but let's wait for @benedikt-voelkel . 
One question: the multiplicity task now worked with a dummy entry, or what exactly?

njacazio:
> Thanks @njacazio , it looks ok to me but let's wait for @benedikt-voelkel . One question: the multiplicity task now worked with a dummy entry, or what exactly?

It works with the pr that avoids the fatal

strogolo:
Hi @njacazio, the check I can do is the same you did...trying running the python script that you and Benedikt have put in O2DPG. Nevertheless, as long as the QC tasks do not check whether the object is empty or not, this solution can work. The issue might be when a check on the content of the object will be introduced...but it's not the current case.

@chiarazampolli: shall I stop the implementation of a dummy object (i.e. flat at unity)?

chiarazampolli:
Ciao @strogolo , 
I would still prefer to have it, since we are currently running with old tags, where the change by Nicolo' to avoid teh fatal is not included. 
Chiara

benedikt-voelkel (approved):
Ok from my side, thanks @njacazio

PR #577, opened 2022-08-17
Use utf-8 encoding

benedikt-voelkel:


chiarazampolli:
Hi @benedikt-voelkel , 
What was the issue with the current version?

benedikt-voelkel:
I found some characters in the TPC reco log files during RelVal that do not conform with ASCII encoding. So I am setting it to `utf-8` explicitly

chiarazampolli:
Thanks for the explanation!

chiarazampolli (approved):

PR #578, opened 2022-08-17
adjust GRPECS meta-field for run number

sawenzel:
it seems to be 'runNumber' and not 'run_number' (like sometime in the past)


sawenzel:
@catalinristea : I think this is what you also noticed, right?

chiarazampolli:
Yes, it was discussed here https://alice.its.cern.ch/jira/browse/O2-3078.
Chiara

PR #582, opened 2022-08-19
Remove extra tolerances for non-anchored MCs

fgrosa:


MC/bin/o2dpg_sim_config.py
```diff
@@ -53,21 +53,52 @@ def add(cfg, flatconfig):
...
+          add(config, {"pvertexer.timeMarginTrackTime" : 0.,
+                       "pvertexer.dbscanDeltaT" : 7,
                        "pvertexer.maxChi2TZDebris": 50,
                        "pvertexer.maxMultRatDebris": 1.,
                        "pvertexer.dbscanAdaptCoef" : 20,
                        "pvertexer.maxVerticesPerCluster" : 20,
-                       "pvertexer.dbscanMaxDist2" : 36})
-
+                       "pvertexer.dbscanMaxDist2" : 36,
+                       "ITSCATrackerParam.sysErrY2[0]": 0.,
```
mpuccio, line 64: This is by default 0, it should not necessary to be set. Did you check that the workflow.json contains this? In my local test it doesn't, and the configuration pointed by Ruben is for apass of data.

MC/bin/o2dpg_sim_config.py
```diff
@@ -53,21 +53,52 @@ def add(cfg, flatconfig):
...
+                       "ITSCATrackerParam.sysErrY2[4]": 0.,
+                       "ITSCATrackerParam.sysErrY2[5]": 0.,
+                       "ITSCATrackerParam.sysErrY2[6]": 0.,
+                       "ITSCATrackerParam.sysErrZ2[0]": 0.,
+                       "ITSCATrackerParam.sysErrZ2[1]": 0.,
+                       "ITSCATrackerParam.sysErrZ2[2]": 0.,
+                       "ITSCATrackerParam.sysErrZ2[3]": 0.,
+                       "ITSCATrackerParam.sysErrZ2[4]": 0.,
+                       "ITSCATrackerParam.sysErrZ2[5]": 0.,
+                       "ITSCATrackerParam.sysErrZ2[6]": 0.})
```
shahor02, line 101: @fgrosa note that this will also affect the ITS tracking and matching to TPC (less fakes will be produced).
In general, we try to use for anchored MC the same settings as used for data. Is it worth changing these settings globally? Perhaps one should do first a dedicated production to test its effect? 
Relaxed timing tolerances lead to false-positive `ambiguous` tracks, which "only" increases the combinatorial background but should not lead to inefficiency (once ambiguities are accounted in the analysis).
Also, note this fix: https://alice.its.cern.ch/jira/browse/O2-3054?focusedCommentId=289529&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-289529, thought it was affecting mostly MFT, for the ITS it effectively added <1\mus extra tolerance.

mpuccio (changes requested):

MC/bin/o2dpg_sim_config.py
fgrosa: Indeed you are right, I did not check the default values (that are 0), so they do not need to be set (reverted)

MC/bin/o2dpg_sim_config.py
fgrosa: Hi @shahor02 thanks a lot! I removed the settings in the case of `302000 <= run <= 309999` which is the one for the high energy pp program. In general, I agree with you, however HF and DQ vertexing codes are not yet designed to account for ambiguous tracks and it was suggested by @mpuccio and @ginnocen to look at the inefficiency that we obtain removing this extra tolerances.

chiarazampolli:
Hello @shahor02 , @fgrosa , @mpuccio , 

To me this is ok. Any further comment?

Chiara

mpuccio:
It's fine with me!

fgrosa:
> Hello @shahor02 , @fgrosa , @mpuccio ,
> 
> To me this is ok. Any further comment?
> 
> Chiara

Also with me, thanks!

cterrevo:
Hello @chiarazampolli, @shahor02, @fgrosa, @mpuccio, 
it seems to me that we all agreed on the proposed modification.  Could we go ahead with the merge of this PR, so that we can start again the 10% MC production?  thanks a lot, Cristina

shahor02:
sorry for late answer, fine for me too.

chiarazampolli (approved):

PR #586, opened 2022-08-22
add anchored MC setup for 2022 JUN pilot beam

sawenzel:


sawenzel:
@chiarazampolli @catalinristea : I am merging this (some iterations might still be required). We should see if in the future we still need one such script for each anchoring period or if we can have one generic version.

PR #589, opened 2022-08-24
ITS Calibration scans in O2DPG

iravasen:
Do not merge yet, waiting for the test.


iravasen:
@davidrohr this is to include/update the ITS calibration workflows in O2DPG as requested. We tested all of them and everything seems to work.

PR #592, opened 2022-08-25
vdrift correction should be wrt refVDrift, when different from 0

chiarazampolli:
@shahor02 : this should followed our discussion from yesterday. Please, check that it is indeed what you meant.

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/getTPCvdrift.C
+++ b/DATA/production/configurations/2022/LHC22f/apass1/getTPCvdrift.C
@@ -52,7 +52,11 @@ float getTPCvdrift(int run, std::string_view ltrUrl = "http://alice-ccdb.cern.ch
   c.init(ltrUrl.data());
   const auto ltrCalib = c.retrieveFromTFileAny<o2::tpc::LtrCalibData>(calibType.data(), metadata, minTime); /// timestamp in the run of interest
   const auto corr = ltrCalib->getDriftVCorrection();
-  const float vcorr = defaultDriftV / corr;
+  float vcorr = defaultDriftV / corr;
+  if (ltrCalib->refVDrift != 0) {
+    printf("refVDrift different from zero: %f (default was %f)\n", ltrCalib->refVDrift, defaultDriftV);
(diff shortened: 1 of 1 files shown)


DATA/production/configurations/2022/LHC22f/apass1/getTPCvdrift.C
```diff
@@ -53,6 +53,10 @@ float getTPCvdrift(int run, std::string_view ltrUrl = "http://alice-ccdb.cern.ch
   const auto ltrCalib = c.retrieveFromTFileAny<o2::tpc::LtrCalibData>(calibType.data(), metadata, minTime); /// timestamp in the run of interest
   const auto corr = ltrCalib->getDriftVCorrection();
   const float vcorr = defaultDriftV / corr;
+  if (ltrCalib->refVDrift != 0) {
+    printf("refVDrift different from zero: %f\n", ltrCalib->refVDrift);
+    vcorr = refVDrift / corr;
+  }
   printf("vdrift = %f\n", vcorr);
```
shahor02, line 60: Yes in spirit, but you cannot reassign the `const float vcorr`.

chiarazampolli: Ah yes, sorry :-) I will fix it.

chiarazampolli: Fixed.
Note, though, that I tried it for run 523148 from 22m, and it accessed the object in https://alice-ccdb.cern.ch/browse/TPC/Calib/LaserTracks/1660197371503, created on 3.08. Here the refVdrift is the same as the default. Does it mean that from some point in time on it is always set equal to the default? And it was zero only long ago?

shahor02:
From epn-20220711 the reference is stored (i.e. 2.58 unless was modified from the command line) and from 
epn-20220730 the reference is normalized to extracted vdrift and the correction is set to 1.

chiarazampolli:
Ok, so the current version is ok, right? In all cases.

shahor02 (approved):
yes

PR #599, opened 2022-09-01
to set the possibility to use one node for TPC IDCs

chiarazampolli:
@davidrohr : this is just to set if we run the IDCs on both sides together or not from the setenv_calib. I put it here to be consistent with the rest: each calibration is on or off in this script.

### DIFF
--- a/DATA/common/setenv_calib.sh
+++ b/DATA/common/setenv_calib.sh
@@ -44,7 +44,14 @@ if [[ $BEAMTYPE != "cosmic" ]] || [[ $FORCECALIBRATIONS == 1 ]] ; then
     if ( has_detectors ITS TPC && has_detector_matching ITSTPC ); then
       if [[ -z ${CALIB_TPC_VDRIFTTGL+x} ]]; then CALIB_TPC_VDRIFTTGL=1; fi
     fi
-    if [[ -z ${CALIB_TPC_IDC+x} ]]; then CALIB_TPC_IDC=0; fi # default is off
+    # IDCs
+    if [[ -z ${CALIB_TPC_IDC+x} ]]; then
+      CALIB_TPC_IDC=0; # default is off
+    else
+      if [[ -z {$CALIB_TPC_IDC_BOTH+x} ]]; then
+	CALIB_TPC_IDC_BOTH=0;
+      fi # by default, A and C side are processed separately
+    fi
   fi
 
   # calibrations for TRD
@@ -104,13 +111,20 @@ if [[ "0$GEN_TOPO_VERBOSE" == "01" ]]; then
   echo "CALIB_TPC_TIMEGAIN = $CALIB_TPC_TIMEGAIN" 1>&2
   echo "CALIB_TPC_RESPADGAIN = $CALIB_TPC_RESPADGAIN" 1>&2
   echo "CALIB_TPC_IDC = $CALIB_TPC_IDC" 1>&2
+  if [[ $CALIB_TPC_IDC == 1 ]]; then
+    if [[ ! -z $CALIB_TPC_IDC_BOTH == 1 ]]; then
+      echo "CALIB_TPC_IDC_BOTH = $CALIB_TPC_IDC_BOTH" 1>&2
+    else
+      echo "CALIB_TPC_IDC_BOTH not set, A and C side will be run separately"
+    fi
+  fi
   echo "CALIB_CPV_GAIN = $CALIB_CPV_GAIN" 1>&2
 fi
 
 # define spec for proxy for TF-based outputs from BARREL
 if [[ -z $CALIBDATASPEC_BARREL_TF ]]; then
   # prim vtx
-  if [[ $CALIB_PRIMVTX_MEANVTX == 1 ]] ; then add_semicolon_separated CALIBDATASPEC_BARREL_TF "pvtx:GLO/PVTX/0"; fi
+  if [[ $CALIB_PRIMVTX_MEANVTX == 1 ]]; then add_semicolon_separated CALIBDATASPEC_BARREL_TF "pvtx:GLO/PVTX/0"; fi
 
   # TOF
   if [[ $CALIB_TOF_LHCPHASE == 1 ]] || [[ $CALIB_TOF_CHANNELOFFSETS == 1 ]]; then add_semicolon_separated CALIBDATASPEC_BARREL_TF "calibTOF:TOF/CALIBDATA/0"; fi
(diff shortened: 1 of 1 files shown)


DATA/common/setenv_calib.sh
```diff
@@ -104,13 +113,20 @@ if [[ "0$GEN_TOPO_VERBOSE" == "01" ]]; then
   echo "CALIB_TPC_TIMEGAIN = $CALIB_TPC_TIMEGAIN" 1>&2
   echo "CALIB_TPC_RESPADGAIN = $CALIB_TPC_RESPADGAIN" 1>&2
   echo "CALIB_TPC_IDC = $CALIB_TPC_IDC" 1>&2
+  if [[ $CALIB_TPC_IDC == 1 ]]; then
+    if [[ ! -z $CALIB_TPC_IDC_BOTH == 1 ]]; then
+      echo "CALIB_TPC_IDC_BOTH = $CALIB_TPC_IDC_BOTH" 1>&2
+    else
+      echo "CALIB_TPC_IDC_BOTH not set, A and C side will be run separately"
```
davidrohr, line 120: `1>&2` missing

DATA/common/setenv_calib.sh
```diff
@@ -11,11 +11,30 @@
...
 
+# by default, all calibrations are disabled, because we don't know which detectors are in the data taking
+[[ -z ${CALIB_PRIMVTX_MEANVTX} ]] && CALIB_PRIMVTX_MEANVTX=0
+[[ -z ${CALIB_TOF_LHCPHASE} ]] && CALIB_TOF_LHCPHASE=0
+[[ -z ${CALIB_TOF_CHANNELOFFSETS} ]] && CALIB_TOF_CHANNELOFFSETS=0
+[[ -z ${CALIB_TOF_DIAGNOSTICS} ]] && CALIB_TOF_DIAGNOSTICS=0
+[[ -z ${CALIB_TPC_SCDCALIB} ]] && CALIB_TPC_SCDCALIB=0
+[[ -z ${CALIB_TPC_TIMEGAIN} ]] && CALIB_TPC_TIMEGAIN=0
+[[ -z ${CALIB_TPC_RESPADGAIN} ]] && CALIB_TPC_RESPADGAIN=0
+[[ -z ${CALIB_TPC_IDC} ]] && CALIB_TPC_IDC=0
```
davidrohr, line 22: you removed the force-disable here, in case TPC is not present.
But this is needed.
Otherwise the current cosmics, which have CALIB_TPC_IDC=1 hardcoded will fail, since they run out of calib nodes.
I think, in this default setting, we should check for all detectors whether they are present, and if not override it to =0 even if it is already set to =1 (like I had done below for TPC, what you removed now)

davidrohr (changes requested):

chiarazampolli:
Ciao @davidrohr , 
I modified the logic, defining initially if calibrations can be run at all or not. What do you think? This allows to not repeat the conditions in the code, which I found not very nice. Let me know.

DATA/common/setenv_calib.sh
```diff
@@ -15,79 +15,109 @@ SOURCE_GUARD_SETENV_CALIB=1
 # Remove this line to only send unbinned residuals
 if [[ -z "$CALIB_TPC_SCDCALIB_SENDTRKDATA" ]];  then export CALIB_TPC_SCDCALIB_SENDTRKDATA=1; fi
 
+# define the conditions for each calibration
+if has_detector_calib ITS && has_detectors_reco ITS && has_detector_matching PRIMVTX && [[ ! -z "$VERTEXING_SOURCES" ]]; then CAN_DO_CALIB_PRIMVTX_MEANVTX=1; else CAN_DO_CALIB_PRIMVTX_MEANVTX=1; fi
```
martenole, line 19: is this a typo? shouldn't it be `then CAN_DO_CALIB_PRIMVTX_MEANVTX=1; else CAN_DO_CALIB_PRIMVTX_MEANVTX=0; fi`?

DATA/common/setenv_calib.sh
```diff
@@ -15,79 +15,109 @@ SOURCE_GUARD_SETENV_CALIB=1
...
+if has_detector_calib TOF && has_detector_reco TOF; then CAN_DO_CALIB_TOF_DIAGNOSTICS=1; else CAN_DO_CALIB_TOF_DIAGNOSTICS=0; fi
+if has_detector_calib TOF && has_detector_reco TOF && (( has_detectors_reco ITS TPC && has_detector_matching ITSTPCTOF ) || ( has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRDTOF )); then CAN_DO_CALIB_TOF_LHCPHASE=1; CAN_DO_CALIB_TOF_CHANNELOFFSETS=1; else CAN_DO_CALIB_TOF_LHCPHASE=0; CAN_DO_CALIB_TOF_CHANNELOFFSETS=0; fi
+if has_detector_calib TPC && has_detectors ITS TPC TOF TRD && has_detector_matching ITSTPCTRDTOF; then CAN_DO_CALIB_TPC_SCDCALIB=1; else CAN_DO_CALIB_TPC_SCDCALIB=0; fi
+if has_detector_calib TPC && has_processing_step TPC_DEDX; then CAN_DO_CALIB_TPC_TIMEGAIN=1; CAN_DO_CALIB_TPC_RESPADGAIN=1; else CAN_DO_CALIB_TPC_TIMEGAIN=0; CAN_DO_CALIB_TPC_RESPADGAIN=0; fi
+if has_detector_calib TPC && has_detectors ITS TPC && has_detector_matching ITSTPC; then CAN_DO_CALIB_TPC_VDRIFTTGL=1; else CAN_DO_CALIB_TPC_VDRIFTTGL=0; fi
+if has_detector_calib TPC; then CAN_DO_CALIB_TPC_IDC=1; else CAN_DO_CALIB_TPC_IDC=1; fi
```
martenole, line 25: also here `CAN_DO_CALIB_TPC_IDC` is set to 1 in any case

chiarazampolli: Yes, thanks for noticing.

DATA/common/setenv_calib.sh
chiarazampolli: Yes, thanks for noticing.

chiarazampolli:
Thanks @martenole ! I will wait till tonight in case @davidrohr takes a look, then I will merge.

chiarazampolli:
Merging.

PR #600, opened 2022-09-01
Calib workflow - force disable TPC IDC if TPC is not present

davidrohr:


chiarazampolli:
Thanks @davidrohr ! We can set them all to xero at the beginning, then the checks will switch on what should and could be run. This would fix the issue at P2. Should i give it a try?

davidrohr:
@chiarazampolli : actually, I would do it not at the beginning.
I would, for all detectors, switch them off in the way I have done it here for TPC IDC..
Setting the default to =0 will not be enough if the detector is missing. Since today, in case of TPC, the problem was that they added `CALIB_TPC_IDC=1` to the env variables, which force-enabled it irrespective of the defaults.
And then they started a run without TPC.
So I think the best is just force-disabling at the end, as I have done here.
I was actually going to ask you if you want to take of the others, but then I left for going climbing first :)

chiarazampolli:
Ciao @davidrohr , I missed your reply, and started already implementing as I proposed. 
I don't understand why you say that it would not work. When we enable the calibrations, we always check if a detector is in the data taking. So each calibration will then be set only if this is true. 
See https://github.com/AliceO2Group/O2DPG/pull/599/commits/87a172ae44b51400b223dca7cfb7fb3a577dd91c. 
I can change, but I think that checking things like "has_detector_calib ITS" (just one example) 2x is not needed.

davidrohr:
But if they force-enable a calibration from the ECS GUI, the value is already set to =1, and it will not be disabled.
Therefore, I put the force-disable.
I think the logic before was fine, except for the case when you have an external force-enable for a missing detector.
This is kind of operator error obviously.
But still, we can just catch it and force-disable if the detector is missing.

chiarazampolli:
Ok, I didn’t think about the setting in ECS. Then i prefer an else. I will change and you tell me what you think.

PR #601, opened 2022-09-02
added ITS qc as optional for calib

iravasen:
Hi @davidrohr this should make the fix discussed via email. Thanks for the suggestion!


chiarazampolli:
I think all is good in this PR. @davidrohr might not be available today, so if he does not approve it, I will do so by tonight.

chiarazampolli (approved):

PR #602, opened 2022-09-02
New feature: Task-level software environments

sawenzel:
The graph pipeline runner can apply a specific
alienv/CVFMS software version on an individual task level.

This allows to mix multiple software versions during execution, which
is actually a benefit of the multi-process (microservice) approach.

Software packages can be defined in the workflow json under
the tag "alternative_alienv_package" as part of a task specification.

The new feature is for now (optionally) applied in the O2DPG MC workflow creation,
to all tasks that are RECO (or further) in the processing. This can be enabled
via the new "--alternative-reco-software" option.

### DIFF
--- a/MC/bin/o2_dpg_workflow_runner.py
+++ b/MC/bin/o2_dpg_workflow_runner.py
@@ -445,6 +445,43 @@ def update_resource_estimates(workflow, resource_json):
         #        cpu *= rel_cpu
         #    task["resources"]["cpu"] = cpu
 
+# a function to read a software environment determined by alienv into
+# a python dictionary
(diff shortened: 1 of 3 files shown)


sawenzel:
@chiarazampolli @catalinristea : This is an implementation of the 2-software version feature. One simple needs
to add `--alternative-reco-software O2sim/v20220601-1` to the workflow creation, and then all reco tasks will be executed with this CVMFS software version. (The package will finally come from a new JDL variable).

The  good news is that we don't actually have to do this in a serial fashion: - "do sim" - "switch software" - "do reco". It all works parallel the same way as before.

PR #604, opened 2022-09-05
Fix nuclei injection

njacazio:


njacazio:
Hi, I added for convenience the clang format of the O2, I can remove it of course and keep it locally if it is not needed

njacazio:
@rutuparnarath

.clang-format
```diff
@@ -0,0 +1,56 @@
+BasedOnStyle: Google
```
sawenzel, line 1: Please don't commit this file.

njacazio: It's ok, that's why I asked ;)

gconesab (approved):

.clang-format
chiarazampolli: I agree.

njacazio: It should be already gone, no?

PR #610, opened 2022-09-07
Enabled FullClusterScan for MFT tracking

sarahherrmann:
New default for pp collisions : full cluster scan for MFT tracking

### DIFF
--- a/MC/bin/o2dpg_sim_config.py
+++ b/MC/bin/o2dpg_sim_config.py
@@ -1,7 +1,7 @@
 def create_sim_config(args):
-#creates a generic simulation config
-#based on arguments args(run number, energy, ...) originally passed
-#to o2dpg_sim_workflow.py
+    # creates a generic simulation config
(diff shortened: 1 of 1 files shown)


MC/bin/o2dpg_sim_config.py
```diff
@@ -1,7 +1,7 @@
 def create_sim_config(args):
-    # creates a generic simulation config
-    # based on arguments args (run number, energy, ...) originally passed
-    # to o2dpg_sim_workflow.py
+#creates a generic simulation config
+#based on arguments args(run number, energy, ...) originally passed
+#to o2dpg_sim_workflow.py
```
sawenzel: please don't change non-relevant sections (such as comments)

sawenzel (approved):

MC/bin/o2dpg_sim_config.py
sawenzel: I see .. you have reverted this

PR #612, opened 2022-09-09
TPC Laser trigger mode

rmunzer:


DATA/production/calib/tpc-laser.sh
```diff
@@ -63,21 +67,23 @@ o2-dpl-raw-proxy $ARGS_ALL \
...
     --send-ce-digits \
+    --decoder-type 0 \
     | o2-tpc-reco-workflow $ARGS_ALL \
     --input-type digitizer  \
     --output-type "tracks,disable-writer" \
     --disable-mc \
     --pipeline tpc-zsEncoder:20,tpc-tracker:8 \
     $GPU_CONFIG \
-    --configKeyValues "$ARGS_ALL_CONFIG;align-geom.mDetectors=none;GPU_global.deviceType=$GPUTYPE;GPU_proc.tpcIncreasedMinClustersPerRow=500000;GPU_proc.ignoreNonFatalGPUErrors=1;$GPU_CONFIG_KEY" \
+    --condition-remap file:///home/wiechula/processData/inputFilesTracking/triggeredLaser/=GLO/Config/GRPECS \
```
chiarazampolli, line 77: How do you maintain this entry when you run online?

DATA/testing/detectors/TPC/tpc-idc-merger.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
```
chiarazampolli, line 1: This file is committed on purpose, right?

DATA/production/calib/tpc-laser.sh
wiechula: We only need a dummy entry to allow for triggered processing, so it will not change.

chiarazampolli (approved):

PR #615, opened 2022-09-12
Fix QC CCDB external server overwrite

martenole:


davidrohr (commented):
Why is this needed?
`GEN_TOPO_QC_OVERRIDE_CCDB_SERVER` should not contain `'` or `,`, so I don't see how it can fail.

Also, don't we introduce problems when we end the string by `'`? E.g. what happens if `GEN_TOPO_QC_OVERRIDE_CCDB_SERVER=" "`

martenole:
Its needed, otherwise you always overwrite with literally $GEN_TOPO_QC_OVERRIDE_CCDB_SERVER...
```47:qc-task-CPV-PhysicsOnEPNs]: [21:06:26][ALARM] Curl request to $GEN_TOPO_QC_OVERRIDE_CCDB_SERVER/CPV/Calib/Gains/1663009586642/ failed with result 6, response code```

**Comment by martenole** [comment]
> Also, don't we introduce problems when we end the string by '? E.g. what happens if GEN_TOPO_QC_OVERRIDE_CCDB_SERVER=" "
Hm, then you would get a `sed: -e expression #1, char 28: unterminated `s' command`.
But would not that be an operator error actually?

**Comment by davidrohr** [comment]
Ah, sure, you are right, since the single quotes do not expand. But then I think the right fix is simply (i.e. we don't need to end the string)
```
sed -i "s,http://alice-ccdb.cern.ch,$GEN_TOPO_QC_OVERRIDE_CCDB_SERVER,g" $MERGED_JSON_FILENAME
```

martenole:
Kicking the CI again. The error was not related to this PR. Running the FST manually went without issues. This should be good to go

martenole:
Failure is unrelated (see https://github.com/AliceO2Group/QualityControl/pull/1443#issuecomment-1246293226)
Could this be merged please @chiarazampolli ?

martenole:
Thanks a lot Chiara

PR #619, opened 2022-09-13
Fix Chi2 bin scaling

benedikt-voelkel:


benedikt-voelkel:
@chiarazampolli @sawenzel  Could we just merge this?

PR #626, opened 2022-09-16
Fix ITS QC json

martenole:
@JianLIUhep 
Without this tests on staging are failing


martenole:
I cannot merge without review. @chiarazampolli or @shahor02 could you approve this please? With tabs in the json file instead of spaces the parameters cannot be parsed and the QC task fails initializing. Tested it on the EPNs with a local O2DPG that this works

shahor02 (approved):

PR #631, opened 2022-09-19
[EMCAL-756] Async QC at cell level must use subspec 0

mfasDa:
Once the recalibrator will be included cells will be
available on 2 supsecifications: 1 - uncalibrated, 0 -
calibrated. Only calibrated cells must be monitored
by the cell-level QC during the asynchronous stage.

### DIFF
--- a/DATA/production/qc-async/emc.json
+++ b/DATA/production/qc-async/emc.json
@@ -32,7 +32,7 @@
         "maxNumberCycles": "-1",
         "dataSource": {
           "type": "direct",
-          "query": "emcal-cells:EMC/CELLS;emcal-triggerecords:EMC/CELLSTRGR"
+          "query": "emcal-cells:EMC/CELLS/0;emcal-triggerecords:EMC/CELLSTRGR/0"
         }
       }
     }


mfasDa:
For the moment the commit is trivial since during the asynchronous stage cells are at the moment only published on subspec 0, so the fixed subspec and the ConcreteDataTypeMatcher will lead to the same 
result. However we are preparing to include the recalibration workflow into the asynchronous reconstruction which will modify the cell collection. It was agreed here that the CTF reader will publish for EMCAL on subspec 1 and the recalibrator will publish on subspec 0, leading to the QC seeing both calibrated and uncalibrated cells when applying the ConcreteDataTypeMatcher. Since we only want to monitor calibrated cells during the asynchronous stage we need to force the subspec.

DATA/production/qc-async/emc.json
```diff
@@ -32,7 +32,7 @@
         "maxNumberCycles": "-1",
         "dataSource": {
           "type": "direct",
-          "query": "emcal-cells:EMC/CELLS;emcal-triggerecords:EMC/CELLSTRGR"
+          "query": "emcal-cells:EMC/CELLS:0;emcal-triggerecords:EMC/CELLSTRGR:0"
```
martenole, line 35: Are you sure that `EMC/CELLS:0` is correct and that it should not be `EMC/CELLS/0`? I don't know, but maybe you tested it already that it works like this?

mfasDa: Sorry, indeed, copy/paste error. Yes, it was tested offline but then I screwed up porting to O2DPG. Thanks for spotting! Will fix immediately!

martenole (approved):
Should this be merged then?

chiarazampolli (approved):

PR #632, opened 2022-09-19
make NEntries test critical when dealing with TEfficiency

DanielSamitz:
small changes in ExtractAndFlatten.C and ReleaseValidation.C to detect if dealing with an efficiency plot in the NEntries test and make it critical in that case.

### DIFF
--- a/RelVal/ExtractAndFlatten.C
+++ b/RelVal/ExtractAndFlatten.C
@@ -268,7 +268,7 @@ void WriteTEfficiency(TEfficiency* hEff, TDirectory* outDir, std::string const& 
   // recreate the efficiency dividing numerator for denominator:
   auto heff = (TH1*)(hEffNomin->Clone("heff"));
   heff->SetTitle(Form("%s", hEff->GetTitle()));
-  heff->SetName(Form("%s", hEff->GetName()));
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
@DanielSamitz 
I was thinking and there are 2 things:
1. Should we maybe rather check really the suffix instead finding any occurrence of `_efficiency`?
1. In addition maybe something that is more unlikely to exist in the first place? Because some users/tasks might have an object ending with `_efficiency`? Maybe we could choose something like `_effNEntriesCrit`?

DanielSamitz:
@benedikt-voelkel 

That is a very good point that it should be something that is not very likely to occur in any name of an input object. But the histogram name will also show up in all overlay plots, json output, summary plots, etc.. , so I think it should still be something meaningful to the user, and something like `_effNEntriesCrit` just refers to an internal information that should maybe not appear in a histogram name. Maybe something like `_ratioFromTEfficiency` (and likewise change also  `_numeratorFromTEfficiency` and `_denominatorFromTEfficiency` for the other two)?

benedikt-voelkel:
They seem like reasonable suggestions indeed. That also shows that these are somehow derived histogram objects, might be good for a user to know

PR #634, opened 2022-09-19
Adjust check if ZDC calib is possible

martenole:
@shahor02 and @chiarazampolli could we not simply change the check here so that we cannot have an inconsistent configuration?


chiarazampolli (approved):

chiarazampolli:
I think that this is indeed the correct things to do: enabling should be done when the needed features are also enabled.

DATA/common/setenv_calib.sh
```diff
@@ -27,7 +27,7 @@ if has_detector_calib TRD && has_detectors ITS TPC TRD; then CAN_DO_CALIB_TRD_VD
 if has_detector_calib EMC && has_detector_reco EMC; then CAN_DO_CALIB_EMC_BADCHANNELCALIB=1; CAN_DO_CALIB_EMC_TIMECALIB=1; else CAN_DO_CALIB_EMC_BADCHANNELCALIB=0; CAN_DO_CALIB_EMC_TIMECALIB=0; fi
 if has_detector_calib PHS && has_detector_reco PHS; then CAN_DO_CALIB_PHS_ENERGYCALIB=1; CAN_DO_CALIB_PHS_BADMAPCALIB=1; CAN_DO_CALIB_PHS_TURNONCALIB=1; CAN_DO_CALIB_PHS_RUNBYRUNCALIB=1; else CAN_DO_CALIB_PHS_ENERGYCALIB=0; CAN_DO_CALIB_PHS_BADMAPCALIB=0; CAN_DO_CALIB_PHS_TURNONCALIB=0; CAN_DO_CALIB_PHS_RUNBYRUNCALIB=0; fi
 if has_detector_calib CPV && has_detector_reco CPV; then CAN_DO_CALIB_CPV_GAIN=1; else CAN_DO_CALIB_CPV_GAIN=0; fi
-if has_detector_calib ZDC && has_detector_reco ZDC; then
+if has_detector_calib ZDC && has_processing_step ZDC_RECO; then
```
shahor02, line 30: As a temporary measure, this is ok, but since detectors reconstruction in sync mode is conditioned by `has_detector_reco`, we should probably revert it once the ZDC is permanently enabled.

martenole: I agree. We should revert this in parallel to when we change the dpl-workflow to enable ZDC permanently

chiarazampolli: Should we not have then `has_detector_reco ZDC && has_processing_step ZDC_RECO`? The most stringent condition should be put here.

martenole: Theoretically yes, but practically `has_detector_reco ZDC` does not have any practical effect. So I think for now it is fine. And later on we can just revert this commit.

PR #636, opened 2022-09-20
remove consul server name in async QC json

noferini:
I cleaned up the consul server name in few qc json for async (others were already without it).
I noted that in local analysis (consul server not reachable) it prevents job to close at the end.
@knopers8 can you confirm it is correct what I did? 
If it is we should check also MC.
Cheers,
Francesco

### DIFF
--- a/DATA/production/qc-async/itstpctof.json
+++ b/DATA/production/qc-async/itstpctof.json
@@ -16,7 +16,7 @@
         "url" : "infologger:///debug?qc"
       },
       "consul" : {
-        "url" : "http://consul-test.cern.ch:8500"
+        "url" : ""
       },
       "conditionDB" : {
         "url" : "ccdb-test.cern.ch:8080"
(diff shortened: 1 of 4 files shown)


chiarazampolli:
Ciao @noferini , 
These json are merged in the final one in async processing, and consul is taken from https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/qc-async/qc-global.json, where you see that it is empty.
Maybe you ran with one such json only?
I don't see any issue with your changes, but let's wait for @knopers8 comment. 
Chiara

chiarazampolli (approved):

noferini:
Ciao @chiarazampolli ,
yes I run in my laptop with a single json to test some changes in one task.
Cheers,
Francesco

knopers8 (approved):
Indeed the consul url should have been taken from the global config normally, so I am not sure how it was breaking the tests. However, we should get rid of any reference to that VM anyway, since we want to remove it (the VM).

noferini:
Hi @knopers8 ,
then if it doesn't matter for us I would remove it since it would help to run easier quick checks. detector side, when needed.
Cheers,
Francesco

PR #637, opened 2022-09-20
Added of ZDC EPN QC

lucaquaglia15:
As discussed with Ole, we created the PR to implement the ZDC QC on EPN

!!Important: we did not yet modify the url and conditionDB in the .json file, please have a look

### DIFF
--- a/DATA/production/qc-sync/zdc.json
+++ b/DATA/production/qc-sync/zdc.json
@@ -3,7 +3,7 @@
     "config": {
       "database": {
         "implementation": "CCDB",
-        "host": "ali-qcdb.cern.ch:8083",
+        "host": "alio2-cr1-hv-qcdb1.cern.ch:8083",
         "username": "not_applicable",
         "password": "not_applicable",
         "name": "not_applicable"
--- a/DATA/production/qc-sync/zdc.json
+++ b/DATA/production/qc-sync/zdc.json
@@ -1,0 +1,77 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "ali-qcdb.cern.ch:8083",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable"
+      },
+      "Activity": {
+        "number": "42",
+        "type": "2"
+      },
+      "monitoring": {
+        "url": "influxdb-unix:///tmp/telegraf.sock"
+      },
+      "consul": {
+        "url": "http://ali-consul.cern.ch:8500"
+      },
+      "conditionDB": {
+        "url": "http://o2-ccdb.internal"
+      }
+    },
+    "tasks": {
+      "QcZDCRecTask": {
+        "active": "true",
+        "className": "o2::quality_control_modules::zdc::ZDCRecDataTask",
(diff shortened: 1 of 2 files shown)


DATA/production/qc-workflow.sh
```diff
@@ -34,6 +34,7 @@ elif [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
     [[ -z "$QC_JSON_FT0" ]] && QC_JSON_FT0=consul://o2/components/qc/ANY/any/ft0-digits-qc-epn
     [[ -z "$QC_JSON_FV0" ]] && QC_JSON_FV0=consul://o2/components/qc/ANY/any/fv0-digits-qc-epn
     [[ -z "$QC_JSON_EMC" ]] && QC_JSON_EMC=consul://o2/components/qc/ANY/any/emc-qcmn-epnall
+    [[ -z "$QC_JSON_ZDC" ]] && QC_JSON_ZDC=consul://o2/components/qc/ANY/any/zdc-rec-epn
```
martenole, line 37:
```suggestion
    [[ -z "$QC_JSON_ZDC" ]] && has_processing_step ZDC_RECO && QC_JSON_ZDC=consul://o2/components/qc/ANY/any/zdc-rec-epn
```

DATA/production/qc-workflow.sh
```diff
@@ -79,6 +80,7 @@ elif [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
     [[ -z "$QC_JSON_FT0" ]] && QC_JSON_FT0=$O2DPG_ROOT/DATA/production/qc-sync/ft0.json
     [[ -z "$QC_JSON_FV0" ]] && QC_JSON_FV0=$O2DPG_ROOT/DATA/production/qc-sync/fv0.json
     [[ -z "$QC_JSON_EMC" ]] && QC_JSON_EMC=$O2DPG_ROOT/DATA/production/qc-sync/emc.json
+    [[ -z "$QC_JSON_ZDC" ]] && QC_JSON_ZDC=$O2DPG_ROOT/DATA/production/qc-sync/zdc.json
```
martenole, line 83:
```suggestion
    [[ -z "$QC_JSON_ZDC" ]] && has_processing_step ZDC_RECO && QC_JSON_ZDC=$O2DPG_ROOT/DATA/production/qc-sync/zdc.json
```

DATA/production/qc-sync/zdc.json
```diff
@@ -0,0 +1,77 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "ali-qcdb.cern.ch:8083",
```
martenole, line 6:
```suggestion
        "host": "alio2-cr1-hv-qcdb1.cern.ch:8083",
```

martenole (commented):
Thanks! I think only the one hostname needs to be adapted, since the others are exchanged automatically. But I am still working on the test so I am not 100% sure. Then for the QC json I checked that the detector QC will always be added when ZDC is present. But since it might be that ZDC is included in the run, but the reconstruction is off, we need to add that one condition so that the ZDC QC is only enabled when the input it relies on is also available

martenole (approved):

PR #641, opened 2022-09-23
Randomizing when we store output, and possibility to run with GPUs

chiarazampolli:
@catalinristea @davidrohr

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
@@ -172,11 +172,34 @@ echo "[INFO (async_pass.sh)] envvars were set to TFDELAYSECONDS ${TFDELAYSECONDS
 if [[ ! -z "$ALIEN_JDL_SHMSIZE" ]]; then export SHMSIZE=$ALIEN_JDL_SHMSIZE; elif [[ -z "$SHMSIZE" ]]; then export SHMSIZE=$(( 16 << 30 )); fi
 if [[ ! -z "$ALIEN_JDL_DDSHMSIZE" ]]; then export DDSHMSIZE=$ALIEN_JDL_DDSHMSIZE; elif [[ -z "$DDSHMSIZE" ]]; then export DDSHMSIZE=$(( 32 << 10 )); fi
 
+# root output enabled only for 2 permille of the cases
+# keeping AO2D.root QC.root o2calib_tof.root MFTAssessment.root mchtracks.root mchclusters.root
+SETTING_ROOT_OUTPUT="ENABLE_ROOT_OUTPUT_o2_mch_reco_workflow= ENABLE_ROOT_OUTPUT_o2_mft_reco_workflow= ENABLE_ROOT_OUTPUT_o2_tof_matcher_workflow= ENABLE_ROOT_OUTPUT_o2_aod_producer_workflow= ENABLE_ROOT_OUTPUT_o2_qc= "
+
+if [[ -z $RND_PERMILLE ]]; then
+  RND_PERMILLE=$((1 + $RANDOM % 2000))
+fi
+echo "Randomly generated number for current reco = $RND_PERMILLE"
+if [[ $RND_PERMILLE -ge 2 ]]; then
+  echo "ROOT ouput will be kept ONLY FOR SOME workflows"
+else
+  echo "ROOT ouput will be kept for ALL workflows"
+  SETTING_ROOT_OUTPUT+="DISABLE_ROOT_OUTPUT=0"
+fi
+echo SETTING_ROOT_OUTPUT=$SETTING_ROOT_OUTPUT
+
+# Enabling GPUs
+if [[ -n "$ALIEN_JDL_USEGPUS" ]]; then
+  echo "Enabling GPUS"
+  export GPUTYPE="HIP"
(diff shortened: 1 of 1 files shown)


DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
```diff
@@ -169,11 +169,34 @@ fi
...
 
+# root output enabled only for 2 permille of the cases
+# keeping AO2D.root QC.root o2calib_tof.root MFTAssessment.root mchtracks.root mchclusters.root
+SETTING_ROOT_OUTPUT="ENABLE_ROOT_OUTPUT_o2_mch_reco_workflow= ENABLE_ROOT_OUTPUT_o2_mft_reco_workflow= ENABLE_ROOT_OUTPUT_o2_tof_matcher_workflow= ENABLE_ROOT_OUTPUT_o2_aod_producer_workflow= ENABLE_ROOT_OUTPUT_o2_qc= "
+
+if [[ -z $RND_PERMILLE ]]; then
+  RND_PERMILLE=$((1 + $RANDOM % 2000))
+fi
+echo "Randomly generated number for current reco = $RND_PERMILLE"
+if [[ $RND_PERMILLE -ge 2 ]]; then
```
martenole, line 180: Don't you set `RND_PERMILLE` to something between 1 and 2000 above?
```suggestion
if [[ $RND_PERMILLE -ge 1000 ]]; then
```

DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
```diff
@@ -169,11 +169,34 @@ fi
...
+if [[ -n "$ALIEN_JDL_USEGPUS" ]]; then
+  echo "Enabling GPUS"
+  export GPUTYPE="HIP"
+  export GPUMEMSIZE=$((25 << 30))
+fi
+
 # reco and matching
 # print workflow
-IS_SIMULATED_DATA=0 WORKFLOWMODE=print DISABLE_ROOT_OUTPUT="" TFDELAY=$TFDELAYSECONDS NTIMEFRAMES=-1 SHMSIZE=16000000000 DDSHMSIZE=32000 ./run-workflow-on-inputlist.sh CTF list.list > workflowconfig.log
+env $SETTING_ROOT_OUTPUT IS_SIMULATED_DATA=0 WORKFLOWMODE=print TFDELAY=$TFDELAYSECONDS NTIMEFRAMES=-1 SHMSIZE=16000000000 DDSHMSIZE=32000 ./run-workflow-on-inputlist.sh CTF list.list > workflowconfig.log
```
martenole, line 197: Does this work to pass all env variables with prepending `env`? I have not tested it

DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
martenole: Ah nevermind, you want to store only one permille...

DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
davidrohr: env will set all the env vars that you have in front of the command.
If you use bash as shell, you could also use `eval`, which is a bash builtin instead of a separate binary like `env`.

chiarazampolli: I read that `eval` can have issues. With `env` for me it works.

chiarazampolli: But I would leave this as WIP till @catalinristea checks it for the randomization, and till we have a stable configuration for the EPNs.

chiarazampolli: For the randomization, note that for now this is a kind of placeholder, since 1) we need to check how RANDOM behaves and 2) if we have less than 1000 we might have nothing.

chiarazampolli:
Hello @catalinristea , 
I changed the logic to deicde if to store or not the files, following what is done in [/alice/cern.ch/user/a/alidaq/LHC22m/apass1/validation.sh](https://alimonitor.cern.ch/catalogue/?path=%2Falice%2Fcern.ch%2Fuser%2Fa%2Falidaq%2FLHC22m%2Fapass1%2Fvalidation.sh)
Costin confirmed that now we will always have as many masterjobs as folders in the run. 
With this logic, we might just risk to end up with a bit more than 2 permille of stored output, when the masterjob will have less than the required sampling. This is very acceptable. Let me know what you think.

catalinristea:
Ciao Chiara,
Look good to me, thanks a lot!
Yours, Catalin.

chiarazampolli:
Ok, I will do some testing, and then merge. 

Thanks a lot, @catalinristea !

chiarazampolli:
Hi @catalinristea , 
The last commit should solve the issue, and make it work also in case we process more runs per job (I added `| head -1` when we get the CTF file name). 
Can you check?
Chiara

chiarazampolli:
@catalinristea , I committed the changes we discussed, please try this out on the grid. 
Would be good to try it with 1 CTF per job, and with more than one, but only if possible. 
Cheers,
Chiara

chiarazampolli:
Latest small fixes added. Merging now.

PR #642, opened 2022-09-23
tof reco: remove --for-calib option

noferini:
All past TOF data now calibrated, --for-calib no longer needed.
To be added only in sync reco/calib


chiarazampolli (approved):

noferini:
I checked and for EPNSYNCMODE --for-calib is already enabled (O2)

PR #644, opened 2022-09-25
Don't override SHM sizes if set externally, + fix some whitespace

davidrohr:
@chiarazampolli : this allows to set the SHMSIZE to be used from extern. We need this in one or the other form. If you'd prefer to do it differently, also fine with me, then please open a PR.

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
@@ -92,7 +92,7 @@ fi
 
 echo processing run $RUNNUMBER, from period $PERIOD with $BEAMTYPE collisions and mode $MODE
 
-###if [[ $MODE == "remote" ]]; then 
+###if [[ $MODE == "remote" ]]; then
     # common archive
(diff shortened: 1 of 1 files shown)


DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
```diff
@@ -169,11 +169,14 @@ fi
 
 echo "[INFO (async_pass.sh)] envvars were set to TFDELAYSECONDS ${TFDELAYSECONDS} TIMEFRAME_RATE_LIMIT ${TIMEFRAME_RATE_LIMIT}"
 
+if [[ -z "$SHMSIZE" ]]; then export SHMSIZE=$(( 16 << 30 )); fi
```
chiarazampolli, line 172: This will not work on alien, in the sense that there, every env variable is exported as `ALIEN_JDL_XXXYYYZZZ` so here it should be:
```
if [[ -z "$ALIEN_JDL_SHMSIZE ]]; then... 
else
SHMSIZE=$ALIEN_JDL_SHMSIZE
```
Same below.

PR #646, opened 2022-09-26
margins for threshold tuning

DanielSamitz:


benedikt-voelkel (commented):
You are computing the new threshold here already which will then be passed to the macro. As far as I can see, all margins and this file are passed to the macro where then again the margin is multiplied. That seems a duplication. In fact it should be enough to do the computation here and nothing ales needs to be passed to the macro.

RelVal/o2dpg_release_validation.py
```diff
@@ -654,17 +654,19 @@ def rel_val_sim_dirs(args):
...
         for histo_name, tests in json_in.items():
             for t in tests:
-                print(t["result"])
                 if not t["comparable"]:
                     continue
-                f.write(f"{histo_name},{t['test_name']},{t['value']}\n")
+                if t["test_name"] not in REL_VAL_TEST_NAMES_MAP:
+                    continue
+                ind = REL_VAL_TEST_NAMES_MAP[t['test_name']]
+                f.write(f"{histo_name},{t['test_name']},{t['value']*threshold_margins[ind]}\n")
```
benedikt-voelkel, line 669: You are computing the new threshold here already which will then be passed to the macro. As far as I can see, all margins and this file are passed to the macro where then again the margin is multiplied. That seems a duplication. In fact it should be enough to do the computation here and nothing ales needs to be passed to the macro.

benedikt-voelkel (approved):

PR #652, opened 2022-09-27
CTF_METAFILES_DIR not required for printing workflow

martenole:
@davidrohr not 100% if its OK like this. When I would like to see the workflow which was running on the EPNs, I usually grep the topology generation command line from the ODC logs and prepend `WORKFLOWMODE=print GEN_TOPO_WORKDIR=$PWD`. But then I get this complaint `$CTF_METAFILES_DIR missing`. So I also prepend `CTF_METAFILES_DIR=foo` so that it works. But that should not be needed I think


davidrohr:
well, the `gen_topo[...].sh` is supposed to run on the EPN for topology generation, and there the `CTF_METAFILES_DIR` is certainly required. I'd prefer to check during the EPN topology generation all the required variables exist.
Running it manually is kind of a debugging test, so I think it is ok to just set `CTF_METAFILES_DIR=foo` manually, that is actually what I am doing myself as well. At least I haven't come up with a better solution yet :)

One possibility would be to restrict the check to the user epn, but that could easily be forgotten in case EPN ever changes their default user.

martenole:
Hm, yes, restricting the check based on the username is also not a good idea I think. Ok then I'll leave it as it is for now and might reopen a PR if I have a better idea

PR #653, opened 2022-09-28
pid TO QC: add other track sources

noferini:
Adding other track sources for TOF PID QC
- DATA: TPC tracks
- MC: TPC and ITS-TPC-TRD tracks


noferini:
@ercolessi 
WIP since it has to be tested (asap)

noferini:
Review open
all jsons tested

For MC I adjusted also the track-type list of o2-global-track-cluster-reader in
MC/bin/o2dpg_sim_workflow.py

This should be enough

noferini:
Also FST was passed in my local test.
merging PR

chiarazampolli (approved):

PR #654, opened 2022-09-29
MCH: allow usage of MULTIPLICITY_FACTOR_RAWDECODERS in pedestal decoder

aphecetche:


aphecetche:
@shahor02 @davidrohr @chiarazampolli @martenole 
Could this one be reviewed/merged ? Thanks.

shahor02 (approved):
Hi @aphecetche 
It is ok but note that default MULTIPLICITY_FACTOR_RAWDECODERS in the setenv.sh is 1.

aphecetche:
@shahor02 yes, but that script is (only) used in the pedestal MCH runs for which we (@aferrero2707 ) "control" the ECS configuration and can add the relevant `MULTIPLICITY_FACTOR_RAWDECODERS` value there.
Thanks.

PR #658, opened 2022-10-06
select proper json depending on sources TOFPID

noferini:
@sawenzel , @knopers8 
This is to fix the problem in https://alice.its.cern.ch/jira/browse/QC-872
Not yet tested.


noferini:
@sawenzel ,
is it possible to run a test with it in alibi before to merge it?
Otherwise I will try to set up a test in my laptop

sawenzel:
I'll test it and report back here.

sawenzel (approved):

sawenzel:
works on alibi with the anchored test

PR #662, opened 2022-10-07
allow list of Summary.json files for calculation of thresholds

DanielSamitz:


benedikt-voelkel (commented):
@DanielSamitz Thanks for the suggestions!
In the current implementation we might (worst case) open/close N_histos * N_tests * N_thresh_files files.
Could we maybe open the files once and put the calculated values in a dict so we have for instance
```python
thresh = {"<histo_name>_<test_name>": <mean-or-max-of-this>, ...}
```
Keeping a threshold dict during runtime will not be a mem problem

PR #667, opened 2022-10-10
Disable QC

benedikt-voelkel:
Du to current hangs of QC, independent of SW version


benedikt-voelkel:
@sawenzel QC disabled here as well

chiarazampolli:
Would it not be enough to run without "--include-local-qc"?

benedikt-voelkel:
Yeah, in principle yes. But there is also the entire `mattermost` feedback which I think we should take out as well. Just so that there is no confusion.
To make it clear, this is why I took out everything around running QC here

PR #672, opened 2022-10-12
Possibility to add more output from JDL, or define what to keep from JDL

chiarazampolli:
Hello @catalinristea , 
This would be useful e.g. for [O2-3268](https://alice.its.cern.ch/jira/browse/O2-3268). What do you think?


catalinristea:
Ciao Chiara,

Seems fine, but I would change the delimiter to either " " or "," which are the string types we already use inside JDL - if possible. I'm not sure if ML will pass correctly ";" to the environment (regardless enclosing into inverted commas), as it could be reserved for EOL. Better stay out of this sort of trouble :)

Thanks!

Yours,
Catalin.

chiarazampolli:
@catalinristea : done! I use now "," like for the packages. 
Note that yesterday I forgot to commit another piece, where we can determine fully from the JDL what to keep. E.g. for [O2-3268](https://alice.its.cern.ch/jira/browse/O2-3268), this could be useful, as we will need only the output for the primVtx and tfidinfo. 

If you have no more comments, I will merge. 

Thanks a lot for the feedback!

PR #675, opened 2022-10-13
Fix typo for QC log file writing

martenole:
@davidrohr  
Test failed 2b6zVyP7Fr1 :(


martenole:
I cannot merge myself without approval. Maybe @shahor02 would you be so kind?

shahor02 (approved):

PR #676, opened 2022-10-13
fixed mother/daughter track IDs when joining  lists of particles

DanielSamitz:
The mother/daughter track IDs get messed up when the list of particles from different generators in the vector of generators of the cocktail are joined at importParticles(). Should be shifted according to the position in new (full) list of particles. Please check if this makes sense.

### DIFF
--- a/MC/config/PWGDQ/external/generator/GeneratorCocktail.C
+++ b/MC/config/PWGDQ/external/generator/GeneratorCocktail.C
@@ -33,15 +33,16 @@ class GeneratorCocktail_class : public Generator
       int nPart = mParticles.size();
       g->importParticles();
       for (auto p : g->getParticles()) {
-        if (p.GetFirstMother() > -1)
-          p.SetFirstMother(p.GetFirstMother() + nPart);
-        if (p.GetSecondMother() > -1)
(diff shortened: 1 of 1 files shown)


MC/config/PWGDQ/external/generator/GeneratorCocktail.C
```diff
@@ -30,9 +30,19 @@ class GeneratorCocktail_class : public Generator
   bool importParticles() override
   {
     for (auto& g : *mEntries) {
+      int nPart = mParticles.size();
       g->importParticles();
-      for (auto& p : g->getParticles())
+      for (auto p : g->getParticles()) {
```
benedikt-voelkel, line 35: You could do
```c++
for (auto& p : g->getParticles()) {
  mParticles.push_back(p);
  auto& pEdit = mParticles.back();
  // do what's needed with pEdit
}
That should be one copy less

DanielSamitz: ok, I implemented it in the way you suggested.

DanielSamitz:
Included also a getter function for the vector of generators in the cocktail

PR #686, opened 2022-10-20
TIME_FRAME_LIMIT in case of GPUs depends on whether we keep the root …

chiarazampolli:
…output
@davidrohr


chiarazampolli:
Discussed with @davidrohr , merging.

PR #689, opened 2022-10-20
FT0: time offset calibration workflow

afurs:


chiarazampolli:
Hello @afurs , 
This is for standalone runs, right?

PR #692, opened 2022-10-21
Align anchor better with alibi framework

benedikt-voelkel:
* make sure analyses are run according to alibi framework when run as part of it


benedikt-voelkel:
Closing, replace by https://github.com/AliceO2Group/O2DPG/pull/700

PR #694, opened 2022-10-21
Switch to full TPC PID

benedikt-voelkel:


benedikt-voelkel:
Ciao @chiarazampolli 
Can we merge this? Potentially before 18:00 if possible :)

gconesab (approved):

PR #703, opened 2022-10-26
Fixes for qc-workflow.sh

davidrohr:
@chiarazampolli @knopers8 : QC tasks should not be added solely by the presense of the detector, but only if requested in the list of QC tasks. Note that the default list of QC tasks is the list available detectors + $LIST_OF_GLORECO.

### DIFF
--- a/DATA/production/qc-workflow.sh
+++ b/DATA/production/qc-workflow.sh
@@ -158,10 +158,10 @@ elif [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
 
   # GLO QC: VTX and ITSTPC_MTCH
   if has_detectors_reco ITS && has_detector_matching PRIMVTX; then
-    add_QC_JSON GLO $QC_JSON_GLO_VTX
+    add_QC_JSON GLO_PRIMVTX $QC_JSON_GLO_VTX
   fi
   if has_detectors_reco ITS TPC && has_detector_matching ITSTPC; then
-    add_QC_JSON GLO $QC_JSON_GLO_ITSTPC_MTCH
+    add_QC_JSON GLO_ITSTPC $QC_JSON_GLO_ITSTPC_MTCH
   fi
 
   # PID QC
--- a/DATA/production/qc-workflow.sh
+++ b/DATA/production/qc-workflow.sh
@@ -149,14 +149,14 @@ elif [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
     add_QC_JSON matchTOF ${QC_JSON_TOF_MATCH}
   fi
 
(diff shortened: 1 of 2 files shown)


DATA/production/qc-workflow.sh
```diff
@@ -87,8 +87,8 @@ elif [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
     [[ -z "$QC_JSON_CPV" ]] && QC_JSON_CPV=$O2DPG_ROOT/DATA/production/qc-sync/cpv.json
     [[ -z "$QC_JSON_PHS" ]] && QC_JSON_PHS=$O2DPG_ROOT/DATA/production/qc-sync/phs.json
     [[ -z "$QC_JSON_TRD" ]] && QC_JSON_TRD=$O2DPG_ROOT/DATA/production/qc-sync/trd.json
-    [[ -z "$QC_JSON_GLO_VTX" ]] && QC_JSON_GLO=$O2DPG_ROOT/DATA/production/qc-sync/glo-vtx-qcmn-epn.json
-    [[ -z "$QC_JSON_GLO_ITSTPC_MTCH" ]] && QC_JSON_GLO=$O2DPG_ROOT/DATA/production/qc-sync/glo-itstpc-mtch-qcmn-epn.json
+    [[ -z "$QC_JSON_GLO_PRIMVTX" ]] && QC_JSON_GLO=$O2DPG_ROOT/DATA/production/qc-sync/glo-vtx-qcmn-epn.json
```
chiarazampolli, line 90: Here there seems to be something wrong... we are redefining `QC_JSON_GLO` with the line below. Am I wrong?

DATA/production/qc-workflow.sh
```diff
@@ -87,8 +87,8 @@ elif [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
...
+    [[ -z "$QC_JSON_GLO_ITSTPC_MTCH" ]] && QC_JSON_GLO=$O2DPG_ROOT/DATA/production/qc-sync/glo-itstpc-mtch-qcmn-epn
```
chiarazampolli, line 91: Why did you remove the `.json`?

davidrohr: aehm, no idea how that happened... will fix it

DATA/production/qc-workflow.sh
davidrohr: indeed, that is wrong. But was wrong before, so it was never active ... :(

chiarazampolli: Yes, sure, it was wrong also before, I pointed it out since you were fixing the script (thanks!).

PR #708, opened 2022-10-27
TPC: add data sampling to Cluster task

tklemenz:
The sampling fraction can be set in line 105. For now, I put 10% that goes through.

### DIFF
--- a/DATA/production/qc-async/tpc.json
+++ b/DATA/production/qc-async/tpc.json
@@ -36,8 +36,8 @@
         "resetAfterCycles": "1",
         "maxNumberCycles": "-1",
         "dataSource": {
-          "type": "direct",
-          "query": "inputClus:TPC/CLUSTERNATIVE"
+          "type": "dataSamplingPolicy",
+          "name": "tpc-clusters"
         },
         "taskParameters": {
           "mergeableOutput": "true",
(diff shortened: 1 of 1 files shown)


DATA/production/qc-async/tpc.json
```diff
@@ -93,5 +93,20 @@
     }
   },
   "dataSamplingPolicies": [
+    {
+      "id": "tpc-clusters",
+      "active": "true",
+      "machines": ["localhost"],
+      "query" : "inputClusters:TPC/CLUSTERNATIVE",
+      "outputs": "sampled-clusters:DS/CLUSTERNATIVE",
```
chiarazampolli, line 101: Hi @tklemenz , 
Is this `outputs` needed? I have never seen it, maybe it is optional? @knopers8 .

tklemenz: Hi Chiara, 
I tested it locally and it did not work without it. That was actually the option I was missing. The `clusterHandler` taking care of the cluster input is looking for the origin `DS` to not mix it up with non-sampled data. Afaik the `outputs` is just re-labeling the messages in the DPL. But I guess Piotr has more insight.

chiarazampolli: Ok, probably it is because you look for this specific origin. I will merge now, if @knopers8 can comment further, I could learn :-)

chiarazampolli (approved):

PR #710, opened 2022-10-28
regex for --include-patterns in inspect command

DanielSamitz:
I added the include_patterns argument and the regex comparison also in `plot_pie_charts` and `plot_values_threshold`, only then noticed that including it in `make_single_summary` already does the full job (at least for "inspect"). For the moment I left it also in the other function, but with default value None. Should I take it out again?

### DIFF
--- a/RelVal/o2dpg_release_validation.py
+++ b/RelVal/o2dpg_release_validation.py
@@ -235,6 +235,19 @@ def file_sizes(dirs, threshold):
     return collect_dict
 
 
+def check_include_patterns(histo_name, include_patterns):
+    if not include_patterns:
+        return True
+    else:
+        include_this = False
+        for ip in include_patterns:
+            if re.search(ip,histo_name):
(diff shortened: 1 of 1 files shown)


RelVal/o2dpg_release_validation.py
```diff
@@ -235,14 +235,23 @@ def file_sizes(dirs, threshold):
...
     for histo_name, tests in summary.items():
+        # check if histo_name is in include patterns
+        if include_patterns:
+            include_this = False
+            for ip in include_patterns:
+                if re.search(ip,histo_name):
+                    include_this = True
+                    break
+            if not include_this:
+                continue
```
benedikt-voelkel, line 254: This appears now in at least 4 places. Could you put that into a small helper function?

benedikt-voelkel (changes requested):

benedikt-voelkel (approved):

PR #714, opened 2022-11-01
Add CTP in production

catalinristea:
Hi @noferini, @chiarazampolli

Here is the commit to add  CTP to the list of detectors for async reco.
Please review and merge if you find it appropriate.

Mail from Evgeny,
_I am trying to understand what is going on with ctp info in Run3 AO2Ds. According to Roman, ctp readout is available for all runs taken after 24/09/22 16:24 and the ctp reco workflow was ready before that. AO2D producer seems to support CTP info as well. However, I still see zeros in the trigger mask stored in AO2Ds that is supposed to be filled from CTP digits. It seems the reason is that o2-ctp-reco-workflow is not yet included in the reconstruction chain._

Follow-up from Francesco,
_I checked in the dpl-workflow.sh and it seems that the option is already there.
I guess it should be enough to add CTP to the detector list._

Thanks!

Yours, Catalin.

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
@@ -9,7 +9,7 @@ export SETENV_NO_ULIMIT=1
 export DPL_DEFAULT_PIPELINE_LENGTH=16
 
 # detector list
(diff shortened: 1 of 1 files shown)


noferini:
Hi @catalinristea , thanks a lot
Francesco

noferini (approved):

lietava:
@shahor02 : this includes also luni workflow ? 
i.e.: https://github.com/AliceO2Group/AliceO2/pull/10008#issuecomment-1272557219 - this one ?

PR #717, opened 2022-11-06
Add CTP lumi information to aggregator

martenole:
To be merged only after https://github.com/AliceO2Group/AliceO2/pull/10224 is merged.


martenole:
Hi @davidrohr this could in principle be merged now. The only caveat is that at P2 we should not have this commit in O2DPG until O2 is updated (otherwise `--enable-ctp` option does not exist for the workflow). Alternatively one would need to set `CALIB_TPC_SCDCALIB=0` in order to disable the calibration which would be fine in any case until O2 is updated. So should we merge this now?

PR #718, opened 2022-11-07
Reenable QC logging

davidrohr:


DATA/production/qc-workflow.sh
```diff
@@ -211,6 +211,8 @@ elif [[ -z $QC_JSON_FROM_OUTSIDE ]]; then
   rm -Rf $FETCHTMPDIR
 fi
 
+[[ $EPNSYNCMODE == 1 && $NUMAGPUIDS == 1 ]] && QC_CONFIG+=" --override-values \"qc.config.infologger.filterDiscardFile=../../qc-_ID_${NUMAID}.log\""
```
martenole, line 214: Hi @davidrohr will you change this to the syntax mentioned by Barth in https://alice.its.cern.ch/jira/browse/QC-882?
So a list of key-values to be overridden and then apply the `override-values` option only once?

davidrohr: Will do that later in a separate PR. The 2 options we have right now can never collide

PR #724, opened 2022-11-10
TimeMachine feature in O2DPG MC workflows

sawenzel:
New option to set the --condition-not-after T time to use a version of CCDB which was valid at time T.

See
https://alice.its.cern.ch/jira/browse/O2-3317

This commit applies to all CCDB queries using DPL fetcher. We still need to constrain queries coming through BasicCCDBManager.


sawenzel:
Needs to be merged after https://github.com/AliceO2Group/AliceO2/pull/10248

PR #725, opened 2022-11-10
Make orbits configurable and default to 128

sawenzel:


chiarazampolli:
Ciao @sawenzel , it seems there are merging conflicts, can you check?

PR #727, opened 2022-11-12
Send CTP digits required by RecoContainer logic

martenole:
I tested only with local aggregator and there it was fine. Now with the proxy the CTP/DIGITS input is missing which is required when using the RecoContainer logic for loading the CTP input:
https://github.com/AliceO2Group/AliceO2/blob/7a344162ae2735fa5f436bb48248b588e3e861b5/DataFormats/Detectors/GlobalTracking/src/RecoContainer.cxx#L403-L412


martenole:
Tests passed and this should be available tomorrow. So merging

PR #729, opened 2022-11-13
[PWGHF] Use embedding with replacement in MC simulations forr FH trigges 

fgrosa:
This PR implements the possibility to trigger on c-cbar or b-bbar events once each N background events using the functionality implemented in https://github.com/AliceO2Group/AliceO2/pull/10227. @sawenzel could you have a look if I implemented it properly? Thanks!


sawenzel:
Looks ok to me.

PR #732, opened 2022-11-15
merging TTrees form AO2D files

DanielSamitz:
- removed `--print` option from `inspect` command
- removed empty plot test_values_thresholds_summary.png
- implemented merging of trees from different DF_\<number>  subdirectories in AO2D files at the histogram level (instead of first merging the trees and then writing it to a histogram)
                  -  advantage: Merging many trees from many DF directories can become very large. Writing each tree immediately to a histogram and then adding only the histograms avoids this problem.
                  - disadvantage: The first DF directory sets the binning and the histogram range for all following trees. Most physical distributions should be very similar in the different DFs, so it should not be a problem. But e.g. for IndexTables this can for sure cut away some parts of later trees, but I think it is anyhow questionable how much sense it makes to merge IndexTables from different DFs, write them to a histogram and then do a shape comparison of that histogram with another one from another AO2D.

### DIFF
--- a/RelVal/o2dpg_release_validation.py
+++ b/RelVal/o2dpg_release_validation.py
@@ -336,7 +336,8 @@ def plot_values_thresholds(summary, out_dir, title, include_patterns=None, exclu
     test_histo_value_map = extract_from_summary(summary, ["value", "threshold"], include_patterns, exclude_patterns)
 
     for which_test, histos_values_thresolds in test_histo_value_map.items():
-
+        if which_test == REL_VAL_TEST_SUMMARY_NAME:
+            continue
         figure, ax = plt.subplots(figsize=(20, 20))
         ax.plot(range(len(histos_values_thresolds["histograms"])), histos_values_thresolds["value"], label="values", marker="x")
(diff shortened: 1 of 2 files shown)


RelVal/o2dpg_release_validation.py
```diff
@@ -336,7 +336,8 @@ def plot_values_thresholds(summary, out_dir, title, include_patterns=None, exclu
     test_histo_value_map = extract_from_summary(summary, ["value", "threshold"], include_patterns, exclude_patterns)
 
     for which_test, histos_values_thresolds in test_histo_value_map.items():
-
+        if which_test == "summary":
```
benedikt-voelkel, line 339: can we change it to compare to the global variable?
https://github.com/AliceO2Group/O2DPG/blob/master/RelVal/o2dpg_release_validation.py#L120

Same below

benedikt-voelkel (changes requested):

RelVal/o2dpg_release_validation.py
DanielSamitz: OK, done

PR #733, opened 2022-11-15
fix completion policies of calib output proxies (O2DPG part)

davidrohr:


davidrohr:
needs https://github.com/AliceO2Group/AliceO2/pull/10281

PR #738, opened 2022-11-17
Option to use pregenerated collision context in simulation

sawenzel:
This introduces the possibility (using argument --pregenCollContext) to generate the collision/timeframe structure **before** the signal transport simulation, instead of the other way around.

This makes sure that, in particular for low interaction rates, a dynamic number of events per timeframe can be simulated.

I assume this mode will become the standard at some point.

Note that the scheme does not yet work with embedding.

Also applying a new limit to constrain TPC digits to within the timeframe limits (to prevent crashes in TPC clusterization).


sawenzel:
Needs https://github.com/AliceO2Group/AliceO2/pull/10304 + https://github.com/AliceO2Group/AliceO2/pull/10306 before it can be merged.

PR #740, opened 2022-11-21
Accompany https connection by the identity and the certification auth…

chiarazampolli:
…orities

@costing
@catalinristea , @noferini : for 22f apass2


chiarazampolli:
This does not work for a user environment.

chiarazampolli:
Fixed, merging, thanks @costing for the help!

PR #743, opened 2022-11-21
Retrieving IR and fix for pass name

chiarazampolli:
@catalinristea , @noferini , @shahor02 : this is needed for pp low IR, were we need to add `pvertexer.timeMarginVertexTime=1.3`. 
I use as threshold 100 kHz, what do you think?

### DIFF
--- a/DATA/production/common/getIR.C
+++ b/DATA/production/common/getIR.C
@@ -1,3 +1,5 @@
+#include <fstream>
+#include<stdio.h>
 #if !defined(__CLING__) || defined(__ROOTCLING__)
 #include "CCDB/BasicCCDBManager.h"
 #include "CommonDataFormat/InteractionRecord.h"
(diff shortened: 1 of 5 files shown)


shahor02:
Hi Chiara

Not sure even for the same system we can have a common threshold for all parameters, we may need to consider a few thresholds. Moreover, 100:1 pp:pbpb conversion is ok only for the plain occupancy related parameters, e.g. for the vertexing ones this will not work.

chiarazampolli:
> Hi Chiara
> 
> Not sure even for the same system we can have a common threshold for all parameters, we may need to consider a few thresholds. Moreover, 100:1 pp:pbpb conversion is ok only for the plain occupancy related parameters, e.g. for the vertexing ones this will not work.

Hello @shahor02 , 
For now, this affects only the PrimVtx. I can then change the macro to return the IR, and we can then modify the script if the need comes. 
For PbPb, since for now we have only low IR, I can not consider the IR. 
I will ping you when I do the changes, so that you can check.
Thanks for the feedback!

chiarazampolli:
Hello @catalinristea , @shahor02 , 
I changed the macro that computes the IR, for now the return value is just printed in the logs. What do you think? Then if we will need to use the IR to define settings, we can do so. 
Note that the settings that were meant to be defined by the IR are now in an `if` in setenv_extra, which is why I don't set them here anymore (namely the TimeMargin).
Chiara

chiarazampolli:
Nothing fundamental in this PR now, but important fixes for the pass name, so I am merging.

PR #744, opened 2022-11-21
Updating settings for LHC22o,p,q,r,s,t

chiarazampolli:
@catalinristea , @noferini : as always this might be needed for QC, if tasks use the files from local directories.
The selection in bookkeeping to get the run number was just "first good physics run".


catalinristea:
Ciao Chiara,

This comes right in time for starting the apass1 of 22p.
Thanks!

Yours, Catalin.

PR #745, opened 2022-11-21
Fixes / more flexibility for async reco script

davidrohr:
@chiarazampolli : Could you check if these changes are OK?

And we might want to apply them to all async scripts.
That said, it is kind of dangerous that we have so much copy&paste.


chiarazampolli (approved):

chiarazampolli:
Ciao @davidrohr , 
With the copy&paste you mean the different folders for MayJune, apass1/2 etc?
Chiara

davidrohr:
> Ciao @davidrohr , With the copy&paste you mean the different folders for MayJune, apass1/2 etc? Chiara

yes, I would imagine that for the future it will only become more folders

chiarazampolli:
AliDPG was not like this. We tried this approach when things were changing so much from 2021 to 2022 pilot, then 2022 13.6. Plus we have special productions (for TPC calib). But we kept using the scripts from 22f in the end, which is the best. For PbPb we are discussing if we want an extra folder.
Next year we might structure the package differently, with one folder only.

PR #747, opened 2022-11-22
Hotfix for empty QC and CALIB during topo generation

martenole:
@davidrohr currently when starting a run from ECS there is no QC running on the EPN and also there are no calibration workflows. I don't see how this would work at all. If I understand correctly ECS is not sending anymore a list of WORKFLOW_DETECTORS_QC and WORKFLOW_DETECTORS_CALIB. Only an exclude list.
With this hack it would before the topology generation is started fill both lists with the WORKFLOW_DETECTORS.
But maybe I have to create new RPMs with this?
Another workaround would be to always put WORKFLOW_DETECTORS_QC and WORKFLOW_DETECTORS_CALIB in the extra env vars with all detectors which are included. But obviously this is quite error prone...


martenole:
Now with a proper fix. Problem was that by replacing TRG with CTP regardless whether the list was empty or not the list was set to an empty string, now that AliECS is not setting them anymore

martenole:
Works on the EPN, merging

PR #749, opened 2022-11-23
Replace CTF_METAFILES_DIR by EPN2EOS_METAFILES_DIR

martenole:
We will write meta files for CTF data as well as for calibration data. All meta files go into the same directory. Reflect this by naming the directory in a generic way.
Currently this exports still CTF_METAFILES_DIR. Once that is replaced in O2 as well this can be removed in a following PR.


davidrohr:
Looks correct

martenole:
Is the CI starting automatically or does it need an approval first?

davidrohr:
Starts automatically, is just not shown for O2DPG.

PR #753, opened 2022-11-24
ITS track and cluster sampling fraction from 30 to 100%

iravasen:
cc @chiarazampolli

### DIFF
--- a/DATA/production/qc-async/its.json
+++ b/DATA/production/qc-async/its.json
@@ -61,7 +61,7 @@
           "vertexXYsize": "0.5",
           "vertexZsize": "15",
           "vertexRsize": "0.8",
-          "NtracksMAX"  : "100",
+          "NtracksMAX"  : "5000",
           "doTTree": "0",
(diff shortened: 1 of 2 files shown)


chiarazampolli:
Hi @iravasen , 
This is not a change in the fraction of accepted tracks, but on the max number of tracks, right?
Chiara

iravasen:
Hi @chiarazampolli, 

- the change of the parameter `fraction` in the json means that 100% of the TFs will be processed by QC 
- the change to `NtracksMAX` produces the change of the x axis of the histogram containing the number of tracks per ROF. This now goes up to 5000 (maybe too high but there is no problem).

chiarazampolli:
Ciao @iravasen , 

Thanks, I was clicking the wrong link from the mail, it brought me to one of the two commits only that is why I was confused. 
Approving and merging.

Chiara

chiarazampolli (approved):

iravasen:
no problem, thanks!

PR #754, opened 2022-11-24
print subcommand and Kolmogorov

DanielSamitz:
- new `print` subcommand
    - filters with `--include-patterns`, `--exclude-patterns`, `--flags` and `--flags-summary`
    - for the moment prints only names of histograms to screen. More information could be added
- new `--flags-summary` filter also for `inspect` subcommand
    - now uses all filters also for plotting test_values_thresholds_{TESTNAME}.png  (missing before) and pie charts (probably anyhow not useful when using filter on flags)
- switched to ROOT's Kolmogorov test instead of bin_cont test
    - for the moment uses the probability that ROOT returns as a metric. Could also use directly the Kolmogorov distance?
    - still have to find reasonable value for default threshold for the Kolmogorov test

### DIFF
--- a/RelVal/o2dpg_release_validation.py
+++ b/RelVal/o2dpg_release_validation.py
@@ -224,7 +224,7 @@ def file_sizes(dirs, threshold):
     return collect_dict
 
 
-def load_patterns(include_patterns, exclude_patterns):
+def load_patterns(include_patterns, exclude_patterns, print_loaded=True):
     """
     Load include patterns to be used for regex comparion
     """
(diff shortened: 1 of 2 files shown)


RelVal/o2dpg_release_validation.py
```diff
@@ -964,6 +988,27 @@ def dir_comp(args):
         json.dump(file_sizes_to_json, f, indent=2)
     return 0
 
+def printTable(args):
```
benedikt-voelkel, line 991: this should be called `print_table` to be consistent with function naming

benedikt-voelkel (changes requested):

benedikt-voelkel (approved):

PR #755, opened 2022-11-24
PWGEM: add GeneratorEMCocktailV2.C for LMee light flavour cocktail simulation

DanielSamitz:
This new external generator has basically the same functionalities as Run2 AliPhysics' [AliGenEMCocktailV2](https://github.com/alisw/AliPhysics/blob/master/PWG/Cocktail/AliGenEMCocktailV2.cxx) for the LMee LF cocktail simulations. It is derived from the `GeneratorCocktail_class` in [GeneratorCocktail.C](https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/PWGDQ/external/generator/GeneratorCocktail.C) and adds generators for different sources to the cocktail, that are generated by [GeneratorParam](https://github.com/AliceO2Group/AEGIS/blob/master/GeneratorParam/GeneratorParam.cxx) from kinematic distributions loaded by [GeneratorParamEMlibV2](https://github.com/AliceO2Group/AEGIS/blob/master/GeneratorParam/GeneratorParamEMlibV2.cxx) and are forced to decay to dileptons.
The GeneratorParamEMlibV2 and some updates to GeneratorParam are necessary for this to work. They are already merged into the AEGIS repository, but not yet included in a tag. I thought it would be good to first discuss this GeneratorEMCocktailV2.C here, to see if changes need to be made that might affect the things in AEGIS, so that I could update them there before tagging.
Any feedback is welcome.


gconesab (commented):
Hi, I am not sure you want to do this. If you need to test your parameters, we can upload the file in the grid job by hand instead of accessing alien. But you have always to put those parameters on git never on alien. If this is ccdb related parameters, then you have the ccdb. Can you explain the use case?

DanielSamitz:
> Hi, I am not sure you want to do this. If you need to test your parameters, we can upload the file in the grid job by hand instead of accessing alien. But you have always to put those parameters on git never on alien. If this is ccdb related parameters, then you have the ccdb. Can you explain the use case?

Hi @gconesab , these files are some ROOT files containing the pT distributions for the mesons produced by GeneratorParam. I'm not sure if ccdb is the right place for this. So do you suggest that we should just put them in O2DPG?

rbailhac:
Hi all, 
in Run 2 we add two possibilities: in AliPhysics (default) or in alien. In Alien made the framework more flexible to test different parametrizations of the pt spectra of pi0, eta...not always meant to be final. I have no strong opinion now for Run 3 where to put them, as long as we keep this flexibility.
Best regards,
Raphaelle

gconesab:
> Hi all, in Run 2 we add two possibilities: in AliPhysics (default) or in alien. In Alien made the framework more flexible to test different parametrizations of the pt spectra of pi0, eta...not always meant to be final. I have no strong opinion now for Run 3 where to put them, as long as we keep this flexibility. Best regards, Raphaelle

Hi all, we will discuss about it in the WP12/13 meeting in about 30 min from now, in case you can connect. In principle alien is not the right place since we cannot trace back changes and storing binary files on git is also discouraged. If somehow this could be stored as a txt file or ini file on git or stored on CCDB it would be ideal

gconesab:
Hi, are there any more developments/discussions expected?

DanielSamitz:
Hi @gconesab, from our side we are fine with this version.

PR #757, opened 2022-11-25
Remove duplicate CTF_METAFILES_DIR

martenole:
Not to be merged yet!
Only after O2 on the EPNs is updated to include https://github.com/AliceO2Group/AliceO2/pull/10347 and no more older O2 versions are used in production then this should be merged and used on the EPNs.


martenole:
Merging now, since this script is not used in the FST anyways. Will update the script on the EPNs accordingly

PR #760, opened 2022-11-28
Adding possibility to re-run EMC calib and TPC residual extraction

chiarazampolli:
This is to have the possibility to run some calib at async stage. 
I use setenv_calib.sh in conjunction with calib-workflow.sh (in O2, PR being opened) to be allowed to use all the wrokflow-setup etc functionalities. 
I took this PR to also include the switch to attach the residual extraction, which is in the apass1_TPCcalib
@davidrohr : this touches the scripts used also in sync processing, so could you please check?
Pinging also:
@mfasDa 
@ehellbar 
@wiechula 
@catalinristea

### DIFF
--- a/DATA/common/setenv_calib.sh
+++ b/DATA/common/setenv_calib.sh
@@ -28,11 +28,16 @@ if has_detector_calib ZDC && has_processing_step ZDC_RECO; then
 else
     CAN_DO_CALIB_ZDC_TDC=0;
 fi
+# for async recalibration
+if has_detector_calib EMC && has_detector_reco EMC && $EPNSYNCMODE != 1; then CAN_DO_CALIB_EMC_ASYNC_RECALIB=1; else CAN_DO_CALIB_EMC_ASYNC_RECALIB=0; fi
 
 # additional individual settings for calibration workflows
 has_detector CTP && export CALIB_TPC_SCDCALIB_CTP_INPUT="--enable-ctp"
 
 if [[ $BEAMTYPE != "cosmic" ]] || [[ $FORCECALIBRATIONS == 1 ]] ; then
+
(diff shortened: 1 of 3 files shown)


davidrohr (approved):
Only question is whether we want to check `$EPNSYNCMODE != 1` or `$SYNCMODE != 1`. In the end, we'll run the async calib in the async reco, also not during the sync reco test on the laptop, or in the FST.

chiarazampolli:
> Only question is whether we want to check `$EPNSYNCMODE != 1` or `$SYNCMODE != 1`. In the end, we'll run the async calib in the async reco, also not during the sync reco test on the laptop, or in the FST.

Yes, indeed, much better. Committed. 
Do you think we can then merge this?

davidrohr:
fine with me. Had the CI passed before you change it? If yes we can merge it.

chiarazampolli:
Which CI? Here we are in O2DPG.

davidrohr:
yes, and there is a fullCI run for each O2DPG PR.
Though it is not reported as expected for some reason when the PR is created / changed. It only appears after some time.
@TimoWilken : why is that so actually. Could we change it to the same way it works for O2?

chiarazampolli:
Hi, 
Ok, I see. Everything was green when I changed, but let's wait that Timo answers, we can also merge tomorrow.
Thanks!
Chiara

TimoWilken:
> why is that so actually. Could we change it to the same way it works for O2?

Done. The way it's done in O2 is that the check is marked as "required"; that way, GitHub knows to expect it and it shows up as "Expected" immediately when the PR is created.

This has the side effect of forbidding a merge when the check fails, instead of just showing a warning, though each PR can still be force-merged with Admin permissions.

PR #768, opened 2022-12-02
Tpc laser aggregator

rmunzer:


DATA/production/calib/tpc-laser.sh
```diff
@@ -74,12 +75,12 @@ o2-dpl-raw-proxy $ARGS_ALL \
     --disable-mc \
     --pipeline tpc-zsEncoder:20,tpc-tracker:8 \
     $GPU_CONFIG \
-    --condition-remap file:///home/wiechula/processData/inputFilesTracking/triggeredLaser/=GLO/Config/GRPECS \
+    --condition-remap "file:///home/wiechula/processData/inputFilesTracking/triggeredLaser/=GLO/Config/GRPECS;file:///home/wiechula/processData/inputFilesTracking/triggeredLaser/=GLO/Config/GRPMagField" \
```
chiarazampolli, line 78: Why do you need this remap?

wiechula: To avoid default momentum of 0.2 in case of B=0 data. We can also solve this differently, of course. But this was the quickest solution. We would need a B-Field dependent momentum cut otherwise.

chiarazampolli: Thanks for the clarification!

chiarazampolli (approved):

PR #773, opened 2022-12-05
replacing reader-delay with delay-1st-tf for tof digit reader

noferini:


chiarazampolli (approved):

chiarazampolli:
Merging, does not involve FST.

PR #774, opened 2022-12-05
[EMCAL-756] Define taskName for EMCAL async QC tasks

mfasDa:
Use taskName to prevent EMCAL QC task for being
shadowed:
- Define taskName with expected task names
- Add tag "EMCAL" to task name in key

### DIFF
--- a/DATA/production/qc-async/emc.json
+++ b/DATA/production/qc-async/emc.json
@@ -23,8 +23,9 @@
       }
     },
     "tasks": {
-      "CellTask": {
+      "CellTaskEMCAL": {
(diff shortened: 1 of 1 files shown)


mfasDa:
This is necessary as the EMCAL cluster task is shadowed by the PHOS cluster task as both had the same name. Introduction of the taskName parameter became necessary similarly to how it is used in the synchronous reconstruction where we also had shadowing of EMCAL tasks due to naming clashes notably with PHOS.

chiarazampolli (approved):

PR #777, opened 2022-12-05
Use TPC cluster time-bin shifts instead of ITS-TPC time shifts

shahor02:
Need a tag with https://github.com/AliceO2Group/AliceO2/pull/10438

@chiarazampolli with this the pbpb apass4 can restart.


chiarazampolli:
Thanks!!!

PR #783, opened 2022-12-07
Async.: Enable full output of residual aggregator

martenole:
The `export CALIB_TPC_SCDCALIB_SENDTRKDATA=1 ` is already enabled in the scripts. The full output type is needed now, since its not the default anymore.

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
@@ -60,7 +60,7 @@ fi
 if [[ $PERIOD == "LHC22s" ]]; then
   # CTP asked to extract their digits
   export ADD_EXTRA_WORKFLOW="o2-ctp-digit-writer"
-    
+
   TPCITSTIMEERR="0.3"
   TPCITSTIMEBIAS="0"
   if [[ $RUNNUMBER -eq 529397 ]]; then
@@ -92,7 +92,7 @@ if [[ $PERIOD == "LHC22s" ]]; then
 fi
 # run-dependent options
 if [[ -f "setenv_run.sh" ]]; then
-    source setenv_run.sh 
+    source setenv_run.sh
 else
     echo "************************************************************"
     echo No ad-hoc run-dependent settings for current async processing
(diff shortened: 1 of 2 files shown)


DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
```diff
@@ -233,6 +233,9 @@ if [[ $ADD_CALIB == "1" ]]; then
   fi
 fi
 
+# ad-hoc settings for TPC residual extraction
+export ARGS_EXTRA_PROCESS_o2_calibration_residual_aggregator="--output-type trackParams,unbinnedResid,binnedResid"
```
chiarazampolli, line 237: Should we put this in the `if` at line 225?

martenole: I think it does not hurt to have it here. In case the `o2-calibration-residual-aggregator` is not included in the global workflow, then this variable is simply ignored

chiarazampolli: I know, but I would prefer to have it next to the relevant piece of code, I find it less confusing.

martenole: Ok sure, I can move it. And is this setting needed for both LHC22f/apass1 and LHC22f/apass1_TPCcalib or do we anyhow only use one of the two now?

chiarazampolli: Now we will use only the one in 22f/apass1. 
The other script is dedicated to TPC calib, so there I did not mind to not have a check. But as I said, 22f/apass1 should from now on be the default (@catalinristea , just to be sure that he uses this one from now on).
I will merge now.

chiarazampolli (approved):

DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
martenole: Ok, thanks a lot

PR #788, opened 2022-12-11
Adding shift for ITSTPC vs FT0 also for pp runs

chiarazampolli:
@ekryshen @shahor02 @catalinristea : this is to set the TPC cluster time according to the file that Evgeny send me, can you please check?

Moreover, for @shahor02 , I think that when you added the ZDC shift for the pp run found by Evgeny, you did not use it, am I wrong?


DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
```diff
@@ -88,8 +88,23 @@ fi
 # fix also ZDC in the pp run 529038
 if [[ $RUNNUMBER -eq 529038 ]]; then
   ZDC_BC_SHIFT=486590350357 # started at orbit 136529280
+  export CONFIG_EXTRA_PROCESS_o2_ctf_reader_workflow="TriggerOffsetsParam.customOffset[11]=$ZDC_BC_SHIFT;"
```
shahor02, line 91: right, thanks for spotting

shahor02 (approved):

ekryshen:
@chiarazampolli the map looks ok but I just want to check if you are sure that there is no need to invert the sign. FYI: I provided the shifts with the same sign convention as in Pb-Pb runs.

ekryshen:
@shahor02 Also I noticed that you applied additional 1 bc shift for ZDC. It seems it is not needed because ZDC bcs are properly aligned with TVX, see hBcTVX and hBcZDC in event-selection-qa-task output:
https://evsel-qa.web.cern.ch/jsroot/?file=/data/2022/LHC22q/apass1/529038/QAResults.root

shahor02:
Hi @ekryshen 

For the sign: yes, the shifts are added, therefore they need to be taken with negative sign.
For the ZDC, ok, I thought the 1BC bias was also there, here are the values for the orbit shift only:
```
529003:  427744323072
529005: 585290686464
529006: 1007373210624
529008: 1379963464704
529009: 1454804955648
529015: 2244823206912
529037: 256033198080
529038: 486590353920
529039: 1399525890048
529043: 3079675095552
```

chiarazampolli:
> 486590353920

Hi, 
Ok, but then what should I use? run 529038 would change by 3563 BCs, so one orbit -1. 
I will merge as it is now, maybe you (@shahor02 , or @ekryshen ) can open a PR with the correct shifts for these runs? I can also translate it into a map, but for now we can just use if's. 

Thanks!

Chiara

PR #792, opened 2022-12-14
Sampling fraction set to 1 for TOF, and MFT and remove sempling when it is 100%

chiarazampolli:
@noferini , @bulukutlu , @rpezzi : I will merge before tonight, let me know in case you don't agree.

### DIFF
--- a/DATA/production/qc-async/its.json
+++ b/DATA/production/qc-async/its.json
@@ -32,8 +32,9 @@
         "maxNumberCycles": "-1",
         "dataSource_comment": "The other type of dataSource is \"direct\", see basic-no-sampling.json.",
         "dataSource": {
-          "type": "dataSamplingPolicy",
-          "name": "compclus"
+          "type": "direct",
+          "query": "compclus:ITS/COMPCLUSTERS/0;clustersrof:ITS/CLUSTERSROF/0;patterns:ITS/PATTERNS/0",
+          "query_comment" : "100% sampling"
         },
         "taskParameters": {
           "layer": "1111111",
@@ -53,8 +54,9 @@
         "maxNumberCycles": "-1",
         "dataSource_comment": "The other type of dataSource is \"direct\", see basic-no-sampling.json.",
         "dataSource": {
-          "type": "dataSamplingPolicy",
-          "name": "tracks"
+          "type": "direct",
+          "query": "Verticesrof:ITS/VERTICESROF/0;Vertices:ITS/VERTICES/0;tracks:ITS/TRACKS/0;rofs:ITS/ITSTrackROF/0;clustersrof:ITS/CLUSTERSROF/0;compclus:ITS/COMPCLUSTERS/0;patterns:ITS/PATTERNS/0;clusteridx:ITS/TRACKCLSID/0",
+          "query_comment" : "100% sampling"
         },
         "taskParameters": {
           "runNumberPath": "",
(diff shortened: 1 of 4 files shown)


rpezzi (approved):
Thanks. also pinging @tomas-herman

bulukutlu:
Hi, the fraction for TPC clusters was reduced to 10% after observations that the task was taking too much time. I am not sure if this problem has been elevated. @davidrohr , @tklemenz could you confirm if it would be okay to raise the sampling fraction?

Cheers.
Berkin

chiarazampolli:
Hello @bulukutlu , 
You mean for *async*, right? We can still keep it low, sure. Just let me kwow.
Chiara

bulukutlu:
> Hello @bulukutlu , You mean for _async_, right? We can still keep it low, sure. Just let me kwow. Chiara

Yes indeed for the async cluster task. I think it is fine if it stays at 10%, @wiechula correct me if you disagree of course.

knopers8:
If you need 100%, I recommend connecting to a data source directly, so there is no copy done in the Dispatcher. See the documention: https://github.com/AliceO2Group/QualityControl/blob/master/doc/ModulesDevelopment.md#bypassing-the-data-sampling

bulukutlu:
Hello @knopers8 , 
we were indeed initially using the `direct` option (https://github.com/AliceO2Group/O2DPG/commit/5957ee39c48decb52d5ba4a4571a9011f33daeca) but it was observed to be still too slow in this case so the sampling was introduced for the clusters.
Berkin

chiarazampolli:
Ciao @knopers8 , 
Ok, sorry, I did not think about the copy and simply tried the simplest change. I will fix it and leave TPC to the original value.
Cheers,
Chiara

chiarazampolli:
Hello @knopers8 , 
Can you check now?
Pinging also @iravasen , @dstocco to check for MID and ITS.
Chiara

dstocco:
This is probably a general issue. Is there any plan to switch to DIRECT under the hood as soon as the detector requires a data fraction of 1, instead of asking each detector to do it?

dstocco:
Anyways, in the meanwhile, please check here: https://github.com/AliceO2Group/O2DPG/pull/798

chiarazampolli:
Hi @dstocco ,
i let @knopers8 comment on your question.
Since your PR is contained here, i would just use this one if you agree and close the other.

cheers,

chiara

dstocco:
Ciao @chiarazampolli ,
sorry I did not realise that the modification for MID was already performed here.
Sure, you can close the other one.
Cheers,
Diego

knopers8:
> This is probably a general issue. Is there any plan to switch to DIRECT under the hood as soon as the detector requires a data fraction of 1, instead of asking each detector to do it?

We could discuss it, but this way we would forbid (or at least make difficult) to fulfill other requirements that were given to QC and eventually we want to achieve:
- allow to reconfigure sampling rate during data taking (moving back or from 100% would change the process topology, which we cannot do)
- allow for isolation of the main data flow and QC (setting 100% through Dispatcher might be still wanted in case we want to avoid blocking the main data flow).

knopers8 (approved):
yes, this is what i meant, thanks!

DATA/production/qc-async/its.json
```diff
@@ -53,8 +54,9 @@
         "maxNumberCycles": "-1",
         "dataSource_comment": "The other type of dataSource is \"direct\", see basic-no-sampling.json.",
         "dataSource": {
-          "type": "dataSamplingPolicy",
-          "name": "tracks"
+          "type": "direct",
+          "query": "Verticesrof:ITS/VERTICESROF/0;Vertices:ITS/VERTICES/0;tracks:ITS/TRACKS/0;rofs:ITS/ITSTrackROF/0;clustersrof:ITS/CLUSTERSROF/0;compclus:ITS/COMPCLUSTERS/0;patterns:ITS/PATTERNS/0;clusteridx:ITS/TRACKCLSID/0",
+          "query_comment" : "100% sampling"
```
knopers8, line 59: FYI, the comment is of course optional

chiarazampolli: Yes, I added a "dummy" one :-)

chiarazampolli:
@noferini , do yo agree with the change?

iravasen:
Hi @knopers8 , all, thanks! For ITS was already at 100% and I agree on the removal of the dataSamplingPolicy. Thank you.

noferini:
> @noferini , do yo agree with the change?

yes, sure

chiarazampolli:
Thanks to everyone for the feedback! Merging

PR #794, opened 2022-12-14
Update MFTTracking param on setenv_extra.sh

rpezzi:
- Reduce number of cluster bins used for track-finding optimization. This increases MFT acceptance with a slight increase in reconstruction CPU cost.
- Remove obsolete configurations (which are defaults and should not change any more)

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
@@ -228,7 +228,7 @@ export ARGS_EXTRA_PROCESS_o2_fv0_reco_workflow="--fv0-reconstructor"
 #...
 
 # ad-hoc settings for MFT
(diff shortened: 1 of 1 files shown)


rpezzi:
I checked this configuration for the PbPb test beam. There is a 20% increase in efficiency and 70% increase in tracking time. Therefore, for the time being I would like to keep this setting for both pp and PbPb.

chiarazampolli (commented):
Hello @rpezzi , the increase by 70% is not too bad considering that privately you wrote me `it goes from 0.3 to 0.5 seconds per TF on my laptop`. But I would put a protection to enable this in pp or PbPb 22s only, just to not forget that it needs optimiazation for PbPb at nominal IR. What do you think?

rpezzi:
Hi @chiarazampolli 

> I would put a protection to enable this in pp or PbPb 22s only

Sure. This is done. There is also one new parameter that improves chi^2 of MFT tracks, without any impact on CPU budget.

chiarazampolli (approved):

rpezzi:
To clarify, the impact of changes introduced here:
- PbPb test beam, CPU time increases by a factor of 1.7.
- pp 500 kHz, CPU time increases by a factor 7. A previous result indicating a minor impact for pp was not representative of typical minimum bias conditions.

PR #797, opened 2022-12-15
WIP: Special TRD reco config

sawenzel:
Supposed to address issue in https://alice.its.cern.ch/jira/browse/O2-2761


sawenzel:
not needed anymore. Will be solved differently

PR #798, opened 2022-12-15
Use direct if data sampling is at 1

dstocco:


dstocco:
Closing since this is already in: https://github.com/AliceO2Group/O2DPG/pull/792

PR #804, opened 2022-12-20
introduce QED background treatment

sawenzel:
* this brings QED background treatment for PbPb MC

* also brings a more consistent use of bunch filling (setup by grp-tool and queried from local CCDB thereafter)


sawenzel:
needs https://github.com/AliceO2Group/AliceO2/pull/10499/

gconesab (approved):

PR #807, opened 2022-12-22
Disable metrics if we don't want it

chiarazampolli:
@catalinristea , @shahor02


chiarazampolli:
Does not affect CI. merging.

PR #809, opened 2023-01-06
[EMCAL-851] Add postprocessing at asynchronous level for EMCAL

mfasDa:


DATA/production/qc-postproc-async/emc.json
```diff
@@ -0,0 +1,398 @@
+{
+    "qc": {
+        "config": {
+            "database": {
+                "implementation": "CCDB",
+                "host": "alio2-cr1-hv-qcdb-gpn.cern.ch:8083",
```
knopers8, line 6:
```suggestion
                "host": "ali-qcdb-gpn.cern.ch:8083",
```

DATA/production/o2dpg_qc_postproc_workflow.py
```diff
@@ -78,6 +78,7 @@ def add_QC_postprocessing(taskName, qcConfigPath, needs, runSpecific, periodSpec
 
   ## The list of QC Post-processing workflows, add the new ones below
   add_QC_postprocessing('example', 'json://${O2DPG_ROOT}/DATA/production/qc-postproc-async/example.json', needs=[], runSpecific=False, periodSpecific=False, passSpecific=True)
+  add_QC_postprocessing('EMC', 'json://${O2DPG_ROOT}/DATA/production/qc-postproc-async/emc.json', needs=[], runSpecific=False, periodSpecific=False, passSpecific=True)
```
knopers8, line 81: FYI, this means that the trends will be done for all the runs and periods which have the specific pass, e.g. `apass2`.

knopers8 (commented):
It looks OK, please use the alias for the host name and we should be good to go.

DATA/production/qc-postproc-async/emc.json
mfasDa: Leftover from testing, will be fixed.

DATA/production/o2dpg_qc_postproc_workflow.py
mfasDa: Ideally we need the trending for each period and pass separately. How do we do this?

mfasDa: OK, thanks, will modify.

chiarazampolli (approved):

mfasDa:
Second (squash-merged commit) includes both changes requested by @knopers8

PR #812, opened 2023-01-13
[QC-860] Fix the rename of MatchingTOF

knopers8:


knopers8:
@noferini @chiarazampolli can you confirm this is what it should be?

noferini:
Hi @knopers8 ,
it is fine with me.
Just a question: what is the difference between the first name (MatchingTOFAll) and the taskName (MatchTrAll)?

knopers8:
The key of the config structure (MatchinTOFAll) shall be unique, so if json files are merged, these do not overlap. The taskName is how we want the task to be named, e.g. in QCDB.

Here putting two different names is probably unnecessary, but it is mandatory for tasks like "Digits", "Clusters", which many detectors have.

noferini:
> The key of the config structure (MatchinTOFAll) shall be unique, so if json files are merged, these do not overlap. The taskName is how we want the task to be named, e.g. in QCDB.
> 
> Here putting two different names is probably unnecessary, but it is mandatory for tasks like "Digits", "Clusters", which many detectors have.

@knopers8 Thanks for the explanation!

knopers8:
@chiarazampolli could you (or someone else) approve please?

davidrohr (approved):

PR #813, opened 2023-01-14
tune LHC22f/apass1/setenv_extra for ITS async mode + recent alignment

shahor02:
Changes: 
Increase TPC-ITS matching chi2 cut to 150 account for smaller ITS errors and TPC distortion bias.
Switch to async mode of tracker-CA with 10 microns syst error at all layers. 
Tune p.vertex settings to account for smaller ITS errors.

@chiarazampolli here are the changes for async ITS mode, I am now running final validation. I've modified directly the LHC22f/setenv_extra, so the new settings will be applied to all periods. Don't know if it is fine or you prefer to save it as a different script.
Note that the TRD time problem is still not solved, once it is done, few extra changes might be needed for the vertexer.


chiarazampolli:
Hello Ruben, 
It is in principle ok to have this script changed, because it is the one used for all data, despite the name pointing to 22f. But what I see now is an issue with MC like mentioned via mail: this same script is called at MC anchored level, and then we will not be able to distinguish anymore what we used in apass1 (and other passes for other periods) and apass2. 
For PbPb, we could just do an apass5, with the new alignment, and never anchored to apass4. Same for the pp 13.6 TeV, if we never anchor to apass1, we are fine. And then we could make an if for Pilot Beam data plus LHC22f, and take the old configuration if the pass is the latest we have now. 
@noferini , @catalinristea , what do yo think? 
We can also make another script, but anyway in MC we will need to know which one to use. So the logic will have to be implemented anyway somewhere.
Chiara

noferini:
Hi @chiarazampolli ,
for pp@13 TeV and PbPb I think that it is not an issue since we are not going to anchor to the current passes (I also think it is not good to anchor PbPb apass4, too many issues there).
For pilot beam and LHC22f I agree we should put an IF.

Now it is not a big issue since we don't have an official reco pass for pp@13, but in future it could be dangerous to have such a tuning in our scripts. Is there a plan to move this kind of configurations to ccdb?
For the moment it is fine with me to go ahead in this way.
Cheers,
Francesco

chiarazampolli:
I would put the if right away for thd latest pass of 22f and pilot beam, and reprocess PbPb as soon as we are confident it is ok for pp. 

As for moving to CCDB, it could be discussed, but i hope such configurations wont be pass dependent.

chiarazampolli:
Hello, 
Let's just merge this as soon as Ruben is happy with his final validation. I will then add the "if" for the periods that were done differently on Monday. This might allow us to gain 1 day.
Chiara

shahor02:
Hi @chiarazampolli 
my 1st tests (running with O2 dev) were getting stuck in AOD writer  with messages like `[800955:internal-dpl-aod-writer]: [16:32:55][ERROR] Dropping AOD/MCCOLLISION/0 Lifetime::0 data in slot 0 with timestamp 1 < 2.`
so I had to disable the AOD producer.
Then I saw some differences between the results with the private script I was tuning and the patched setenv_extra which was supposed to produce the same results (I get more TPCITS matches and the surplus seems to be dominated by fakes, leading to more vertex splitting). Will sort this out tomorrow.

shahor02:
Hi @chiarazampolli , took longer than I thought.
Old settings are now kept for the `ALIGNLEVEL=0` while the new ones are applied for `ALIGNLEVEL=1`.
At the moment the `ALIGNLEVEL` is set to 0 for
```
[[ $BEAMTYPE == PbPb || $PERIOD == MAY || $PERIOD == JUN || $PERIOD == LHC22c || $PERIOD == LHC22d || $PERIOD == LHC22e || $PERIOD == LHC22f ]]
```
otherwise, it is set to 1. We should add pass selection for `ALIGNLEVEL` definition, but it seems the pass is not available in this script since it is overridden to apass1 to get to the script location...

catalinristea:
Hi Ruben,

Pass to the script would come from ALIEN_JDL_O2DPGPATH="LHC22f" and the more recent ALIEN_JDL_O2DPGPASSPATH="apass1" (the last one was added to keep in path the apass1 for apass2), while it should be any of ALIEN_JDL_LPMRAWPASS="2" or ALIEN_JDL_LPMPASSNAME="apass2" would be also fine. PERIOD is coming from ALIEN_JDL_LPMPRODUCTIONTAG="LHC22o", for example.

Thanks!

Yours, Catalin.

shahor02:
@catalinristea thanks, I wanted to avoid using ALIEN_JDL stuff, but ok...
So, should I use `ALIGNLEVEL=0` only for `$ALIEN_JDL_LPMRAWPASS<3` of MAY,  JUN,  LHC22c,  LHC22d,  LHC22e, LHC22f` and `ALIGNLEVEL=1` for anything else?

chiarazampolli:
Hello,

The check should be done on the alien env var for MC anchoring, i need to check what it is. I will add it here.

Chiara

DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
```diff
@@ -172,18 +172,39 @@ else
   export ARGS_ALL_EXTRA=" --resources-monitoring 50 --resources-monitoring-dump-interval 50"
 fi
 
+#ALIGNLEVEL=0: before December 2022 alignment, 1: after December 2022 alignment
+ALIGNLEVEL=1
+if [[ $BEAMTYPE == "PbPb" || $PERIOD == "MAY" || $PERIOD == "JUN" || $PERIOD == "LHC22c" || $PERIOD == "LHC22d" || $PERIOD == "LHC22e" || $PERIOD == "LHC22f" ]]; then
+  ALIGNLEVEL=0
+fi
```
chiarazampolli, line 179: I would use something like this, to consider the anchoring, assuming we will not anchor to anything smaller than apass2 for 13.6 TeV, and apass5 for PbPb:
```
if [[ $BEAMTYPE == "PbPb" || $PERIOD == "MAY" || $PERIOD == "JUN" || $PERIOD == "LHC22c" || $PERIOD == "LHC22d" || $PERIOD == "LHC22e" || $PERIOD == "LHC22f" ]]; then
    ALIGNLEVEL=0
    if [[ $ALIEN_JDL_LPMPRODUCTIONTYPE == "MC" ]]; then
	# extract pass number
	ANCHORED_PASS=$ALIEN_JDL_LPMANCHOREPASSNAME
	ANCHORED_PASS_NUMBER=`echo $ANCHORED_PASS | sed 's/^apass//'`
	echo ANCHORED_PASS_NUMER = $ANCHORED_PASS_NUMBER
	if [[ $PERIOD == "MAY" || $PERIOD == "JUN" ]] && [[ $ANCHORED_PASS_NUMBER -gt 1 ]]; then
	    ALIGNLEVEL=1
	elif [[ $PERIOD == "LHC22c" || $PERIOD == "LHC22d" || $PERIOD == "LHC22e" || $PERIOD == "LHC22f" ]] && [[ $ANCHORED_PASS_NUMBER -gt 2 ]]; then
	    ALIGNLEVEL=1
	fi
    fi
fi
```
Note that apass2 of 13.6 TeV will be done with a mixed configuration, though... This will be trickier, we need to go to the run numbers. 
@catalinristea , what do you think?

catalinristea:
Ciao Chiara,

This seems the best we can do.
Thanks!

Yours, Catalin.

shahor02:
OK, I've applied the fix suggested by @chiarazampolli .

chiarazampolli:
Thanks! Merged.

PR #817, opened 2023-01-16
MCH: add async qc

aphecetche:
This PR introduces the long overdue async part of the QC for MCH.

Note that I've moved the `add_QC_JSON` up in the `qc-workflow.sh` script in order to be able to use it in the async part to construct the MCH part from parts.

@aferrero2707 FYI

### DIFF
--- a/DATA/production/qc-async/mch-digits.json
+++ b/DATA/production/qc-async/mch-digits.json
@@ -1,0 +1,65 @@
+{
+    "qc": {
+        "config": {
+            "database": {
+                "implementation": "CCDB",
+                "host": "localhost:6464",
+                "username": "not_applicable",
+                "password": "not_applicable",
+                "name": "not_applicable"
+            },
+            "Activity": {
+                "number": "42",
+                "type": "2"
+            },
+            "monitoring": {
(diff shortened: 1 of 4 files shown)


knopers8 (approved):
QC configs look OK, I can't judge for the script

chiarazampolli:
Hello @davidrohr , 
Maybe you want to take a look at the script before we merge?
Chiara

davidrohr:
Moving the code up is fine with me. What I am not fully sure is the change of the database host to localhost. Ccdb-test certainly does not make much sense, but why do we set it to localhost?

aphecetche:
@davidrohr good catch. That's actually a remnant of my local tests, not intended. Not in front of a (real) computer right now, will change it later today.

aphecetche:
@davidrohr I've reverted the unintended change to the ccdb in `QC_CONFIG_OVERRIDE` (and actually use add_QC_JSON in two more places for MCH, like I've been using in my latest tests)

chiarazampolli:
Hello @davidrohr , 
Can this be merged now?
Chiara

PR #819, opened 2023-01-16
Proposal for O2DPG tests

benedikt-voelkel:
* add a first test (example) for custom PWGDQ generator

* can be used to base an additional CIs for instance


benedikt-voelkel:
@sawenzel You think something like this makes sense?
As mentioned, we can base CIs on tests like this in the future or extend the FST.

We should maybe have a discussion on that in person.

sawenzel (approved):

PR #822, opened 2023-01-18
Do not produce plots on each update in EMC QC postprocessing

knopers8:
With the concerned value set to `true`, we save a new plot to QCDB for each new point in the plots (each `update` trigger). It is completely enough if we upload the plots only at `finalize`. cc @mfasDa


chiarazampolli:
I will merge after @mfasDa checks.
Thanks!

mfasDa (approved):
@knopers8 Thanks a lot for the fix! @chiarazampolli Please go ahead merging.

chiarazampolli (approved):

PR #835, opened 2023-01-26
Possibility to run meanVertex calib in async reco

chiarazampolli:
@noferini , @shahor02 , @catalinristea

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
@@ -136,6 +136,10 @@ if [[ -n "$ALIEN_JDL_DOTRDVDRIFTEXBCALIB" ]]; then
   export ADD_CALIB=1
 fi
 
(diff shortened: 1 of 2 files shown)


chiarazampolli:
@catalinristea : to extract mean vertex, one need to set:
```
WORKFLOWDETECTORS=ITS,TPC,TRD,TOF
DOMEANVTXCALIB=1
AODOFF=1
QCOFF=1
DOUPLOADSLOCALLY=1
```
in the JDL, and save in output the folder `GLO`, which will look like:
```
GLO/Calib/MeanVertex/o2-dataformats-MeanVertexObject_1674746042633_1660433693912_1663025703912.root 
```
We might need to fix 
```
export ADD_EXTRA_WORKFLOW="o2-calibration-ccdb-populator-workflow"
```
in case more are needed, but for now, we can go on like this.

shahor02 (approved):

chiarazampolli:
Thanks @shahor02 for checking. Merging. A slight improvement can be done for thd populator, but this already is good to start.

PR #841, opened 2023-02-01
TIMEFRAME_RATE_LIMIT depends on IR

chiarazampolli:
Hello @catalinristea , 
You can try the high IR with this. If it works, we can merge.
@noferini , @davidrohr , FYI.

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
@@ -369,6 +369,11 @@ else
   #
   if [[ "0$ASYNC_PASS_NO_OPTIMIZED_DEFAULTS" != "01" ]]; then
     export TIMEFRAME_RATE_LIMIT=3
+    if (( $(echo "$RUN_IR > 800000" | bc -l) )); then
+      export TIMEFRAME_RATE_LIMIT=1
+    elif (( $(echo "$RUN_IR < 50000" | bc -l) )); then
+      export TIMEFRAME_RATE_LIMIT=6
+    fi
     export OMP_NUM_THREADS=6
     export SHMSIZE=16000000000
   fi


DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
```diff
@@ -366,6 +366,9 @@ else
   #
   if [[ "0$ASYNC_PASS_NO_OPTIMIZED_DEFAULTS" != "01" ]]; then
     export TIMEFRAME_RATE_LIMIT=3
+    if (( $(echo "$RUN_IR > 800000" | bc -l) )); then
```
davidrohr, line 369: What is the content of $RUN_IR? Is is a complicated string, or just an integer?

chiarazampolli: The IR is a float that we extract from CCDB at the very beginning, it is here:
https://github.com/AliceO2Group/O2DPG/blob/99361c5fcf9c2db3bf97a57c57c7bfae78aaa1eb/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh#L149-L155

davidrohr (approved):
ok, first though one could do the comparison in bash, but apparently it cannot do floating point. Don't think there is a better solution than what you have...

catalinristea:
Ciao,

Thanks, we're giving it a try in grid.

Yours, Catalin.

catalinristea:
Ciao @chiarazampolli,

We need to merge this one, please, as it proved to be a day saver for HIR runs.
You may also want to update  TIMEFRAME_RATE_LIMIT to 6 for lowIR runs.
Thanks!

Yours, Catalin.

chiarazampolli:
Hello @catalinristea , 
I updated the RATE_LIMIT to 6, check that this is what you meant (I thought we used 3 on the grid...).
Chiara

catalinristea:
Ciao @chiarazampolli,

Thanks, but 3 should still be the default. 
6 could help (or not) for very low IR, like ~800 Hz of run 526689 in 22o, and maybe IR < 14 kHz for the runs of 22q. We found more helpful reducing the number of CTFs per job for such runs.

Yours, Catalin.

chiarazampolli:
Ah, sorry, I did not understand what you meant - too many settings recently to change by hand :-) 
Ok, I will put a threshold at 50 kHz.
Chiara

chiarazampolli:
@catalinristea , like this?

catalinristea:
Ciao @chiarazampolli,

Yes, this is what I meant. Thanks, and sorry for the noise.

Yours, Catalin.

chiarazampolli:
Not at all, and great that you checked!
Chiara

PR #842, opened 2023-02-01
use default MCH cluster resolution

pillot:
The default value is ok with the new alignment. Beside, I am going to remove this option to add 2 new ones for setting the resolution in x and y directions separately.

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
@@ -293,7 +293,7 @@ else
   export CONFIG_EXTRA_PROCESS_o2_mft_reco_workflow="MFTTracking.MFTRadLength=0.084;$MAXBCDIFFTOMASKBIAS_MFT"
 fi
 # ad-hoc settings for MCH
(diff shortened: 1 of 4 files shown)


chiarazampolli:
Hello @pillot , 
I could merge this PR, but we are still processing apass2 (for which the new MCH alignment is not there, see https://alice.its.cern.ch/jira/browse/O2-3514?filter=-3). If we change tag to fix the remaining runs, this will be taken. Would it be ok?
Otherwise, we need to wait till apass2 is completed.
Chiara

pillot:
Hi @chiarazampolli ,
Indeed, I missed that the tag could change for the apass2 and this be taken. Better to wait till apass2 is completed then.
Thanks!

chiarazampolli:
Hello @pillot ,

agreed, but please, remind me to merge before apass3.

chiara

chiarazampolli:
Hello @catalinristea , @pillot , @shahor02 , @noferini . 
I would merge this now, if you agree, because the remaining apass2 will not change tag (unless Catalin you want to try https://github.com/AliceO2Group/AliceO2/pull/10712 for low IR pp, but I would not). 
Let me know what you think.
Since the RelVal will be on 22m, and 22m is not affected by the new alignment, I guess this change is not needed there. @pillot , can you confirm?

pillot:
Yes, it is not needed for the RelVal, although it would also work fine with it.

chiarazampolli (approved):

PR #843, opened 2023-02-01
Add pythia8 box gun generator

fmazzasc:
Similar to the FairROOT box gun generator, but the decay of the injected particle is done by pythia8. Add minimal example to simulate omega and omega_c signal embedded into pythia background


benedikt-voelkel:
Hello @fmazzasc 
Looks good to me.

One question/recommendation: We recently added some testing mechanism to O2DPG and there is for instance a test for a PWGDQ generator.
It would be great if we had a test for your generator as well.
In the current scenario, a reasonable test could be:
* run simulation
* make sure, all particles you would expect are inside the kinematics

There would be an example here: https://github.com/AliceO2Group/O2DPG/blob/master/test/test-pwgdq_generator.sh
In the same directory there is also a README.

Let me know if you would like to add a basic test to this PR and also, let me know in case you have questions.

fmazzasc:
Hi @benedikt-voelkel ,
thanks a lot for the information. I added a simple test script based on what it is done for the pwgdq generator

MC/run/PWGHF/run_OmegaCInjected.sh
```diff
@@ -0,0 +1,32 @@
...
+SYSTEM=${SYSTEM:-pp}
+ENERGY=${ENERGY:-13600}
+[[ ${SPLITID} != "" ]] && SEED="-seed ${SPLITID}" || SEED=""
+
+${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow.py -e ${SIMENGINE} ${SEED} -eCM 13600 -col pp -colBkg pp -gen external -genBkg pythia8 -procBkg inel -j ${NWORKERS} -ns ${NSIGEVENTS} -nb ${NBKGEVENTS} -tf ${NTIMEFRAMES} -interactionRate 500000 -confKey "Diamond.width[2]=6." -mod "--skipModules ZDC" \
+        --embedding -ini $O2DPG_ROOT/MC/config/PWGHF/ini/GeneratorHFOmegaCEmb.ini 
+
+
+# run workflow
+${O2DPG_ROOT}/MC/bin/o2_dpg_workflow_runner.py -f workflow.json -tt aod --cpu-limit 64
```
benedikt-voelkel, line 32: For the test to work as expected you would need something like
```bash
TARGETTASK=${TARGETTASK:-aod}
# ...
${O2DPG_ROOT}/MC/bin/o2_dpg_workflow_runner.py -f workflow.json -tt ${TARGETTASK} --cpu-limit 64
```

benedikt-voelkel, line 32: probably setting the default to 64 CPUs is not convenient. That most likely would not run on a laptop as someone would maybe expect for a pp->pp embedding scenario.
It could be set via an env variable like `NWORKERS`. The same could be done for the memory a la `--mem-limit ${MEMLIMIT}`

MC/run/PWGLF/run_OmegaInjected.sh
```diff
@@ -0,0 +1,32 @@
...
+SYSTEM=${SYSTEM:-pp}
+ENERGY=${ENERGY:-13600}
+[[ ${SPLITID} != "" ]] && SEED="-seed ${SPLITID}" || SEED=""
+
+${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow.py -e ${SIMENGINE} ${SEED} -eCM 13600 -col pp -colBkg pp -gen external -genBkg pythia8 -procBkg inel -j ${NWORKERS} -ns ${NSIGEVENTS} -nb ${NBKGEVENTS} -tf ${NTIMEFRAMES} -interactionRate 500000 -confKey "Diamond.width[2]=6." -mod "--skipModules ZDC" \
+        --embedding -ini $O2DPG_ROOT/MC/config/PWGLF/ini/GeneratorLFOmegaEmb.ini 
+
+
+# run workflow
+${O2DPG_ROOT}/MC/bin/o2_dpg_workflow_runner.py -f workflow.json -tt aod --cpu-limit 64
```
benedikt-voelkel, line 32: see above comment regarding the CPU and memory

test/test-pythiabox_generator.sh
```diff
@@ -0,0 +1,26 @@
+
+#!/bin/bash
+
+########################################################################
+# A very basic test to test a custom generator implementation of PWGDQ #
```
benedikt-voelkel, line 5: PWGDQ --> PWGHF

MC/config/PWGLF/pythia8/generator_pythia8_box.C
```diff
@@ -0,0 +1,99 @@
+#include "Pythia8/Pythia.h"
+#include "FairGenerator.h"
+#include "FairPrimaryGenerator.h"
+#include "Generators/GeneratorPythia8.h"
+#include "TRandom3.h"
+#include "TParticlePDG.h"
+#include "TDatabasePDG.h"
+#include "TMath.h"
+#include <cmath>
```
benedikt-voelkel, line 9: This should be surrounded to guard the includes wrt `CLING` and `ROOTCLING` (see https://aliceo2group.github.io/simulation/docs/generators/generatorscustom.html#rootclingcling-related-problems)

benedikt-voelkel (changes requested):

benedikt-voelkel:
Hello @fmazzasc 
Thanks for the implementation of the test as well. I just had a few minor comments.
Let me know, if anything is unclear

MC/config/PWGLF/pythia8/generator_pythia8_box.C
benedikt-voelkel: Sorry, `using namespace Pythia8;` should go in there as well

fmazzasc: thanks, I have just pushed the fixes. Still with the guards the test fails:

`/home/fmazzasc/alice/sw/ubuntu2004_x86-64/O2DPG/fix-local1/MC/config/PWGLF/pythia8/generator_pythia8_box.C:66:7: error: unknown type name 'Particle'
      Particle myparticle;`

benedikt-voelkel: Thanks, I will take a look!

benedikt-voelkel: sorry for the confusion. Of course
```cpp
#include Pythia8/Pythia.h
using namespace Pythia8;
```
must be both outside the include guards.
There are no dictionaries for those and to get the type right etc, the header should be there.

I also tested quickly now and it works like a charm.
So once, you can add this tiny fix, we are good to merge.

fmazzasc: ok everything clear now:) i pushed the fix, thanks!

benedikt-voelkel (approved):

PR #846, opened 2023-02-03
MCH: update qc-async mch-tracks.json

aphecetche:
Following QC PR https://github.com/AliceO2Group/QualityControl/pull/1608
 the reading of the geometry is now done using the GrpGeomHelper and
 thus the JSON configuration must reflect that.

### DIFF
--- a/DATA/production/qc-async/mch-tracks.json
+++ b/DATA/production/qc-async/mch-tracks.json
@@ -33,11 +33,21 @@
                 "maxNumberCycles": "-1",
                 "dataSource": {
                     "type": "direct",
-                    "query": "trackMCH:MCH/TRACKS;trackMCHROF:MCH/TRACKROFS;trackMCHTRACKCLUSTERS:MCH/TRACKCLUSTERS;mchtrackdigits:MCH/CLUSTERDIGITS;mchgeo:GLO/Config/0?lifetime=condition&ccdb-path=GLO/Config/GeometryAligned"
+                    "query": "trackMCH:MCH/TRACKS;trackMCHROF:MCH/TRACKROFS;trackMCHTRACKCLUSTERS:MCH/TRACKCLUSTERS;mchtrackdigits:MCH/CLUSTERDIGITS"
                 },
                 "taskParameters": {
                     "maxTracksPerTF": "600",
                     "GID": "MCH"
+                },
+                "grpGeomRequest": {
+                    "geomRequest": "Aligned",
+                    "askGRPECS": "false",
+                    "askGRPLHCIF": "false",
+                    "askGRPMagField": "false",
+                    "askMatLUT": "false",
+                    "askTime": "false",
+                    "askOnceAllButField": "false",
+                    "needPropagatorD": "false"
                 }
             }
         }


chiarazampolli:
Hello @aphecetche , 
Did you check this with a reconstruction? Today I opened https://github.com/AliceO2Group/AliceO2/pull/10684 because otherwise for me it was not working. 
Also with QC attached to the o2-global-track-cluster-reader for me it does not work, but that is a different issue, which is being worked on.
So, the question is: how did you check it?

aphecetche:
@chiarazampolli tested with : 

```
WORKFLOWMODE="run" TFDELAY=10 NTIMEFRAMES=10 QC_REDIRECT_MERGER_TO_LOCALHOST=1 QC_CONFIG_PARAM="--local-batch=QC.root" WORKFLOW_DETECTORS="MCH" WORKFLOW_PARAMETERS="QC" WORKFLOW_DETECTORS_RECO="MCH" GEN_TOPO_WORKDIR=$HOME/tmp WORKFLOW_EXTRA_PROCESSING_STEPS="" CONFIG_EXTRA_PROCESS_o2_mch_reco_workflow="MCHDigitFilter.timeOffset=126" ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow="--condition-remap http://localhost:6464=GLO/Config/GeometryAligned" bash ../O2/prodtests/full-system-test/run-workflow-on-inputlist.sh CTF list.1.txt
```

where `list.1.txt` contains 1 CTF from /LHC22t/529691 (actually a CTF containing only MCH+MID, but I don't think that matters), and the condition-remap is to use latest Javier's alignment (so not relevant either for the QC test I would say)

Note that prior to this one the MCH QC tracks was not reading properly the geometry indeed.

aphecetche:
@chiarazampolli actually I'd need to modify this PR to add a second track task... We currently have two different QC tasks for tracks (the old one and the new one) and I've just added the new one, but then I realised it does not produce all the histograms of the old one, namely the ClusterPerDualSampa (@aferrero2707 FYI) that is then used to assess the sim-data likeliness is missing... 
So if possible I'd like to add the old task as well to get this kind of crucial piece of information for apass3. Is that still possible ?

chiarazampolli:
Hello @aphecetche , 
Yes, please, go ahead and modify this PR.
Cheers,
Chiara

aphecetche:
@chiarazampolli @aferrero2707 I did change the PR but it's actually worse than I thought. At some point with the new task we lost the cluster map but in the old task where the cluster map is still there there's a bug and the cluster map is wrong...

So there are two things to fix : 

- fix the JSON configuration to remove the `mchgeo:GLO/Config/0?lifetime=condition&ccdb-path=GLO/Config/GeometryAligned` part (following the merging of  https://github.com/AliceO2Group/QualityControl/pull/1608. This PR will do.
- fix the QualityControl code of the old track task if we want the cluster map back and correct

I'll work on the second point today even if I assume we won't be able to have it used for apass3, given the time constrains, am I right ?

chiarazampolli:
> @chiarazampolli @aferrero2707 I did change the PR but it's actually worse than I thought. At some point with the new task we lost the cluster map but in the old task where the cluster map is still there there's a bug and the cluster map is wrong...
> 
> So there are two things to fix :
> 
>     * fix the JSON configuration to remove the `mchgeo:GLO/Config/0?lifetime=condition&ccdb-path=GLO/Config/GeometryAligned` part (following the merging of  [Mch geometry from ccdb fetcher QualityControl#1608](https://github.com/AliceO2Group/QualityControl/pull/1608). This PR will do.
> 
>     * fix the QualityControl code of the old track task if we want the cluster map back and correct
> 
> 
> I'll work on the second point today even if I assume we won't be able to have it used for apass3, given the time constrains, am I right ?

Hello Laurent,

As you might have seen from the mail to WP13 list, we will freeze the software tonight.

Cheers,

Chiara

aphecetche:
Well, anyways it seems I’m no longer able to get the GrpGeomHelper to work as before with the QC (I’ve updated both O2 and QualityControl), so my track task is consistently crashing if I try to access the geometry, so this will require more investigations/time to debug, I’m afraid…

chiarazampolli:
Hello @aphecetche , 
Should I merge?
Chiara

aphecetche:
@chiarazampolli yes, please. That should make the "geometry is missing" messages (that you were observing in your tests) disappear

chiarazampolli (approved):

chiarazampolli:
Thanks!

PR #850, opened 2023-02-06
Don't place more than 8 ITS trackers per NUMA domain

davidrohr:


martenole:
@davidrohr  thanks! Should I update CONSUL to have this already now at P2?

davidrohr:
I wanted to try it first, but all deployments failed on the FLPs...
But my FST just failed again, so I think this is not sufficient anyway...

martenole:
Too bad

PR #858, opened 2023-02-08
Update analysis QC to run with compatibility mode

njacazio:
@benedikt-voelkel what do you think? Tagging also @mfaggin @noferini @chiarazampolli

### DIFF
--- a/MC/analysis_testing/o2dpg_analysis_test_workflow.py
+++ b/MC/analysis_testing/o2dpg_analysis_test_workflow.py
@@ -337,7 +337,8 @@ def add_analysis_tasks(workflow, input_aod="./AO2D.root", output_dir="./Analysis
         needs: iter (optional)
             if specified, list of other tasks which need to be run before
     """
-    input_aod = abspath(input_aod)
+    if not input_aod.startswith("alien://"):
+        input_aod = abspath(input_aod)
     if input_aod.endswith(".txt") and not input_aod.startswith("@"):
         input_aod = f"@{input_aod}"
     data_or_mc = ANALYSIS_VALID_MC if is_mc else ANALYSIS_VALID_DATA
--- a/MC/analysis_testing/o2dpg_analysis_test_workflow.py
+++ b/MC/analysis_testing/o2dpg_analysis_test_workflow.py
@@ -319,7 +319,7 @@ def add_analysis_post_processing_tasks(workflow):
         task["cmd"] = f"root -l -b -q {post_processing_macro}{cmd}"
         workflow.append(task)
 
-def add_analysis_tasks(workflow, input_aod="./AO2D.root", output_dir="./Analysis", *, analyses_only=None, is_mc=True, config=None, needs=None):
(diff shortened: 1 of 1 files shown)


chiarazampolli (approved):
Looks ok to me!

benedikt-voelkel:
Did you test it on one of the failing productions @njacazio ?
I think for now this fix is okay, but as soon as we stumble upon similar things we can generalise this maybe a little bit.

njacazio:
I did, on the one suggested by Catalin.
```
alien_cp alien:///alice/data/2022/LHC22m/523141/apass2/2150/o2_ctf_run00523141_orbit0170111871_tf0000014382_epn051/001/AO2D.root file:a.root
${O2DPG_ROOT}/MC/bin/o2dpg_analysis_test_workflow.py -f a.root --only-analyses Efficiency --autoset-converters
${O2DPG_ROOT}/MC/bin/o2_dpg_workflow_runner.py -f workflow_analysis_test.json -tt Analysis_Efficiency
```
I tried by feeding directly the input file `alien:///alice/data/2022/LHC22m/523141/apass2/2150/o2_ctf_run00523141_orbit0170111871_tf0000014382_epn051/001/AO2D.root` instead of `a.root` but I cannot open the file, probably the delay is too large.

MC/analysis_testing/o2dpg_analysis_test_workflow.py
```diff
@@ -428,6 +452,7 @@ def main():
     parser.add_argument("--period-name", dest="period_name", help="period name")
     parser.add_argument("--config", help="overwrite the default config JSON. Pass as </path/to/file>, will be automatically configured to json://")
     parser.add_argument("--only-analyses", dest="only_analyses", nargs="*", help="filter only on these analyses")
+    parser.add_argument("--autoset-converters", dest="autoset_converters", action="store_true", help="Compatibility mode to automatically set the converters for the analysis")
```
mfaggin, line 455: Thanks @njacazio it looks very good to me! Just for completeness: I'd rename `autoset_converters` to something like `autoset_collision_converter` or something similar

benedikt-voelkel (approved):
Ok, let's go ahead with this hot fix for now

MC/analysis_testing/o2dpg_analysis_test_workflow.py
benedikt-voelkel: I think, this option makes sense as is. We might need to add other converters in the future. I think this is what was foreseen here?!

njacazio: No, these should be general, it could also work on e.g. mcParticles and so on

njacazio: Yes, thanks @benedikt-voelkel

mfaggin: apologize, you are right, it makes sense as it is because the proper converter is switched-on if a table is not found

PR #862, opened 2023-02-08
hftodielectronsMCconfiguration

rbailhac:
Dear all, 

this pull request is meant for the preparation of enhanced dielectrons MC from correlated heavy-flavour hadron decays. 

For charms, we would prefer to use inelastic pythia 8 with the colour reconnection mode 2, since it reproduced better the amount of baryons, relevant for DCAee templates.

For beauty, to save CPU, we would switch on only the bbbar process. If this is a problem, we can go back to the already existing pythia8 configuration for hf (both ccbar and bbbar) cfg.

We are opened to any comments....

Best regards,

Raphaelle


benedikt-voelkel:
@rbailhac 
The failing CI is unrelated. We put it in place to do some O2DPG-related tests, but there are problems with running the simulation apparently.
None of those tests incorporates any of the files pushed by you, so there is nothing to worry about on your side.
I will discuss with @sawenzel and @TimoWilken on how to proceed with the test.

gconesab (approved):

PR #864, opened 2023-02-10
Extend log file printing for tests when failing

benedikt-voelkel:
* incorporate printing of relevant lines of further log files in case of failure. Those might be produced by launched sub-tasks of the test such as those produced by the simulation or the o2_dpg_workflow_runner.py


benedikt-voelkel:
@sawenzel can we merge this? At least we will have some more insight in the future when e.g. sim WFs fail as this PR provides us with better log verbosity in case of failure.
 That will be in particular useful for the O2DPG sim-related CI

chiarazampolli (approved):

PR #866, opened 2023-02-10
source gen_topo_helper_functions.sh from O2DPG_ROOT directly

shahor02:
avoid overriding MYDIR in setenv.sh

### DIFF
--- a/DATA/common/setenv.sh
+++ b/DATA/common/setenv.sh
@@ -4,8 +4,7 @@
 if [[ -z $SOURCE_GUARD_SETENV ]]; then
 SOURCE_GUARD_SETENV=1
 
-MYDIR="$(dirname $(realpath $0))"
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Hi @shahor02 , 
But this won't allow to overwrite it. I guess it is ok, but maybe you should check with @davidrohr , @martenole ?

shahor02:
It was actually overwriting, we checked with Ole. On top of that, `MYDIR="$(dirname $(realpath $0))"` was assigning the directory not of the setenv.sh location (i.e. `DATA/common`)  but of the of the top level script which in turn was sourcing `setenv.sh`.

chiarazampolli:
Ok, thanks for the explanation and sorry that this was creating an issue at P2 (see https://github.com/AliceO2Group/AliceO2/pull/10750#issuecomment-1426247454).

PR #867, opened 2023-02-11
properly propagate the bcPatternFile parameter to all tasks

pbuehler:


gconesab:
Hi @sawenzel  and @benedikt-voelkel 

Since this affects the handling of access to ccdb, can you have a look and approve if really ok? I have no problem with it.

Cheers ... Gustavo

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -515,7 +515,7 @@ def getDPL_global_options(bigshm=False, ccdbbackend=True):
    QEDXSecExpected=35237.5  # expected magnitude of QED cross section
    PreCollContextTask=createTask(name='precollcontext_' + str(tf), needs=precollneeds, tf=tf, cwd=timeframeworkdir, cpu='1')
    PreCollContextTask['cmd']='${O2_ROOT}/bin/o2-steer-colcontexttool -i ' + signalprefix + ',' + str(INTRATE) + ',' + str(args.ns) + ':' + str(args.ns) + ' --show-context ' + ' --timeframeID ' + str(tf-1 + int(args.production_offset)*NTIMEFRAMES) + ' --orbitsPerTF ' + str(orbitsPerTF) + ' --orbits ' + str(orbitsPerTF) + ' --seed ' + str(TFSEED) + ' --noEmptyTF'
-   PreCollContextTask['cmd'] += ' --bcPatternFile ccdb'  # <--- the object should have been set in (local) CCDB
+   PreCollContextTask['cmd'] += ' --bcPatternFile ' + str(args.bcPatternFile)  # <--- the object should have been set in (local) CCDB
```
sawenzel, line 518: No, this is not ok ! The reason is that the first task "GRP creation" has the possibility to initialise from bcPatternFile **and** it will publish this as a local CCDB object. Thereafter, everyone else should simply read this object. If this doesn't work it's a  bug that needs fixing otherwise.

pbuehler: ok I see. Then I need to check this local ccdb mechanism again.

PR #872, opened 2023-02-13
Analysis QC: align to O2Physics

njacazio:


njacazio:
I propose to format the json files to allow easier comparison in the versioning, I added a file to format them.
Also, added the `pid-tpc-base` workflow that was needed, I tested it and every workflow works. 
@benedikt-voelkel

benedikt-voelkel (approved):
Merging and following up on the remaining issues to enable all analysis tests again.

PR #880, opened 2023-02-15
Updates for topology generation: use scripts from updateable RPMs and write log files

davidrohr:


davidrohr:
not tested anyway, merging

PR #881, opened 2023-02-15
Protection for noisy MID runs and possibility to disable 3D currents

chiarazampolli:
@catalinristea 
@dstocco , I am now excluding MID from the two runs giving issues - anyway even with code fixes, they are BAD for MID. @matthias-kleiner , I am adding the possibility to disable the 3D currents, just in case.


chiarazampolli:
Merged since it is not tested in the CI.

PR #882, opened 2023-02-15
MFT: Renaming json files and minor updates.

tomas-herman:


tomas-herman:
This PR has to be deployed together with 

QC: https://github.com/AliceO2Group/QualityControl/pull/1628
CW: https://github.com/AliceO2Group/ControlWorkflows/pull/540

as these changes are interconnected. Also the automatic checks might fail due this reason.

knopers8 (approved):

martenole (approved):
The changes are fine with me. The other PRs are already merged. Once a new QC version is tagged we can merge this. This commit should not be deployed at P2 before the new QC version is installed there as well.

davidrohr:
In my opinion wr can also merge this now. We have a workimg o2pdpsuite installation at p2, and in case we need zo update o2dpg, we have to be careful.

One remark though: the conditions db in the qc-sync folder is set to localhost:8084 in this pr, which is not correct.
Qc-sync is used for the CI for sync processing. Files for sync processing at p2 are in consul.

Probably we should add a qc option to rewrite the conditions db setting.

tomas-herman:
Hi,

@chiarazampolli sorry about the issue with async, in the future we will make the O2DPG changes in 2 separate PRs for online and async. 

@davidrohr this is probably due to my lack of understanding of the use and implementation of the qc-sync in O2DPG. I tried to copy what we use at P2 basically. What should be the correct conditions db address? I can make additional commit to fix it asap. 

Cheers,
Tomas

davidrohr:
merging this. For the conditions, I'll add a global override for all files from qc-sync. In that way, they can simply be copied from consul.

PR #883, opened 2023-02-15
Fix topology generation failing silently

martenole:
Fix for silent topology generation failure: in the arithmetic context variable names are substituted with their values. Empty variables are substituted with 0: http://mywiki.wooledge.org/ArithmeticExpression
The forced `true` at the end is what caused the topology generation not to fail although it did not work properly

### DIFF
--- a/DATA/production/workflow-multiplicities.sh
+++ b/DATA/production/workflow-multiplicities.sh
@@ -6,21 +6,21 @@ SOURCE_GUARD_MULTIPLICITIES=1
 # ---------------------------------------------------------------------------------------------------------------------
 # Threads
 
-[[ -z $NITSDECTHREADS ]] && NITSDECTHREADS=2
-[[ -z $NMFTDECTHREADS ]] && NMFTDECTHREADS=2
+if [[ -z $NITSDECTHREADS ]]; then NITSDECTHREADS=2; fi
+if [[ -z $NMFTDECTHREADS ]]; then NMFTDECTHREADS=2; fi
 
-[[ -z $NMFTTHREADS ]] && NMFTTHREADS=2
+if [[ -z $NMFTTHREADS ]]; then NMFTTHREADS=2; fi
 
-[[ -z $SVERTEX_THREADS ]] && SVERTEX_THREADS=$(( $SYNCMODE == 1 ? 1 : 2 ))
+if [[ -z $SVERTEX_THREADS ]]; then SVERTEX_THREADS=$(( SYNCMODE == 1 ? 1 : 2 )); fi
 # FIXME: multithreading in the itsmft reconstruction does not work on macOS.
 if [[ $(uname) == "Darwin" ]]; then
     NITSDECTHREADS=1
     NMFTDECTHREADS=1
 fi
 
-[[ $SYNCMODE == 1 ]] && NTRDTRKTHREADS=1
+if [[ $SYNCMODE == 1 ]]; then NTRDTRKTHREADS=1; fi
 
-[[ -z $NGPURECOTHREADS ]] & NGPURECOTHREADS=-1 # -1 = auto-detect
+if [[ -z $NGPURECOTHREADS ]]; then NGPURECOTHREADS=-1; fi # -1 = auto-detect
 
 # ---------------------------------------------------------------------------------------------------------------------
 # Process multiplicities
@@ -34,10 +34,10 @@ if [[ ! -z $OPTIMIZED_PARALLEL_ASYNC ]]; then
   # Tuned multiplicities for async processing
   if [[ $OPTIMIZED_PARALLEL_ASYNC == "pp_8cpu" ]]; then
     NGPURECOTHREADS=5
-    [[ -z $TIMEFRAME_RATE_LIMIT ]] && TIMEFRAME_RATE_LIMIT=3
+    if [[ -z $TIMEFRAME_RATE_LIMIT ]]; then TIMEFRAME_RATE_LIMIT=3; fi
   elif [[ $OPTIMIZED_PARALLEL_ASYNC == "pp_1gpu" ]]; then
-    [[ -z $TIMEFRAME_RATE_LIMIT ]] && TIMEFRAME_RATE_LIMIT=8
-    [[ -z $SHMSIZE ]] && SHMSIZE=24000000000
+    if [[ -z $TIMEFRAME_RATE_LIMIT ]]; then TIMEFRAME_RATE_LIMIT=8; fi
+    if [[ -z $SHMSIZE ]]; then SHMSIZE=24000000000; fi
     N_TOFMATCH=2
     N_MCHCL=3
     N_TPCENTDEC=3
@@ -47,8 +47,8 @@ if [[ ! -z $OPTIMIZED_PARALLEL_ASYNC ]]; then
     NGPURECOTHREADS=8
     NTRDTRKTHREADS=2
   elif [[ $OPTIMIZED_PARALLEL_ASYNC == "pp_4gpu" ]]; then
-    [[ -z $TIMEFRAME_RATE_LIMIT ]] && TIMEFRAME_RATE_LIMIT=32
-    [[ -z $SHMSIZE ]] && SHMSIZE=80000000000
+    if [[ -z $TIMEFRAME_RATE_LIMIT ]]; then TIMEFRAME_RATE_LIMIT=32; fi
+    if [[ -z $SHMSIZE ]]; then SHMSIZE=80000000000; fi
     NGPURECOTHREADS=8
     NTRDTRKTHREADS=2
     NGPUS=4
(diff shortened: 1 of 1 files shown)


martenole:
Not yet verified, please don't merge yet

martenole:
Ok, now I verified that the topology generation is working with these changes. Also it is properly failing if I for example revert commit https://github.com/AliceO2Group/O2DPG/commit/2d9ec70f8452de1e550743659aea21163314c640 where the script path was corrected.
The problem is, that in bash if I use the shortcut for comparisons, the return code is not necessarily zero
```
oschmidt@lenovo:~/alice/O2DPG$ [[ "0$FOO" == "01" ]] && echo Hi
oschmidt@lenovo:~/alice/O2DPG$ echo $?
1
```
Whereas it is always zero if I use the longer version
```
oschmidt@lenovo:~/alice/O2DPG$ if [[ "0$FOO" == "01" ]]; then echo Hi; fi
oschmidt@lenovo:~/alice/O2DPG$ echo $?
0
```
About the usage if (( $varname )) instead of (( varname )) I am not sure. Seems both is OK. If we use (( $varname )) we will get an error if varname is not set, otherwise it will be evaluated to zero. Not sure what we prefer.

DATA/production/workflow-multiplicities.sh
```diff
@@ -6,21 +6,21 @@ SOURCE_GUARD_MULTIPLICITIES=1
...
 # FIXME: multithreading in the itsmft reconstruction does not work on macOS.
 if [[ $(uname) == "Darwin" ]]; then
     NITSDECTHREADS=1
     NMFTDECTHREADS=1
 fi
 
-[[ $SYNCMODE == 1 ]] && NTRDTRKTHREADS=1
+if [[ $SYNCMODE == 1 ]]; then NTRDTRKTHREADS=1; fi
 
-[[ -z $NGPURECOTHREADS ]] & NGPURECOTHREADS=-1 # -1 = auto-detect
```
martenole, line 23: The typo here always forced auto-detect I think. Not sure if you have used external `NGPURECOTHREADS` in the past for some checks @davidrohr ?

davidrohr:
These short conditi9nal statements were exactly the reason for the true at thr end, to make sure the script does not return false if the last condition was false.

Kind Regards
David Rohr

martenole:
Yes, but forcing a good return value via true at the end leads to the topology generation failing silently. While with these changes I get
```
stderr: /home/marten/O2DPG/DATA/production/setenv.sh: line 8: /home/marten/O2DPG/DATA/production/gen_topo_helper_functions.sh: No such file or directory
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/dpl-workflow.sh: line 23: workflow_has_parameter: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/dpl-workflow.sh: line 24: workflow_has_parameter: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 36: add_comma_separated: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 36: add_comma_separated: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 39: add_comma_separated: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 53: has_detectors_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 54: has_detectors_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 55: has_detectors_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 56: has_detectors_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 57: has_detectors_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 58: has_detectors_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 59: has_detectors_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 60: has_detectors_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 61: has_detectors_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 66: has_detector_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 64: has_detector: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 66: has_detector_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 66: has_detector_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 66: has_detector_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 64: has_detector: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 66: has_detector_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 66: has_detector_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 66: has_detector_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 64: has_detector: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 64: has_detector: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 64: has_detector: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 64: has_detector: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 66: has_detector_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 64: has_detector: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/workflow-setup.sh: line 66: has_detector_reco: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/dpl-workflow.sh: line 30: workflow_has_parameter: command not found
/opt/alisw/el8/O2/a5b95dbfa4c6e216847d93e985f28e8dc3a7b9de-1/prodtests/full-system-test/dpl-workflow.sh: line 40: has_detector: command not found
workflow-multiplicities.sh failed
```
If I just put back the true at the end then I get an empty workflow and no errors
```
#Workflow command:
echo '{}' | \
o2-dpl-raw-proxy --session default --severity info --shm-segment-id 0 --shm-segment-size 120259084288  --early-forward-policy noraw --infologger-severity important --monitoring-backend influxdb-unix:///tmp/telegraf.sock --resources-monitoring 15 --bad-alloc-max-attempts 60 --bad-alloc-attempt-interval 1000 --no-cleanup --shm-no-cleanup on --shmid 1 --fairmq-rate-logging 0 --child-driver 'numactl --membind 0 --cpunodebind 0' --timeframes-rate-limit 684 --timeframes-rate-limit-ipcid 0 --dataspec "dd:FLP/DISTSUBTIMEFRAME/0;RAWIN0:CPV/RAWDATA;RAWIN1:FDD/RAWDATA;RAWIN2:FT0/RAWDATA;RAWIN3:FV0/RAWDATA;RAWIN4:ITS/RAWDATA;RAWIN5:MFT/RAWDATA;RAWIN6:TPC/RAWDATA" --readout-proxy "--channel-config \"name=readout-proxy,type=pull,method=connect,address=ipc://@tf-builder-pipe-0,transport=shmem,rateLogging=1;name=metric-feedback,type=pull,method=connect,address=ipc://@metric-feedback-0,transport=shmem,rateLogging=0\"" --timeframes-shm-limit 60129542144   | \
o2-dpl-run --session default --severity info --shm-segment-id 0 --shm-segment-size 120259084288  --early-forward-policy noraw --infologger-severity important --monitoring-backend influxdb-unix:///tmp/telegraf.sock --resources-monitoring 15 --bad-alloc-max-attempts 60 --bad-alloc-attempt-interval 1000 --no-cleanup --shm-no-cleanup on --shmid 1 --fairmq-rate-logging 0 --child-driver 'numactl --membind 0 --cpunodebind 0' --timeframes-rate-limit 684 --timeframes-rate-limit-ipcid 0  -b --dds-workflow-suffix _reco2
```

DATA/production/workflow-multiplicities.sh
davidrohr: thx for spotting, that was never used indeed!

davidrohr:
ok, so what happened is that the helper_function_script was not sources, due to the wrong path that was fixed quite some time ago by Ruben.
This didn't create any error message since we were not checking whether scripts were sources successfully. This is fixed by https://github.com/AliceO2Group/AliceO2/pull/10782 and https://github.com/AliceO2Group/O2DPG/pull/885

I still believe the true at the end of workflow-multiplicities is correct, otherwise we can get random non-zero return codes from the short conditionals.

What is true is that we do not check the return value of each and every line.
In order to do so, we would need to change all the scripts and replace all the `&&` /  `||` conditionals to full `if` `else` `fi` expressions, and then run with `set -eu`. Not sure if we want this.

As it is, we have indeed a problem, that a typo is not always caught. Although, we already have a check for certain error messages in the `parse` script. I'll add `command not found` etc. there as well, such that it should always trigger a failure.

davidrohr:
Added such a check to https://github.com/AliceO2Group/O2DPG/pull/885
I also added your typo fix for the GPU NTHREADS.

shahor02:
> ok, so what happened is that the helper_function_script was not sources, due to the wrong path that was fixed quite some time ago by Ruben.

out of curiosity, which path do you refer to? Did I screw it up or fixed?

davidrohr:
You fixed it, the $MYDIR being overwritten

martenole:
Ok, I can close this one then. An alternative to putting the longer `if`, `else`, `fi` would be to add each line with the short conditional form a `|| true` at the end. Then we could run with `set -eu` as you said and would always exit on failure.
In that case your checks if scripts have been sourced successfully would not be needed. But of course we'd had to change many more other lines. Especially, since in addition for each line where  we do something like `has_detector ITS && ...` there we would need to add the `|| true` as well, since such checks also can give non zero return values.

davidrohr:
I also thought about `|| true`, but that is too much a mess. I would not go that way.

For the `set -eu`: somehow I would prefer this, but changing all conditionals to if/else will make the scripts much longer...
Not sure if we want that.

martenole:
Maybe we should still go for it? Then we can merge this as a start :)

From what I see the dpl-workflow script is the only one which would get significantly longer. For the other scripts the changes are similar to this PR and I don't think it would make them less readable.

shahor02:
> What is true is that we do not check the return value of each and every line.
> In order to do so, we would need to change all the scripts and replace all the `&&` / `||` conditionals to full `if` `else` `fi` expressions, and then run with `set -eu`

we could move to the beginning of the script all function calls with full checks, assigning there local env.vars which are then used with `&& / ||` as now, e.g.
```
# in the beginning
if [[ has_detector ITS ]]; then 
  has_detector_ITS = 1
fi
```
then use 
`[[ ! -z $has_detector_ITS ]] &&`
instead of `has_detector ITS`

davidrohr:
Ole and me discussed this privately a bit more, also `-u` might be interesting. But it has other side effects. In principle, I'd leave it as is for the time being. Particularly, I'd like to avoid complicating the scripts and making them longer. The additional checks I added to the parser should reveal most such issues anyway.

martenole:
Superseded by https://github.com/AliceO2Group/O2DPG/pull/893

PR #887, opened 2023-02-16
Add configuration for event filtering

mpuccio:
Dear @chiarazampolli﻿,

I would put the file here, as we do not foresee to change it period by period.
Things might still change in the next few weeks, but this should be fine for the first running in cascade to the reconstruction.

Thanks a lot,
Regards,
Max


chiarazampolli:
Hello @mpuccio , 
I would move this file in a dedicated folder under the path:
`/DATA/production/configurations`
especially since you say it won't change. 
The folder with $year/$period was created with the idea to have one setting per period, but finally it turned to be too much work to maintain it. We keep using it for practical reasons, but what is not pass/period dependent (as you say the trigg conf is not), should not go in this type of structure. 
 
Cheers,
Chiara

mpuccio:
Ciao @chiarazampolli 
we might still change the configuration if the TPC calibration arrives. Is this still with your proposal? If so, I will move the configuration file!
Cheers,
Max

PR #888, opened 2023-02-16
Removing storing TOF output

chiarazampolli:
@noferini , as you confirmed we can do.


noferini (approved):
Sure

chiarazampolli:
Thanks @noferini !
Merging, it is not tested in the CI.

PR #892, opened 2023-02-17
TPC map creation macro must run compiled

martenole:


martenole:
Tested on lxplus with `O2PDPSuite/nightly-20230215-1` loaded from cvmfs

chiarazampolli:
Thanks @martenole !

PR #893, opened 2023-02-17
Additional robustness for topology generation

martenole:
Following discussion in https://github.com/AliceO2Group/O2DPG/pull/883 these changes would allow to add additional flags `set -eu` to the workflow generation scripts dpl-workflow.sh and aggregator-workflow.sh

I am not sure if we want to go this way, but I started now with the changes and somehow was also interested to see if its possible at all. Will open a draft PR to O2 where I enabled `set -eu` and could run different workflows successfully. Of course I could not check all combinations, so some issues might prevail. So maybe it would be better to not enable them right away.

At least I already found some small typos with the unbound variables check in the setenv_calib (for `CALIB_PHS_TURNONCALIB` and `CALIB_PHS_BADMAPCALIB`).

In general one always needs to think for one variable if one would expect it to be set from outside, so then it should throw an error, or if it would be OK if its empty.


martenole:
Moving this to WIP, since it should be merged not before March 6 so that I am available for testing when its deployed at P2

martenole:
Since we are only going for the `set -u` check for unbound variables I don't need to change everywhere the short form for `if ..; then..; else`. I tested it already with different settings, sync and async, but I am not fully sure I have not missed anything. So I would not enable the flag already now. Anyhow can only be done after both PRs in O2DPG and O2 are merged

martenole:
@davidrohr one more thing I forgot to mention: I still want to get rid of workflow-setup.sh and put the content in setenv and gen_topo_helper_functions. In order to do that I need to copy it first from O2 to O2DPG in order to not break anything. Because of the include guard it should not be a problem. If its fine with you then I merge this once the checks are through.

davidrohr:
should be fine, if the CI passed I would merge

PR #902, opened 2023-02-22
Global init section for MC graph workflows

sawenzel:
Introducing a mechanism allowing to define
global environment variable that should be used in all tasks of a MC workflow.

This is done by defining a special task '__global_init_task__' which is parsed by the runtime engine. This task will not do anything else than defining environment variables for the moment.


sawenzel:
Used to finish the time-machine feature for CCDB access.

PR #925, opened 2023-03-08
MFT QC: Remove async task from sync QC. It is crashing, and anyway not sure if it should be there

davidrohr:
@tomas-herman : Is this async task supposed to be part of the sync QC?
In any case, this task is crashing and breaks the full system test on the EPNs.

### DIFF
--- a/DATA/production/qc-sync/mft-full.json
+++ b/DATA/production/qc-sync/mft-full.json
@@ -88,32 +88,6 @@
         "remoteMachine": "alio2-cr1-qme05.cern.ch",
         "remotePort": "47797",
         "localControl": "odc"
-      },
-      "MFTAsyncTask": {
-        "active": "true",
-        "className" : "o2::quality_control_modules::mft::QcMFTTrackTask",
-        "moduleName": "QcMFT",
-        "detectorName": "MFT",
-        "cycleDurationSeconds": "120",
-        "maxNumberCycles": "-1",
-        "dataSource" : {
-          "type" : "direct",
-          "query" : "tracks:MFT/TRACKS/0;tracksrofs:MFT/MFTTrackROF/0"
-        },
-        "location": "local",
-        "localMachines": [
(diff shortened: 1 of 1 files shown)


tomas-herman:
Hi @davidrohr , 

the async QC task is supposed to be part of sync QC. It has an old name but when @knopers8 and @Barthelemy rename the online QC tasks and checks it will be called Track task. 

As you can see here: https://github.com/AliceO2Group/O2DPG/blob/af386b9e1bcb69e6f01d3d9388010d20cc49fce1/DATA/production/qc-workflow.sh#L106-L110  we are using 2 QC workflows, one with track task and one without track task (in the current code still called async task). The corresponding json file is then selected. If this is crashing the full system test we need to find the issue and fix it. It could me some mistake in my PRs with the renaming if the other components were unchanged.

davidrohr:
ok, then it is a bit poorly named :)
ok, will retry with that task, and try to get a stack trace. But if I can reproduce the crash, I'd remove it for now, and you can revert this once the issue is fixed.

tomas-herman:
Yes, the name is really confusing at the moment so that is why we want to rename it to Track Task. But it was called async task for historical reasons as initially it was planned to run only in async...

Anyway, thanks for looking into it. We are running this workflows also at p2 from consul json files so if it is a mistake in the changes I did it will affect also the QC at p2 after the SW update on Monday.

davidrohr:
@tomas-herman : I cannot reproduce it any more now, but I also updated QualityControl to the tag that was released today. Can it be that it would have crashed with the old QualityControl?

davidrohr:
seems to work with new QC

tomas-herman:
> @tomas-herman : I cannot reproduce it any more now, but I also updated QualityControl to the tag that was released today. Can it be that it would have crashed with the old QualityControl?

Yes, it would have failed with the old QC.

PR #930, opened 2023-03-09
Workflow testing and small updates

benedikt-voelkel:
* add one full WF run when MC/bin changes G3, 1 TF 1, pp event

* add help message to run_workflow_tests.sh


MC/bin/tests/embedding_test.sh
```diff
@@ -15,21 +15,17 @@
...
 
 # create workflow
 ${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow.py -eCM 5020 -col pp -gen pythia8 -proc ${PYPROCESS} \
-                                           -colBkg PbPb -genBkg pythia8 -procBkg "heavy_ion" \
-                                           -tf ${NTIMEFRAMES} -nb ${NBKGEVENTS}              \
-                                           -ns ${NSIGEVENTS} -e ${SIMENGINE}                 \
-                                           -j ${NWORKERS} --embedding -interactionRate 50000 \
-                                           --include-analysis -run 310000 ${SEED}
+                                           -tf ${NTIMEFRAMES} -ns ${NSIGEVENTS} -e ${SIMENGINE} \
+                                           -j ${NWORKERS} --include-analysis -run 310000 ${SEED}
```
sawenzel, line 31: this is no longer an embedding test? so maybe the filename should be renamed. Also better to give the interaction rate on the command line and specify a fixed seed (at the moment it's random which is not optimal for reproducable testing)

benedikt-voelkel: We can of course leave it as an embedding test, namely 1 pp into 1 PbPb in that case?

sawenzel (approved):

PR #932, opened 2023-03-09
adapting to run split wf

chiarazampolli:
@davidrohr , @catalinristea , @noferini : This is to be able to run the WF splitting it in 3 parts. It needs https://github.com/AliceO2Group/AliceO2/pull/10928 still, then we could try it on the grid.


davidrohr (approved):

davidrohr:
As mentioned on mattermost, I would add checks for the return values (in general). But can also be done later.

chiarazampolli:
Yes, i wanted to add it here, but i had to leave. This way this is now done. Next the return code.

PR #935, opened 2023-03-10
Update analysis

benedikt-voelkel:


benedikt-voelkel:
The changes are fully transparent, except for one thing:
It is right now not possible to run
```bash
o2dpg_analysis_test_workflow.py --config <local-configuration>
```
Than feature will come back if needed

chiarazampolli (approved):

PR #940, opened 2023-03-13
Make ready for AnalysisQC

benedikt-voelkel:
* revise analysis_testing.sh to include more functionality

* re-enable
  * PIDFull -> works just fine
  * K0STrackingEfficiencyQA -> all prcess functions of lambdakzero-preselector were disabled; worked before but something in the framework seems to have changed; enabled a process function.

* disable (since no maintainer yet)
  * Validation
  * MCHistograms
  * EventSelectionQA
  * WeakDecayTutorial

### DIFF
--- a/MC/analysis_testing/analysis_test.sh
+++ b/MC/analysis_testing/analysis_test.sh
@@ -10,7 +10,7 @@
 # (using the "needs" variable and doing a "merge" operation with the original workflow)
 
 # to be eventually given externally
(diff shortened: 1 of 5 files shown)


benedikt-voelkel:
@chiarazampolli @catalinristea @gconesab I think we should be good here.
Also on alibi we are testing exactly the same analyses.

gconesab (approved):

PR #941, opened 2023-03-13
Recognise fair::FatalException

benedikt-voelkel:


benedikt-voelkel:
@sawenzel This is important as otherwise some errors are not recognised from the `o2sim_serverlog`

sawenzel (approved):

PR #943, opened 2023-03-14
workaround: wf won't start on the grid otherwise

chiarazampolli:
@catalinristea


chiarazampolli:
Merging, not tested anyway in the CI.

PR #944, opened 2023-03-14
Add test for AnalysisQC workflow construction

benedikt-voelkel:
* tests whether the AnalysisQC workflow can be created. would fail if
  * python script is corrupted
  * any config JSONs are broken


benedikt-voelkel:
@chiarazampolli this prevents us from accidentally breaking at least the principal functionality of the AnalysisQC

chiarazampolli (approved):

PR #945, opened 2023-03-14
Add ITSTrackerCA config dof+conf for PbPb

mconcas:


mconcas:
Weird lines in the (passed) `
build/O2DPG/sim/o2` check:

```6-o2dpg_sim_workflow: Error! embedding background generator name not provided
7:[FATAL]: O2DPG_TEST Workflow creation failed

[0;32mAll required workflow tests successful[0m```

As I touched something related to embedding I worried about it. Is it genuine? @sawenzel

File: MC/bin/o2dpg_sim_config.py [hunk 1]
```diff
@@ -16,6 +12,12 @@ def add(cfg, flatconfig):
            d[sk] = flatconfig[entry]
            cfg[mk] = d
 
+    COLTYPEIR = args.col
+    if args.embedding==True:
+        COLTYPEIR = args.colBkg
+        add(config, {"ITSCATrackerParam.trackletsPerClusterLimit": 20,
+                     "ITSCATrackerParam.cellsPerClusterLimit": "20"})
+
```
sawenzel, line 20: I don't fully understand this. Is this setting something specific to **all** embedding MCs? Even when we embedd pp into pp? Or is it rather to be applied whenever we have PbPb collisions (irrespective of embedding mode)?

mconcas: > I don't fully understand this. Is this setting something specific to **all** embedding MCs? Even when we embedd pp into pp? Or is it rather to be applied whenever we have PbPb collisions (irrespective of embedding mode)?

The latter you said

mconcas: in pp-pp embedding I don't expect you to need this

benedikt-voelkel:
> Weird lines in the (passed) ` build/O2DPG/sim/o2` check:
> 
> ```
> 7:[FATAL]: O2DPG_TEST Workflow creation failed
> 
> �[0;32mAll required workflow tests successful�[0m
> ```
> 
> As I touched something related to embedding I worried about it. Is it genuine? @sawenzel @chiarazampolli

No, that is fine.
What has failed was the workflow creation of a PWG. I should make that more clear. But it also says "All **required** workflow tests passed."
Sorry for the confusion

MC/bin/o2dpg_sim_config.py
benedikt-voelkel: but it wouldn't do any harm, meaning the settings can be applied in embedding regardless of the collision system?

mconcas: These settings will eventually influence the resources made available for the reconstruction.
If thresholds are exceeded, you have the tracker skipping the TF.
In the pp case, I would just assume this would just allow for larger combinatorics, which in principle, is just an overestimation.

benedikt-voelkel: If this is a decision per TF, shouldn't we then set the additional confKey whenever the signal or the background system is PbPb? And not make it depend on embedding?

mconcas: That would cover the cases correctly, yes.

benedikt-voelkel (approved):

sawenzel (approved):

PR #950, opened 2023-03-16
PWG-LF Resonance injector for pp

BongHwi:
Developed with @njacazio

The first commit of the MC injector of Resonance particles in the pp collision system.
Injected resonances:

- K*
- K*pm
- phi
- f_0(980)
- rho(770)
- rho(770)pm
- Sigma(1385)+
- Sigma(1385)-
- Xi(1530)0

Removed for the moment due to the decay table:
- f_0(1500)
- f_0(1710)
- Xi(1820)0
- Xi(1820)-0


gconesab:
Dear @BongHwi ,
please have a look to the presentation of Benedikt on Monday analysis meeting and check if you need to adapt your commit.
Cheers ... Gustavo

benedikt-voelkel:
Thanks a lot @BongHwi 
Let me know if you have any questions

MC/config/PWGLF/ini/tests/GeneratorLF_Resonances_pp.C
```diff
@@ -0,0 +1,138 @@
...
+                std::cerr << "Number of generated" << decayDaughters[i][j] << "lower than expected\n";
+                return 1;
+            }
+        }
+    }
+
+    return 0;
+}
+
+int Pythia8()
```
benedikt-voelkel, line 133: Thanks a lot for implementing the test macro.
However, if you do not run Pythia8 with the configuration `ini`, could you please remove the Pythia8 function from the macro?
There must be at least one test, but on the other hand, there should not be any trivial test.

Once G4 is there with the correct O_c lifetime, we can move forward and implement an actual test.

benedikt-voelkel (changes requested):

MC/config/PWGLF/ini/tests/GeneratorLF_Resonances_pp.C
BongHwi: Hi @benedikt-voelkel ,

Let me finalise meanwhile.
I found that the test is not optimised since we have a number of decay modes of injected particles.. 
that means we have less than 1 decay daughter pair per injected particle.
I'm thinking about the proper solution to this.

Cheers,
Bong-Hwi

BongHwi: Hi @benedikt-voelkel ,

I have an error to get the MC event tracks.
It shows the error below:
```
terminate called after throwing an instance of 'std::out_of_range'
what():  vector::_M_range_check: __n (which is 18446744073709551615) >= this->size() (which is 66)
```

This happens when we get the MC tracks:
```
std::vector<o2::MCTrack> *tracks{};
tree->SetBranchAddress("MCTrack", &tracks);
```

This was found in this file:
`o2dpg_tests/generators/1_GeneratorLF_Resonances_pp.ini_External_dir/o2dpg-test-kine.log`

Cheers,
Bong-Hwi

benedikt-voelkel: Hello @BongHwi ,
thanks for the implementation and your feedback. I will test your PR locally as well today and will let you know.
It seems weird that It happens when the branch is connected to the variable as you are saying. I will have a look.

benedikt-voelkel: Coming back to this:
Since the buffer is only flushed when the problem occurs, it looks like it arises at the point you were mentioning. However, it is appears to be here: https://github.com/AliceO2Group/O2DPG/pull/950/files#diff-d91b7abbff10fdfcbd3e6dcf0bf9475dd536a4f081a2180312888c15bf9d8c37R109

I can see a case where the daughters are `-1` and `-1` for instance

BongHwi: I see... 
I thought I removed all non-decay injections.
I'll check once again.

benedikt-voelkel: ok, let me know once you tested locally. Then, on your next push, I will approve so that we will have the CI running here

BongHwi: I found that most injected particles have a -1 -1 decay output. 
Is it related to my injection configuration?

benedikt-voelkel: Note that only the generators are tested and there is no transport. If your particles are expected to be decayed during transport, that won't be reflected here.
Could that be the issue?
If so, you should only test those which are supposed to be decayed by Pythia8

benedikt-voelkel: I am not sure what exactly the numbers mean in your previous comment.

Anyway: From what I understand, only `3124` is decayed by Pythia. 
Everything else seems to be switched off in https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/common/pythia8/decayer/base.cfg

is that correct?

BongHwi: I'm not 100% sure about Lambda1520(`3124`) decay in Pythia8.

What I've done was prepare the decay table in `resonances.cfg` file 
and check the produced AO2D file in the O2Physics task, 
and finally found that the Lambda1520 was decayed to the daughters that I set on the decay table.

Without the custom decay table, the injected Lambda1520 didn't decay.

https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/common/pythia8/decayer/base.cfg
This file is okay for the resonance injection.

benedikt-voelkel: Ok, but again: It means that `3124` is decayed by Pythia8 and everything else mentioned in https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/common/pythia8/decayer/base.cfg is explicitly inhibited from the decay in Pythia8. So to my understanding, that is expected to be decayed during transport.

That means, that **only** for `3124` you would find decay products during your test macro because we only do generation but **no transport**.

What I meant earlier "I am not sure what exactly the numbers mean in your previous comment." I was referring to 
`Daughter -2212: 0 <- # of decay in this mode`, what does the `0` mean?

BongHwi: Hello @benedikt-voelkel ,

Together with @njacazio, we updated the injector to add the option to use the Pythia decayer and now they are decayed properly.

And we changed the test script logic so it returns an error only when we can't find the injected particle.

Please take a look and let us know if we need to update further.
Cheers,
Bong-Hwi

benedikt-voelkel (approved):
Thanks a lot for all the work. Approving to run the CI now

benedikt-voelkel:
The CI for MCs failed. That is due to the fact that the file MC/config/PWGLF/ini/GeneratorLFDeTrHe_pp.ini was tested but there is no test script.

The reason **why** that INI file was tested is because the macro [MC/config/PWGLF/pythia8/generator_pythia8_longlived_multiple.C](https://github.com/AliceO2Group/O2DPG/pull/950#diff-e7ccbf49925c183d3dfdedf1e11dab12e5abf06b7f24c7ccd5e9725adaefb256) has changed and it is used in the other INI file as well.

I would say, we can merge this PR in the end but it would be a great idea if you (your group) could follow up and add a test for the INI file that failed @BongHwi Would that be possible?

BongHwi:
> The CI for MCs failed. That is due to the fact that the file MC/config/PWGLF/ini/GeneratorLFDeTrHe_pp.ini was tested but there is no test script.
> 
> The reason **why** that INI file was tested is because the macro [MC/config/PWGLF/pythia8/generator_pythia8_longlived_multiple.C](https://github.com/AliceO2Group/O2DPG/pull/950#diff-e7ccbf49925c183d3dfdedf1e11dab12e5abf06b7f24c7ccd5e9725adaefb256) has changed and it is used in the other INI file as well.
> 
> I would say, we can merge this PR in the end but it would be a great idea if you (your group) could follow up and add a test for the INI file that failed @BongHwi Would that be possible?

Sure, let me prepare the test script for that injection script.

PR #952, opened 2023-03-16
Removing metrics by default; possibility to not do the AOD merging th…

chiarazampolli:
…at optmizes the DFs

@catalinristea : now the metrics is disabled by default to save time. One can still enable it via JDL env var `ALIEN_JDL_ENABLEMONITORING`.
In addition, the "fake-merging" that we do on each AOD file to improve the DFs handling can now be disabled via `ALIEN_JDL_DONOTMERGEAODS=1`

@davidrohr


chiarazampolli:
Merged, not tested anyway CI.

PR #953, opened 2023-03-16
Custom configuration for EventTrackQA.

mfaggin:
Hi @benedikt-voelkel , thanks to your PR #935 I developed these small changes for the `EventTrackQA` analysis. The idea is just to change the `minPropagationDistance` parameter of the `track-prpagation` dependency to `5`. This helps to reduce by a factor ~2 or more the execution time, by keeping the same results. This should help to make easier to run this analysis in the `AnalysisQC`.

Please, find here the commands I used:
 - `python3 ${O2DPG_ROOT}/MC/analysis_testing/o2dpg_analysis_test_workflow.py -f AO2D_LHC22b1a_302018_005.root -a . --only-analyses EventTrackQA --autoset-converters` (coll. converter necessary for the AO2D I used for the test)
 - `python3 ${O2DPG_ROOT}/MC/bin/o2_dpg_workflow_runner.py -f workflow_analysis_test.json`

Here the change in performance before and after the change (values taken from `EventTrackQA/Analysis_EventTrackQA.log_time` created when doing the test):
 - BEFORE
    #walltime 18.96
    #systime 2.85
    #usertime 27.70
    #maxmem 747704
    #CPU 161%
 - AFTER
    #walltime 7.99
    #systime 2.69
    #usertime 11.92
    #maxmem 748684
    #CPU 183%


MC/config/analysis_testing/json/analyses_config.json
```diff
@@ -31,7 +31,7 @@
             "expected_output": ["AnalysisResults.root"],
             "config": {
                 "mc": "json://${O2DPG_ROOT}/MC/config/analysis_testing/json/analysis-testing-mc.json",
-                "data": "json://${O2DPG_ROOT}/MC/config/analysis_testing/json/analysis-testing-data.json"
+                "data": "json://${O2DPG_ROOT}/MC/config/analysis_testing/json/analysis-testing-data-qa-event-track.json"
```
benedikt-voelkel, line 34: Do you only want to set this for data? Sure, often the MC and data configuration can be used for either. But the idea is to have different configs for data and MC in general.

MC/config/analysis_testing/json/analysis-testing-data-qa-event-track.json
```diff
@@ -0,0 +1,105 @@
+{
```
benedikt-voelkel, line 1: could wen call this file maybe
`analysis-testing-EventTrackQA-data.json`
to make the connection between the analysis task and JSON as clear as possible?
(same in case of MC)

benedikt-voelkel (changes requested):

MC/config/analysis_testing/json/analyses_config.json
mfaggin: The configs as they are now are ok for both data and MC in terms of analysis. My change only modifies a dependency to spare resources w/o affecting the final results. Usually in MC this is not a big issue given the lower amount of data. I'd go ahead in this way for the moment.

MC/config/analysis_testing/json/analysis-testing-data-qa-event-track.json
mfaggin: done

benedikt-voelkel (approved):

njacazio:
Ciao @mfaggin could you add the new json to the format script and run it so that future comparison will be easier? Thanks!

mfaggin:
> Ciao @mfaggin could you add the new json to the format script and run it so that future comparison will be easier? Thanks!

ciao @njacazio , done in  commmit [0cfaf9aaa7127337f051a2e2b399ed06e0b15190](https://github.com/AliceO2Group/O2DPG/pull/953/commits/0cfaf9aaa7127337f051a2e2b399ed06e0b15190)

njacazio:
super mattia
thanks!!!

PR #955, opened 2023-03-17
Validate analysis output

benedikt-voelkel:
./o2dpg_analysis_test_config.py validate-output

* check if expected outputs are there for given analyses --> by default check all enabled analyses

* for more options see --help

### DIFF
--- a/MC/analysis_testing/o2dpg_analysis_test_config.py
+++ b/MC/analysis_testing/o2dpg_analysis_test_config.py
@@ -3,7 +3,7 @@
 import sys
 import argparse
 from os import environ
-from os.path import join
+from os.path import join, exists
 import json
 
 # make sure O2DPG + O2 is loaded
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
@catalinristea What do you think?

E.g.
```bash
${O2DPG_ROOT}/MC/analysis_testing/o2dpg_analysis_test_config.py validate-output -d Analysis
```

would check if all expected outputs are there.

catalinristea:
Hi Benedikt, all

PR is now complete - I have tested it and found it very useful for the validation of AnalysisQC stage.
@chiarazampolli, @sawenzel, @gconesab,  

please merge it, such that we can resume this stage in production.

Thanks a lot!

Yours, Catalin.

benedikt-voelkel:
@chiarazampolli I think you can merge if you have the power. The FST will be skipped in any case. At the moment, it seems that something in the machinery is stuck though so it doesn't even recognise that it should be skipped.
Pinging also @TimoWilken

chiarazampolli:
Done!

PR #957, opened 2023-03-20
use default MCH parameters for PbPb

pillot:
Hello @chiarazampolli ,
Since the PbPb apass5 will be done with a recent O2 version (if I understood correctly), it can use the default reconstruction parameters, which have been updated recently (#10927).


chiarazampolli (approved):

chiarazampolli:
Thanks, @pillot !

PR #962, opened 2023-03-22
Parallelize AOD DFs merging

chiarazampolli:
@davidrohr , @jgrosseo , this is to merge in parallel several AODs. 
The number of parallel jobs is either the number of CPUs in the job, or 8. But if passed externally, then the external value will be used. 
The procedure used `jobs` to verify that we submit only at max what we declare, and `wait` to check that the processes finished correctly.


DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
```diff
@@ -589,70 +605,114 @@ if [[ $ALIEN_JDL_AODOFF != 1 ]]; then
...
+      done
+      run_AOD_merging $line &
+      arr[$i]=$!
+      aods[$i]=$line
+      i=$((i+1))
+      CURRENT_POOL_SIZE=`jobs -r | wc -l`
+    done < $JOB_LIST
+    # collecting return codes of the merging processes
+    for i in ${!arr[@]}; do
+      wait ${arr[$i]}
```
jgrosseo, line 632: To my understanding this will wait for the PID you provide. This means if you have for example 16 mergings to do and you have 8 cores, it will submit the first 8. Then you wait for the first one to finish. So even if the second one finishes before, you will not put another job. Is this correct?

chiarazampolli: The injection of the mergings happens in the `while`. The `for` is executed later on. But I can check again.

DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
```diff
@@ -589,70 +605,114 @@ if [[ $ALIEN_JDL_AODOFF != 1 ]]; then
...
+  if [[ -z $ALIEN_JDL_DONOTMERGEAODS ]] || [[ $ALIEN_JDL_DONOTMERGEAODS == 0 ]]; then
+    # spawning the parallel merging
+    timeStartMerge=`date +%s`
+    arr=()
+    aods=()
+    mergedok=()
+    i=0
+    while IFS= read -r line; do
+      while [[ $CURRENT_POOL_SIZE -ge $MAX_POOL_SIZE ]]; do
+	CURRENT_POOL_SIZE=`jobs -r | wc -l`
```
davidrohr, line 622: you should add a sleep here, otherwise we have one processes spinning checking the variable

davidrohr (changes requested):

davidrohr:
> The only thing that I am not sure is working fine is getting the return code from the function. Below I set it to `45` but it prints that it is always `0`, so I am not sure how I can check the return value of the `sleep` (or `aod-merger`).

For getting the return code, I have tried https://unix.stackexchange.com/a/230740, which works for me. Your code looks very similar, and is correct in my opinion. Though I haven't tried it yet.

jgrosseo:
@chiarazampolli maybe not important but if you have
  echo "PID of last sleep is $!"
  exitcode=$?

then exitcode will contain the exit code of echo... (of course it is fine if you comment the line before but it can be confusing when debugging)

jgrosseo:
You could maybe test it by providing an empty input_list.txt. In this case the merger gives a non-0 exit code...

DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
chiarazampolli: Ok, I will. It will decrease the performance, though, no?

chiarazampolli:
> @chiarazampolli maybe not important but if you have echo "PID of last sleep is $!" exitcode=$?
> 
> then exitcode will contain the exit code of echo... (of course it is fine if you comment the line before but it can be confusing when debugging)

It could be. I am sure that I had tested the return code before opening the PR, then yesterday, when I modified my standalone test to post it here, I did not see it working anymore. I will check.

chiarazampolli:
The problem in the test script that I posted is the line:
```
echo "Process ${arr[$i]} finished"
```
then the `$?` takes the exit of this command, instead of wait. I added it to make it clearer for you, but I messed up the logic, sorry. So things work fine, the exit code is taken. 
For safety, I will try with an empty list for the merging. 
And I will add the sleep.

DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
chiarazampolli: `sleep 1` added.

chiarazampolli:
Done, and also tested with an empty list. 
If you want to test, you can try teh script below. The difference with the previous one is practically the suppression of the echo preventing the `$?` to get the return code of `wait`.
```
for dir in 001 002 003 004 005 006 007 008 ; do
  echo $dir
  mkdir $dir
  touch $dir/AO2D.root
done

AOD_LIST_COUNT=`find . -name AO2D.root | wc -w`
AOD_LIST=`find . -name AO2D.root`
MAX_POOL_SIZE=3
JOB_LIST=job-list.txt
if [[ -f $JOB_LIST ]]; then
  rm $JOB_LIST
fi
rm */AO2D_merged.root

run_AOD_merging() {
  sleep 5
  #local latest_PID=$?
  #echo "PID of last sleep is $!"
  exitcode=$?
  #echo "The exit code will be $1"
  #echo $latest_PID
  #return 45
  return $1
  #o2-aod-merger --input list_$1.list --output $1/AO2D_merged.root > $1/merging.log
}

for (( i = 1; i <=$AOD_LIST_COUNT; i++)); do
  AOD_FILE=`echo $AOD_LIST | cut -d' ' -f$i`
  AOD_DIR=`dirname $AOD_FILE | sed -e 's|./||'`
  echo AOD_DIR = $AOD_DIR
  ls $AOD_FILE > list_$AOD_DIR.list
  #echo "list is"
  #cat list_$AOD_DIR.list
  echo "$AOD_DIR" >> $JOB_LIST
done

_log() {
        echo " $(date +'[%F %T]') - $1"
}

T1=$(date +%s)
arr=()
aods=()
mergedok=()
i=0
while IFS= read -r line; do
  while [[ $CURRENT_POOL_SIZE -ge $MAX_POOL_SIZE ]]; do
#    _log "Pool is full, with $CURRENT_POOL_SIZE jobs. waiting for jobs to exit..."
    CURRENT_POOL_SIZE=`jobs -r | wc -l`
  done
  #  _log "Current pool size before = $CURRENT_POOL_SIZE"
  #  _log "Starting job $line"
  run_AOD_merging $line &
  #echo "PID of latest sleep = " $!
  #jobs -p 
  arr[$i]=$!
  aods[$i]=$line
  echo "PID of latest bkg job = " ${arr[$i]} # this is the PID of the function
  echo "We have started " ${aods[$i]}
  i=$((i+1))
  CURRENT_POOL_SIZE=`jobs -r | wc -l`
done < $JOB_LIST

echo "We are done with STARTING all processes, now we check the exit code of each"

for i in ${!arr[@]}; do
  wait ${arr[$i]}
#  echo "Process ${arr[$i]} finished"
  exitcode=$?
  echo PID ${arr[$i]} terminated with exit code $exitcode
  # the test on value "5" is just to try
  if [[ $exitcode != "5" ]]; then
    echo "merging not ok for ${aods[$i]}"
    mergedok[$((10#${aods[$i]}))]=0
  else
    mergedok[$((10#${aods[$i]}))]=1
    echo "All ok for ${aods[$i]}"
  fi  
done

for (( i = 1; i <=$AOD_LIST_COUNT; i++)); do
  AOD_FILE=`echo $AOD_LIST | cut -d' ' -f$i`
  AOD_DIR=`dirname $AOD_FILE | sed -e 's|./||'`
  echo "checking ${mergedok[$((10#$AOD_DIR))]} for $AOD_DIR"
  if [[ ${mergedok[$((10#$AOD_DIR))]} == 0 ]]; then
    echo "we cannot rm this AOD"
  else
    echo "we can remove this AOD"
  fi
done

T2=$(date +%s)

_log "All jobs completed in $((T2-T1)) seconds. Parent process exiting."
_log "Final output is written in $OUTPUT"
```

jgrosseo:
I see!
I would still replace `#echo "PID of last sleep is $!"` by `#echo "PID of last sleep is $exitcode" to avoid similar issues in the future...

DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh
jgrosseo: > Ok, I will. It will decrease the performance, though, no?

It will increase it. Without the sleep, one cpu is busy with the script...

chiarazampolli:
> similar

But this is a test script…

Do you (@jgrosseo , @davidrohr ) think that the PR be merged?

davidrohr:
Fine with me to merge
If you want me to test it beforehand on an EPN I can also do so, but I assume you tested locally already.

chiarazampolli:
I did test it, yes. For me it works, and I got a speed up factor or 3 when merging 5 files, with 3 as parallel threads. 
If you could merge it in a relatively short time, it could be useful, otherwise I would merge it, since I have several pending modifications to the scripts.

PR #964, opened 2023-03-22
Do not add extra errors in 22q if requested to not do so

chiarazampolli:
@shahor02


shahor02 (approved):

chiarazampolli:
Thanks! 
Merging since it is not tested.

PR #966, opened 2023-03-23
Adding wither throttling, or multiplicities

chiarazampolli:
@shahor02 , is this what you mean in https://alice.its.cern.ch/jira/browse/O2-3565?focusedCommentId=298580&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-298580?
@catalinristea , FYI


DATA/production/configurations/CTFSkimming/ctf-skim-workflow.sh
```diff
@@ -77,6 +77,13 @@ has_detector_ctf CTP && add_W o2-ctp-entropy-encoder-workflow "--select-ir-frame
 
 add_W o2-ctf-writer-workflow "--output-dir $CTF_DIR --min-file-size ${CTF_MINSIZE} --max-ctf-per-file ${CTF_MAX_PER_FILE} --onlyDet $WORKFLOW_DETECTORS_CTF $CTF_MAXDETEXT"
 
+if [[ $ALIEN_JDL_CPUCORES == 8 ]]; then
+  export MULTIPLICITY_PROCESS_tpc_entropy_decoder=2
+  MULTIPLICITY_PROCESS_tpc_entropy_encoder=2
+else
+  export TIMEFRAME_RATE_LIMIT=2
+fi
+
```
shahor02, line 86: No, the whole setting should go before the workflow composition, e.g. to https://github.com/AliceO2Group/O2DPG/blob/da75bf959363d9f944a144c72d7acdbe940460a7/DATA/production/configurations/CTFSkimming/ctf-skim-workflow.sh#L14 and the TIMEFRAME_RATE_LIMIT should be always set. I.e.
```
TIMEFRAME_RATE_LIMIT=2
if [[ $ALIEN_JDL_CPUCORES == 8 ]]; then
  export MULTIPLICITY_PROCESS_tpc_entropy_decoder=2
  export MULTIPLICITY_PROCESS_tpc_entropy_encoder=2
  export TIMEFRAME_RATE_LIMIT=3
fi
```
What I am not sure is who sets the NUMAID for the throttling. I remember on the grid it is set as a job ID to avoid conflicts.
Perhaps @catalinristea knows?

chiarazampolli: Ah yes, sorry, the settings go above, I was too quick. 
Why do you need the NUMAID?

catalinristea: Ciao,

I'm only aware of OPTIMIZED_PARALLEL_ASYNC=pp_4gpu, which I think does the switch to full NUMA domain. But probably this isn't what you need.

Yours, Catalin.

chiarazampolli: I moved the lines, can you check now?

shahor02: Hi,
NUMAID is added as an argument of the ` --timeframes-rate-limit-ipcid`, apparently it is converted to grid job ID internally.
I've added some fixes.
Cheers,
 Ruben

shahor02: Hi,
No, since we add the option, we must provide also the value. Ok, let's use 0 as I've added it until it is proven that it is wrong.

davidrohr:
@chiarazampolli @shahor02 : We should not set NUMAID externally to separate the jobs. This can lead to bogus GPU settings. Apparently this was used so far to make sure channels for metric-feedback of different jobs were seperate, but I think it just worked by chance.

I have created https://github.com/AliceO2Group/O2DPG/pull/969/files and https://github.com/AliceO2Group/AliceO2/pull/11005.
Could we change the async scripts to set O2JOBID instead, hope that works to keep the channels separate without messing with NUMAID.

shahor02:
@davidrohr ok, I guess it can be set from the grid jobID, but in this case the script was starting CPU only workflow. What I am not sure how the workflows on the grid work now with the same `--timeframes-rate-limit-ipcid 0` for potentially multiple instances on the same node.

davidrohr:
> @davidrohr ok, I guess it can be set from the grid jobID, but in this case the script was starting CPU only workflow. What I am not sure how the workflows on the grid work now with the same `--timeframes-rate-limit-ipcid 0` for potentially multiple instances on the same node.

It won't work. That is why in dpl-workflow with the patch I am setting `--timeframes-rate-limit-ipcid $(($NUMAID + 10 * ${O2JOBID:-0}))`
So if O2JOBID is set, it will use different ipcid per job

shahor02:
It actually works with `--timeframes-rate-limit-ipcid 0` because the final channel name is composed as https://github.com/AliceO2Group/AliceO2/blob/6a82f93015c9cb20e6a98a8572b3cf7849fe0b64/Framework/Core/src/WorkflowHelpers.cxx#L311-L314 (you did this, right?) and on the grid the `ChannelSpecHelpers::defaultIPCFolder()` is augmented by the grid jobID: 
https://github.com/AliceO2Group/AliceO2/blob/6a82f93015c9cb20e6a98a8572b3cf7849fe0b64/Framework/Core/src/WorkflowCustomizationHelpers.cxx#L23.

But I agree, an extra diversification in the  `--timeframes-rate-limit-ipcid` will not hurt.

davidrohr:
> on the grid the `ChannelSpecHelpers::defaultIPCFolder()` is augmented by the grid jobID:

I think this was done by Giulio. But I think in any case this diversification is good indeed, so one can also manually run 2 chains.
Anyhow, the GRID job should never set / change NUMAID. Can you confirm this is not the case?

shahor02:
If you mean setting it to the grid job I'd, no, it is never done.
This pr sets NUMAID to 0 in the skimming script to pass it for the IPCID (dpl-workflow.sh is not used here).

davidrohr:
OK, then I propose to change that to O2JOBID, and use the same syntax as I added in https://github.com/AliceO2Group/AliceO2/pull/11005 for the IPCID (assuming my PR goes through, didn't test locally)

shahor02:
https://github.com/AliceO2Group/O2DPG/pull/970

PR #967, opened 2023-03-23
Add Skimming for TPC to analysis QC

njacazio:


njacazio:
@noferini @jezwilkinson this is my suggestion

jezwilkinson:
Hi @njacazio - thanks very much, this looks good.
Just one question, does the configuration JSON become hardcoded here or can it be customised? We may need to still tweak the downsampling factors depending on the size of the datasets

njacazio:
Hi @jezwilkinson the JSON is hardcoded, one thing could be to have the downsampling fetched via CCDB so that it automagically matches the correct data. But this requires some dev.
I let @benedikt-voelkel comment on the best approach

benedikt-voelkel:
Hej @njacazio 
Do you have an estimation how long this task takes on - say - 1,000 pp events? Just so that we keep track of what we have compared to MC or reco.

MC/config/analysis_testing/json/analyses_config.json
```diff
@@ -96,6 +96,34 @@
...
         },
+        {
+          "name": "TPCSkimming",
+          "enabled": true,
+          "expected_output": [
+            "AnalysisResults.root",
+            "AnalysisResults_trees.root"
+          ],
+          "config": {
+            "data": "json://${O2DPG_ROOT}/MC/config/analysis_testing/json/analysis-skimming.json",
```
benedikt-voelkel, line 107: I suggest to rename to `analysis-testing-TPCSkimming.json` to align with the other configs we have

benedikt-voelkel (changes requested):

jezwilkinson:
> Do you have an estimation how long this task takes on - say - 1,000 pp events? Just so that we keep track of what we have compared to MC or reco.

@benedikt-voelkel  I don't have a concrete estimate for time/1k events, but the running statistics can be found in the previous Hyperloop runs for the wagon (e.g. https://alimonitor.cern.ch/hyperloop/train-run/70232, https://alimonitor.cern.ch/hyperloop/train-run/70257) if that helps

benedikt-voelkel:
Thanks!
ok, let's do it line this:
@njacazio could you set the `enabled` flag in the configuration to `false` for now?
Then - as soon as you have tested that it works in principle - we can merge this PR.
After that, I will test it on some of our nightlies to have an idea what the timing is roughly.

njacazio:
Hi @benedikt-voelkel done!

benedikt-voelkel (approved):

njacazio:
Hi @benedikt-voelkel thanks a lot for the prompt action! Then we wait for you and the alibi test before going forward

PR #970, opened 2023-03-24
use O2JOBID in the skimming workflow timeframes-rate-limit-ipcid

shahor02:


chiarazampolli:
Hello @shahor02 , 
I don't understand how this should work: who passes "O2JOBID" to the job? I don't see it in the various PRs about it:
https://github.com/AliceO2Group/AliceO2/pull/11005/files
https://github.com/AliceO2Group/O2DPG/pull/969
Does it have to come from the job?
Chiara

shahor02:
At the moment O2JOBID is not set centrally, so effectively it will be set to 0 and have no effect. But it can be set as external env.var. or from the steering script according e.g. ALIEN_PROC_ID.

chiarazampolli:
Ok, I will add this.
Chiara

PR #972, opened 2023-03-24
Consistently use MeanVertex from CCDB

sawenzel:
* the whole MC workflow will now (by default) use the GLO/Calib/MeanVertex object from CCDB

* For expert mode / debugging, it is still possible to use the `-confKey "Diamond.width[2]=..."` options to set vertex properties. In this case, a local MeanVertex object will be created in the CCDB cache (instead of being fetched from the server) so that all sub-sequent tasks pick this up automatically.


sawenzel:
Needs https://github.com/AliceO2Group/AliceO2/pull/11009 before merging

PR #977, opened 2023-03-28
Do not get B field from external file anymore, but from CCDB - O2DPG

chiarazampolli:
@knopers8 , @shahor02 , goes with 
https://github.com/AliceO2Group/AliceO2/pull/11044
https://github.com/AliceO2Group/QualityControl/pull/1692

### DIFF
--- a/DATA/production/qc-async/mft.json
+++ b/DATA/production/qc-async/mft.json
@@ -39,8 +39,17 @@
           "maxClusterROFSize" : "50000",
           "maxDuration" : "60000",
           "timeBinSize" : "0.1",
-          "ROFLengthInBC" : "594",
-          "geomFileName" : "o2sim_geometry-aligned.root"
+          "ROFLengthInBC" : "594"
+        },
+	"grpGeomRequest" : {
+          "geomRequest": "Aligned",
+          "askGRPECS": "false",
+          "askGRPLHCIF": "false",
+          "askGRPMagField": "true",
+          "askMatLUT": "false",
+          "askTime": "false",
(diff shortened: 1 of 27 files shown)


tomas-herman:
I think it should be fine but I am wondering about the json files for MC and the json files used for tests of PR builds (located in the sync QC json folder). They should be also updated, right? (and as you mentioned also in Consul once the PRs are deployed)

chiarazampolli:
Hello @tomas-herman , 
Yes, indeed, also the json in the sync folder need to be changed. I will take care.
Chiara

DATA/production/qc-async/mft.json
```diff
@@ -39,8 +39,17 @@
...
+        },
+	"grpGeomRequest" : {
+          "geomRequest": "Aligned",
+          "askGRPECS": "false",
+          "askGRPLHCIF": "false",
+          "askGRPMagField": "true",
+          "askMatLUT": "false",
+          "askTime": "false",
+          "askOnceAllButField": "false",
+          "needPropagatorD":  "false"
```
tomas-herman, line 52: I assume that the Geometry Request which is here, should be added also to:
- https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/qc-sync/mft-full.json
- https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/qc-sync/mft-full-no-tracks.json
- https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/QC/json/mft-clusters.json
I will try to do a local test but I need to rebuild as I have old versions of the SW,

PR #979, opened 2023-03-29
ITSreco: Do not use 0.f in config values - just 0.

sawenzel:
Fixing an error message "Error in setValue (string) bad lexical cast: source type value could not be interpreted as target" seen in PbPb MC jobs. Most likely the cut setting was then ignored and did not have the targeted effect.


sawenzel:
@mconcas : Maybe this already helps with embedding_G4

sawenzel:
I should improve the string parsing in config-Values or at least do a check that there is not trailing "f" behind numbers....

PR #980, opened 2023-03-29
Adding some statistics after processing

chiarazampolli:
@costing @catalinristea 
This will produce an empty file named as follows:

${nCTFsFilesInspected}_${nCTFsFilesOK}_${nCTFsProcessed}_${PVs}_$1_stat.txt

where:
nCTFsFilesInspected = number of CTFs files for which "alien_cp" was called; this is 0 in case a local file(s) is (are) used;
nCTFsFilesOK = number of CTFs files that were successfully processed
nCTFsProcessed = number of single CTFs that were processed (not files, but CTFs!)
PVs = number of reconstructed prim vertices (can be interesting also for @shahor02 )
$1 = argument passed: it is the log file that is parsed to obtain the above. We need it since in case the wf is split, CTFs are read 2x, so I cannot use stdout. 

E.g., from a local test:

0_1_4_27792_latest_reco_1.log_stat.txt

I use these numbers also in case of split wf to compare that the first step where we read CTFs for the first time reads the same number of files/ctfs as the second (of course, it could still be that the files are different, but let's hope not).

@noferini , @davidrohr , FYI.

@catalinristea : you need to add "*stat.txt" to the output file list in the JDLs to have these.

### DIFF
--- a/DATA/production/common/getStat.sh
+++ b/DATA/production/common/getStat.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# script to extract statistics from job
+
+PVs=0
+CTFsFilesProcessedMsg=$(grep "CTFReader stops processing," $1)
+nCTFsFilesOK=`echo $CTFsFilesProcessedMsg | sed 's/.*processing, \([0-9]*\) .*/\1/'`
+nCTFsFilesInspected=`grep alien_cp $1 | grep -v ALARM | grep -v non-zero | wc -l`
+nCTFsProcessed=`grep "Read CTF" $1 | tail -1 | sed 's/^.*Read CTF \([0-9]*\).*$/\1/'`
+while read -r line; do
+  currentPVs=`echo $line | sed 's/^.*Found \([0-9]*\) PVs.*/\1/'`
+  PVs=$((PVs + currentPVs))
+done < <(grep "Found" $1 | grep "PVs")
+
+echo CTFsFilesProcessedMsg = $CTFsFilesProcessedMsg
+echo nCTFsFilesInspected = $nCTFsFilesInspected
+echo nCTFsFilesOK = $nCTFsFilesOK
+echo nCTFsProcessed = $nCTFsProcessed
+echo PVs = $PVs
+
+touch ${nCTFsFilesInspected}_${nCTFsFilesOK}_${nCTFsProcessed}_${PVs}_$1_stat.txt
+
--- a/DATA/production/common/getStat.sh
+++ b/DATA/production/common/getStat.sh
@@ -1,22 +1,36 @@
 #!/bin/bash
 
 # script to extract statistics from job
+# $1 is the file to parse
+# $2 is the (possible) extension
 
 PVs=0
 CTFsFilesProcessedMsg=$(grep "CTFReader stops processing," $1)
 nCTFsFilesOK=`echo $CTFsFilesProcessedMsg | sed 's/.*processing, \([0-9]*\) .*/\1/'`
 nCTFsFilesInspected=`grep alien_cp $1 | grep -v ALARM | grep -v non-zero | wc -l`
+nCTFsFilesFailed=`grep "FileFetcher: non-zero exit code [0-9]*" $1 | wc -l`
 nCTFsProcessed=`grep "Read CTF" $1 | tail -1 | sed 's/^.*Read CTF \([0-9]*\).*$/\1/'`
-while read -r line; do
-  currentPVs=`echo $line | sed 's/^.*Found \([0-9]*\) PVs.*/\1/'`
-  PVs=$((PVs + currentPVs))
-done < <(grep "Found" $1 | grep "PVs")
+nCTFsProcessed=$((nCTFsProcessed + 1))
+
+if [[ $nCTFsFilesInspected != $((nCTFsFilesFailed + nCTFsFilesOK)) ]]; then
+  echo "Something went wrong with parsing the log file: CTF files inspected ($nCTFsFilesInspected) is not the sum of those successfully processed ($nCTFsFilesOK) and those that failed ($nCTFsFilesFailed)"
+else
+  while read -r line; do
+    currentPVs=`echo $line | sed 's/^.*Found \([0-9]*\) PVs.*/\1/'`
+    PVs=$((PVs + currentPVs))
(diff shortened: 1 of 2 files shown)


DATA/production/common/getStat.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# script to extract statistics from job
+
+PVs=0
+CTFsFilesProcessedMsg=$(grep "CTFReader stops processing," $1)
+nCTFsFilesOK=`echo $CTFsFilesProcessedMsg | sed 's/.*processing, \([0-9]*\) .*/\1/'`
+nCTFsFilesInspected=`grep alien_cp $1 | grep -v ALARM | grep -v non-zero | wc -l`
+nCTFsProcessed=`grep "Read CTF" $1 | tail -1 | sed 's/^.*Read CTF \([0-9]*\).*$/\1/'`
+while read -r line; do
+  currentPVs=`echo $line | sed 's/^.*Found \([0-9]*\) PVs.*/\1/'`
```
shahor02, line 11: Would not this work only for the lines which contain the vertexer report? I would add at least
```
currentPVs=`echo $lll  | grep -P "Found ([0-9]*) PVs" | sed 's/^.*Found \([0-9]*\) PVs.*/\1/'`
```

chiarazampolli:
> Ciao Chiara,
> 
> The accounting code is looking for file name patterns like `A_B_C_D.stat` where: `A` : total input (events in the past ; files here) `B` : passed (filtered ; files copied ok here) `C` : errors (I think this was always zero in analysis; files failed to be copied now) `D` : output (events ; files correctly processed here)
> 
> So normally `A=B+C` and `D=B` for CTF files. And having for example `D<B` would show a crash in reconstruction, not related to data access.
> 
> I could extend this to ignore the new value, PV, but could you please arrange the first 4 tokens to match the above? And then the PV could become a fifth token.
> 
> As file extensions, default is to look for `.stat` but also consider others (in this order), in case several steps run in the same job: `".stat", ".stat.rec", ".stat.qa", ".stat.qa_barrel", ".stat.aod", ".stat.aod_barrel", ".stat.barrel"`
> 
> Cheers,
> 
> .costin

Hello Costin, 

I will take a look on Friday, since tomorrow I will be offline.

Chiara

DATA/production/common/getStat.sh
chiarazampolli: I am not sure what you mean: the while loops over the result of the grep, a bit below.

shahor02: sorry, missed the grep below

chiarazampolli:
> Ciao Chiara,
> 
> The accounting code is looking for file name patterns like `A_B_C_D.stat` where: `A` : total input (events in the past ; files here) `B` : passed (filtered ; files copied ok here) `C` : errors (I think this was always zero in analysis; files failed to be copied now) `D` : output (events ; files correctly processed here)
> 
> So normally `A=B+C` and `D=B` for CTF files. And having for example `D<B` would show a crash in reconstruction, not related to data access.
> 
> I could extend this to ignore the new value, PV, but could you please arrange the first 4 tokens to match the above? And then the PV could become a fifth token.
> 
> As file extensions, default is to look for `.stat` but also consider others (in this order), in case several steps run in the same job: `".stat", ".stat.rec", ".stat.qa", ".stat.qa_barrel", ".stat.aod", ".stat.aod_barrel", ".stat.barrel"`
> 
> Cheers,
> 
> .costin

Hello @costing , 

While I can do so that A, B, C match what you wrote (even though it sounds redundant to have both the number of successful and failing alien_cp), I don't see any use in doing D = B. 
What is instead interesting is to see how many single CTFs we have processed (there is a bug since I should add 1 to the counting), which cannot be = B, since 1 file contains more CTFs. 
The extension that you propose do not work for several steps, but anyway, I will use internally different extension and then move one single file to ".stat" (if there is inconsistency between 2 steps, the job should fail).
So I would do (the underscores were present also before):

${nCTFsFilesInspected}_${nCTFsFilesFailed}_${nCTFsFilesOK}_${nCTFsProcessed}_${PVs}.stat

where:
nCTFsFilesInspected = number of CTFs files for which "alien_cp" was called; this is 0 in case a local file(s) is (are) used;
nCTFsFilesFailed = number of CTFs files for which "alien_cp" failed; this is 0 in case a local file(s) is (are) used;
nCTFsFilesOK = number of CTFs files that were successfully processed
nCTFsProcessed = number of single CTFs that were processed (not files, but CTFs!)
PVs = number of reconstructed prim vertices (can be interesting also for @shahor02 )

I understand that this might not match Run 2 format, but data have changed... 

Cheers,

Chiara

chiarazampolli:
Hello @costing , 
Please, check how it is done now. There are 5 parameters, like I described above:

${nCTFsFilesInspected}_${nCTFsFilesFailed}_${nCTFsFilesOK}_${nCTFsProcessed}_${PVs}.stat

where:
nCTFsFilesInspected = number of CTFs files for which "alien_cp" was called; this is 0 in case a local file(s) is (are) used;
nCTFsFilesFailed = number of CTFs files for which "alien_cp" failed; this is 0 in case a local file(s) is (are) used;
nCTFsFilesOK = number of CTFs files that were successfully processed
nCTFsProcessed = number of single CTFs that were processed (not files, but CTFs!)
PVs = number of reconstructed prim vertices (can be interesting also for @shahor02 )

Chiara

costing:
Ciao Chiara,

For compatibility purposes the meaning of the first 4 tokens must be similar to the past one. Adding more fields is ok but shuffling order is not. That would just mean swapping two fields, like in : `${nCTFsFilesInspected}_${nCTFsFilesOK}_${nCTFsFilesFailed}_${nCTFsProcessed}_${PVs}.stat`

Cheers,

.costin

chiarazampolli:
Ah, sorry, my bad, when changing the code, I was sure you wanted first the errors :-) I will swap.

chiarazampolli:
@costing , like this? As said parameter 4 will have a slightly different meaning than in the past.

costing:
Looks good to me. Clearly the meaning is different but for as long as they are in the same unit and "output" / "input" is still relevant as a success percentage (per each job and summed up across jobs), it shouldn't matter.

davidrohr (approved):

PR #985, opened 2023-03-31
Disabling a few settings not needed for MC

chiarazampolli:
@shahor02 : following your comment in the discussion about MC anchoring to data:
```
tuned time offsets, i.e. TriggerOffsetsParam.customOffset... and GPU_rec_tpc.clustersShiftTimebins.. settings used in  the setenv_extra.sh are not needed.
TPC ad hoc corrections (debiasing) is not needed, i.e. trackTuneParams.useTPCInnerCorr=true;trackTuneParams.tpcParInner[....] should be omitted, but the cov.matrix modifications
trackTuneParams.sourceLevelTPC=true;trackTuneParams.tpcCovInnerType=1;trackTuneParams.tpcCovInner[..] should be preserved for pp. In pbpb apasses we did not used them so they are not needed.
```
For PbPb, I did nothing so far because we will anchor to a new pass which we are preparing. This period is anyway not in the `if's` for the debiasing + extra TPC cov matrix, for now, so neither data nor MC will have these settings. 
This makes me wonder: should we add something for apass5? 
Thanks!

### DIFF
--- a/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
+++ b/DATA/production/configurations/2022/LHC22f/apass1/setenv_extra.sh
@@ -91,7 +91,9 @@ if [[ $PERIOD == "LHC22s" ]]; then
     ZDC_BC_SHIFT=0
   fi
   CTP_BC_SHIFT=-293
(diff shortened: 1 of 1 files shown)


shahor02:
Hi @chiarazampolli 
Looks fine. For the PbPb, given that the IR is very low, the debiasing is certainly not needed. We could expand the cov. matrix by the same amount as in LHC22q, but it was probably an overkill also there. I can try to retune the cov. matix modification for the low rate (on 22s).

chiarazampolli:
Hello @shahor02 , 
Thanks for the feedback, then I merge this one now. 
Chiara

chiarazampolli:
Not tested in CI, merged.

PR #986, opened 2023-03-31
Fix extra args and config + TRD CTP offset for 22c, e, d, f

chiarazampolli:
@shahor02 , I added the option `--correct-trd-trigger-offset` for 22f, but also the Pilot Beam. 
@lietava, can you confirm that it is needed also for the Pilot beam (22c, d, e)? 
Since we need this urgently for 22f, I would merge this PR soon, if no feedback for 22c, d, e comes and it is wrong, we can fix it later.
@davidrohr , I fixed the PR where I was appending new options too.


chiarazampolli:
Not testes in CI, merged.

chiarazampolli:
Hello @shahor02 , this option for TRD is anyway needed only for MC right? If it is like this, I need to protect it... 

Chiara

shahor02:
Actually, it is needed only for the raw data before 22m. So yes, the MC needs to be protected from it.

PR #988, opened 2023-04-01
Async pass: Support list input not as list.list from current folder

davidrohr:
@chiarazampolli : This supports other file list files with `.list` extension, also from other folders, as argument.
Was kind of convenient for me for testing.

Another comment is: shouldn't we move async_pass.sh to a proper directory, not inside `2022/LHC22f/apass1`, since now I think we use this script for everything.


chiarazampolli:
Thanks!
Yes the script is going to be moved. There were many changes on which i worked in the last two weeks, this is lower priority, but i think that after two more next week, it will come next.

PR #989, opened 2023-04-01
Add small unbiasing and TPC cov.mat. tuning for LHC22s

shahor02:
@chiarazampolli as discussed, this is a tuning for 22s: I have added a small unbiasing (apparently the static distortions effect) and TPC cov.matrix inflation, 10% of what we had in pp, so it should not affect strongly the resonances mass resolution. 
In principle, the same settings could be applied to 22q in case we want to reprocess it.


chiarazampolli:
Hello @shahor02 , 
Thanks. If this is better for 22q, then let's use it also there - will you make the change?
What about 22f and PilotBeam? We have not started yet those (22f was done but it needs to be restarted).
Chiara

shahor02:
@chiarazampolli done, note that I've suppressed the ALIEN_JDL_NOEXTRAERR22Q

chiarazampolli:
Thanks, merging!

PR #992, opened 2023-04-03
askOnceAllButField changed to true in the GeomHelper

chiarazampolli:
@shahor02 , @davidrohr , @knopers8

### DIFF
--- a/DATA/production/qc-async/its.json
+++ b/DATA/production/qc-async/its.json
@@ -52,7 +52,7 @@
           "askGRPMagField": "false",
           "askMatLUT": "false",
           "askTime": "false",
-          "askOnceAllButField": "false",
+          "askOnceAllButField": "true",
           "needPropagatorD":  "false"
         }
       },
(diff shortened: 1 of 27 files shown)


shahor02 (approved):

chiarazampolli:
> Thx! Can you check with @martenold to do an fst on the EPNs. Also the detectors must update their configs in consul, so please check with them as well.

Yes, we are updating. Only missing one is MFT (@tomas-herman , I pinged you also via mail)

martenole:
Hi @chiarazampolli 
the FST on the EPNs is running stably with this change. The next software update is planned for Tuesday. Do we need to add the `askOnceAllButField` option to the json files at P2 synchronously again or has that already been done, do you know?

chiarazampolli:
Hello, 
This is an optimization, if not there, it won't be critical, but better to have it. 
For GLO, I already changed it on Monday. ITS also did it (they added a log entry if I am not wrong). Can you check with MFT, and TOF?
Chiara

tomas-herman:
Hi, for MFT I also put this option in the Consul already on Monday.

martenole:
For TOF I still see it disabled. Who should I ping for the TOF QC json files?

chiarazampolli:
I think @noferini can update consul for TOF.

PR #993, opened 2023-04-03
Setting O2JOBID using ALIEN_PROC_ID

chiarazampolli:
@shahor02 , @davidrohr , should it not be like this? Otherwise the change that uses O2JOBID will be ineffective, or not?

### DIFF
--- a/DATA/production/common/setVarsFromALIEN_PROC_ID.sh
+++ b/DATA/production/common/setVarsFromALIEN_PROC_ID.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Script to set a couple of variables that depend on ALIEN_PROC_ID.
+# Since several scripts might need this, it is in a separate script.
+
+# let's get the last 10 (for INT) and 5 (for int16) digits of ALIEN_PROC_ID, to be passed to define NUMAID and shm-segment-id via O2JOBID, which are int and int16 respectively. Then we make them an int or int16
+ALIEN_PROC_ID_MAX_NDIGITS_INT32=10
+ALIEN_PROC_ID_MAX_NDIGITS_INT16=9
+ALIEN_PROC_ID_LENTGH=`echo $ALIEN_PROC_ID | wc -l`
+ALIEN_PROC_ID_OFFSET_INT32=$((ALIEN_PROC_ID_LENTGH - ALIEN_PROC_ID_MAX_NDIGITS_INT32))
+ALIEN_PROC_ID_OFFSET_INT16=$((ALIEN_PROC_ID_LENTGH - ALIEN_PROC_ID_MAX_NDIGITS_INT16))
+# lets's take the last 10 (int) or 5 (int16) digits
+ALIEN_PROC_ID_OFFSET_INT32=${ALIEN_PROC_ID:$((ALIEN_PROC_ID_OFFSET_INT32))}
+ALIEN_PROC_ID_OFFSET_INT16=${ALIEN_PROC_ID:$((ALIEN_PROC_ID_OFFSET_INT16))}
+# let's make them int32 or int16, but not with the max possible value (which would be 0x7fffffff and 0x7fff respectively)
+# but a bit less, to allow to add [0, 15] on top afterwards if needed (e.g. we usually add
+# the NUMAID), see https://github.com/AliceO2Group/O2DPG/pull/993#pullrequestreview-1393401475
+export O2JOBID=$((ALIEN_PROC_ID_OFFSET_INT32 & 0x7ffffff0))
+export O2JOBSHMID=$((ALIEN_PROC_ID_OFFSET_INT16 & 0x7ff0))
+echo "ALIEN_PROC_ID = $ALIEN_PROC_ID, we will set O2JOBID = $O2JOBID, SHMEMID = $SHMEMID"
--- a/DATA/common/getCommonArgs.sh
+++ b/DATA/common/getCommonArgs.sh
@@ -9,7 +9,7 @@ if [[ -z $SEVERITY || -z $NUMAID || -z $SHMSIZE || -z $FILEWORKDIR || -z $EPNSYN
   exit 1
 fi
 
-ARGS_ALL="--session ${OVERRIDE_SESSION:-default} --severity $SEVERITY --shm-segment-id $(($NUMAID + ${O2JOBID:-0})) --shm-segment-size $SHMSIZE ${ARGS_ALL_EXTRA:-} --early-forward-policy noraw"
+ARGS_ALL="--session ${OVERRIDE_SESSION:-default} --severity $SEVERITY --shm-segment-id $(($NUMAID + ${O2JOBSHMID:-0})) --shm-segment-size $SHMSIZE ${ARGS_ALL_EXTRA:-} --early-forward-policy noraw"
 ARGS_ALL_CONFIG="NameConf.mDirGeom=$FILEWORKDIR;NameConf.mDirGRP=$FILEWORKDIRRUN;keyval.input_dir=$FILEWORKDIR;keyval.output_dir=/dev/null;${ALL_EXTRA_CONFIG:-}"
 if [[ $EPNSYNCMODE == 1 ]]; then
   ARGS_ALL+=" --infologger-severity $INFOLOGGER_SEVERITY"
--- a/DATA/common/getCommonArgs.sh
+++ b/DATA/common/getCommonArgs.sh
@@ -9,7 +9,7 @@ if [[ -z $SEVERITY || -z $NUMAID || -z $SHMSIZE || -z $FILEWORKDIR || -z $EPNSYN
   exit 1
 fi
 
-ARGS_ALL="--session ${OVERRIDE_SESSION:-default} --severity $SEVERITY --shm-segment-id $(($NUMAID + ${O2JOBSHMID:-0})) --shm-segment-size $SHMSIZE ${ARGS_ALL_EXTRA:-} --early-forward-policy noraw"
+ARGS_ALL="--session ${OVERRIDE_SESSION:-default} --severity $SEVERITY --shm-segment-id ${O2JOBSHMID:-$NUMAID} --shm-segment-size $SHMSIZE ${ARGS_ALL_EXTRA:-} --early-forward-policy noraw"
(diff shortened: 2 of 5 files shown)


davidrohr:
Yes, should be like this. But please do me a favor, can you run a test job manually with a larger jobid, in the same range we expect as from.the grid. I want to avoid that there is a limit with the ids we can use.

Kind Regards
David Rohr

chiarazampolli:
> Yes, should be like this. But please do me a favor, can you run a test job manually with a larger jobid, in the same range we expect as from.the grid. I want to avoid that there is a limit with the ids we can use. Kind Regards David Rohr Sent from my mobile. (Excuse the typos!)
> […](#)
> On 3 April 2023 09:52:08 CEST, Chiara Zampolli ***@***.***> wrote: @shahor02 , @davidrohr , should it not be like this? Otherwise the change that uses O2JOBID will be ineffective, or not? You can view, comment on, or merge this pull request online at: #993 -- Commit Summary -- * Setting O2JOBID using ALIEN_PROC_ID -- File Changes -- M DATA/production/configurations/2022/LHC22f/apass1/async_pass.sh (2) M DATA/production/configurations/CTFSkimming/ctf-skim-workflow.sh (1) -- Patch Links -- https://github.com/AliceO2Group/O2DPG/pull/993.patch https://github.com/AliceO2Group/O2DPG/pull/993.diff -- Reply to this email directly or view it on GitHub: #993 You are receiving this because you were mentioned. Message ID: ***@***.***&gt;

Indeed it does not work with a number like those on the grid (I copied one from a job, 2828091786), I will take a look. Below the error.

```
[INFO] WorkflowSpec getRecoWorkflow useMC 0 CCDB  http://alice-ccdb.cern.ch/
For later analysis we write a core dump to core_dump_2575207
[FATAL] error while setting up workflow in o2-ctf-reader-workflow: stoi
For later analysis we write a core dump to core_dump_2575208
[FATAL] error while setting up workflow in o2-tfidinfo-writer-workflow: stoi
For later analysis we write a core dump to core_dump_2575209
[FATAL] error while setting up workflow in o2-gpu-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575210
[FATAL] error while setting up workflow in o2-tof-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575211
[FATAL] error while setting up workflow in o2-its-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575212
[FATAL] error while setting up workflow in o2-ft0-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575213
[FATAL] error while setting up workflow in o2-trd-tracklet-transformer: stoi
For later analysis we write a core dump to core_dump_2575214
[FATAL] error while setting up workflow in o2-tpcits-match-workflow: stoi
For later analysis we write a core dump to core_dump_2575215
[FATAL] error while setting up workflow in o2-trd-global-tracking: stoi
For later analysis we write a core dump to core_dump_2575216
[FATAL] error while setting up workflow in o2-tof-matcher-workflow: stoi
For later analysis we write a core dump to core_dump_2575217
[FATAL] error while setting up workflow in o2-tpc-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575218
[FATAL] error while setting up workflow in o2-mid-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575219
[FATAL] error while setting up workflow in o2-mch-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575220
[FATAL] error while setting up workflow in o2-mft-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575221
[FATAL] error while setting up workflow in o2-fdd-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575222
[FATAL] error while setting up workflow in o2-fv0-reco-workflow: stoi
For later analysis we write a core dump to core_dump_2575223
[FATAL] error while setting up workflow in o2-zdc-digits-reco: stoi
For later analysis we write a core dump to core_dump_2575224
[FATAL] error while setting up workflow in o2-hmpid-digits-to-clusters-workflow: stoi
For later analysis we write a core dump to core_dump_2575225
[FATAL] error while setting up workflow in o2-muon-tracks-matcher-workflow: stoi
For later analysis we write a core dump to core_dump_2575226
[FATAL] error while setting up workflow in o2-globalfwd-matcher-workflow: stoi
For later analysis we write a core dump to core_dump_2575227
```

sawenzel:
Instead of `atoi` one probably needs to use `atol` for this purpose.

chiarazampolli:
Hello @davidrohr , @shahor02 ,

Adding also @ktf .

First I changed the code to use only a number of digits (from the least significant ones) such to fit an int. But this was not enough. 
As discussed yesterday with Ruben, also shm-segment-id is set using the same logic:
```
DATA/common/getCommonArgs.sh:ARGS_ALL="--session ${OVERRIDE_SESSION:-default} --severity $SEVERITY --shm-segment-id $(($NUMAID + 10 * ${O2JOBID:-0})) --shm-segment-size $SHMSIZE ${ARGS_ALL_EXTRA:-} --early-forward-policy noraw"
```
Since this value is an int16, we have to limit ourselves to even less digits, so I implemented like committed now:
```
# let's get the last 5 digits of ALIEN_PROC_ID, to be passed to define NUMAID and shm-segment-id via O2JOBID, which are int and int16 respectively. Then we make them an int16
ALIEN_PROC_ID_MAX_NDIGITS=5
ALIEN_PROC_ID_LENTGH=`echo $ALIEN_PROC_ID | wc -l`
ALIEN_PROC_ID_OFFSET=$((ALIEN_PROC_ID_LENTGH - ALIEN_PROC_ID_MAX_NDIGITS))
# lets's take the last 5 digits
ALIEN_PROC_ID_OFFSET_SHORT=${ALIEN_PROC_ID:$((ALIEN_PROC_ID_OFFSET))}
# let's make them int16
export O2JOBID=$((ALIEN_PROC_ID_OFFSET & 0x7fff))
echo "ALIEN_PROC_ID = $ALIEN_PROC_ID, we will set O2JOBID = $O2JOBID"
```
(and I removed the "* 10" in `$NUMAID + 10 * ${O2JOBID:-0` from:
```
prodtests/full-system-test/dpl-workflow.sh:[[ ! -z $TIMEFRAME_RATE_LIMIT ]] && [[ $TIMEFRAME_RATE_LIMIT != 0 ]] && ARGS_ALL+=" --timeframes-rate-limit $TIMEFRAME_RATE_LIMIT --timeframes-rate-limit-ipcid $(($NUMAID + ${O2JOBID:-0}))"
```)
This is also not enough, because the same is used also for the metrics:
```
prodtests/full-system-test/dpl-workflow.sh:    [[ ! -z $TIMEFRAME_RATE_LIMIT ]] && [[ $TIMEFRAME_RATE_LIMIT != 0 ]] && PROXY_CHANNEL+=";name=metric-feedback,type=pull,method=connect,address=ipc://${UDS_PREFIX}metric-feedback-$(($NUMAID + 10 * ${O2JOBID:-0})),transport=shmem,rateLogging=0"
```
and the metrics can be 512 at max. 
Discussing with Giulio, we tried to see with @costing if there is a way to uniquely identitfy a job on a WN with a number that is smaller than 512, but it seems not possible. 
Now the idea is to remove completely when in async the setting for the shm-segment-id and metrics, since according to @ktf  they are not needed when starting the processes via the DPL driver. If this is the case, we can go back to using ALIEN_PROC_ID, keeping only the last (from the least significant one) 9 digits to fit an int.

Let's discuss next week, when @davidrohr is back.

Cheers,

Chiara

**Comment by chiarazampolli** [comment]
> Instead of `atoi` one probably needs to use `atol` for this purpose.

Yes, this is the first thing I did. Then I encountered the issue with shm-segment-id which is in FairMQ and is an int16. 
Of course, if we get rid of setting shm-segment-id and the metrics as Giulio suggested, we can (i) either change the ipcid to long, (ii) or just take the last digits of ALIEN_PROC_ID, which for an int are well enough to insure that the value is unique.

**Comment by chiarazampolli** [comment]
Hello @davidrohr , @shahor02 , @ktf , 
Now the O2JOBID is used for the `--timeframes-rate-limit-ipcid` while SHMEMID is used for `--shm-segment-id `. The metrics does not need anything. 
Note also https://github.com/AliceO2Group/AliceO2/pull/11189 : the extra factor `10` should not be used anymore, or we exceed int32 and int16 limits.

File: DATA/production/common/setVarsFromALIEN_PROC_ID.sh [hunk 1]
```diff
@@ -0,0 +1,18 @@
...
+ALIEN_PROC_ID_MAX_NDIGITS_INT16=9
+ALIEN_PROC_ID_LENTGH=`echo $ALIEN_PROC_ID | wc -l`
+ALIEN_PROC_ID_OFFSET_INT32=$((ALIEN_PROC_ID_LENTGH - ALIEN_PROC_ID_MAX_NDIGITS_INT32))
+ALIEN_PROC_ID_OFFSET_INT16=$((ALIEN_PROC_ID_LENTGH - ALIEN_PROC_ID_MAX_NDIGITS_INT16))
+# lets's take the last 10 (int) or 5 (int16) digits
+ALIEN_PROC_ID_OFFSET_INT32=${ALIEN_PROC_ID:$((ALIEN_PROC_ID_OFFSET_INT32))}
+ALIEN_PROC_ID_OFFSET_INT16=${ALIEN_PROC_ID:$((ALIEN_PROC_ID_OFFSET_INT16))}
+# let's make them int32 or int16
+export O2JOBID=$((ALIEN_PROC_ID_OFFSET_INT32 & 0x7fffffff - 1)) # subtracting 1 since later we add the NUMAID which can be up to 1
+export SHMEMID=$((ALIEN_PROC_ID_OFFSET_INT16 & 0x7fff - 1)) # subtracting 1 since later we add the NUMAID which can be up to 1
```
davidrohr, line 17: @chiarazampolli : I think this is wrong, you cannot simply subtract one. If the orignal id was 0, you'll end up with -1 :)
I think I'd use `ALIEN_PROC_ID_OFFSET_INT16 % 0x7ff0`, i.e. modulo with a slightly smaller module. This will guarantee we can add 0 -- 15 on top afterwards.

also, SHMEID is a bit too generic, and not clear where it comes from.
Could we call it O2JOBSHMID?

davidrohr (changes requested):

DATA/production/common/setVarsFromALIEN_PROC_ID.sh
chiarazampolli: You are right!!! Implemented.

davidrohr (approved):

PR #998, opened 2023-04-05
TRD split the json file for pulseheight tasks that matching and do not match tracks

bazinski:
-  qc json that matches with ITS TPC for the pulseheight that matches tracks to digits.
-  qc json that works when ITS or TPC are not there, and of course then excludes the track matching pulseheight.

This still leaves the correct workflow being chosen for the qc workflows.

### DIFF
--- a/DATA/production/qc-async/trditstpc.json
+++ b/DATA/production/qc-async/trditstpc.json
@@ -54,6 +54,19 @@
           "type": "direct",
           "query": "digits:TRD/DIGITS;tracklets:TRD/TRACKLETS;triggers:TRD/TRKTRGRD"
         }
+      },
+      "TRDPulseHeightTrackMatch": {
+        "active": "true",
+        "taskName": "PulseHeightTrackMatch",
+        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "direct",
+          "query": "digits:TRD/DIGITS/0;tracklets:TRD/TRACKLETS/0;triggers:TRD/TRKTRGRD/0;tracks:TRD/MATCH_ITSTPC;trigrectrk:TRD/TRGREC_ITSTPC"
+        }
       }
     }
   },
--- a/DATA/production/qc-sync/trditstpc.json
+++ b/DATA/production/qc-sync/trditstpc.json
@@ -124,6 +124,35 @@
           "3",
           "1"
         ]
+      },
+      "TRDPulseHeightTrackMatch": {
+        "active": "true",
+        "taskName": "TRDPulseHeightTrackMatch",
+        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "dataSamplingPolicy",
+          "name": "trdall"
+        },
+        "taskParameters": {
+          "peakregionstart": "7.0",
+          "peakregionend": "20.0",
+          "pulseheightpeaklower": "1.0",
+          "pulseheightpeakupper": "5.0"
+        },
+        "location": "local",
+        "localMachines": [
+          "epn",
+          "localhost"
+        ],
+        "remoteMachine": "alio2-cr1-qc05.cern.ch",
+        "remotePort": "47745",
+        "mergingMode": "delta",
+        "mergerCycleMultiplier": "2",
+        "localControl": "odc"
       }
     }
   },
@@ -132,7 +161,7 @@
       "id": "trdall",
       "active": "true",
       "machines": [],
-      "query": "digits:TRD/DIGITS/0;tracklets:TRD/TRACKLETS/0;triggers:TRD/TRKTRGRD/0;rawstats:TRD/RAWSTATS/0",
+      "query": "digits:TRD/DIGITS/0;tracklets:TRD/TRACKLETS/0;triggers:TRD/TRKTRGRD/0;rawstats:TRD/RAWSTATS/0;tracks:TRD/MATCH_ITSTPC;trigrectrk:TRD/TRGREC_ITSTPC",
       "samplingConditions": [
         {
           "condition": "random",
--- a/DATA/production/qc-workflow.sh
+++ b/DATA/production/qc-workflow.sh
@@ -87,7 +87,13 @@ elif [[ -z ${QC_JSON_FROM_OUTSIDE:-} ]]; then
       fi
     fi
     [[ -z "${QC_JSON_CPV:-}" ]] && QC_JSON_CPV=consul://o2/components/qc/ANY/any/cpv-physics-qcmn-epn
-    [[ -z "${QC_JSON_TRD:-}" ]] && QC_JSON_TRD=consul://o2/components/qc/ANY/any/trd-full-qcmn-nopulseheight-epn
+    if [[ -z "${QC_JSON_TRD:-}" ]]; then
+      if has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRD; then
+        QC_JSON_TRD=consul://o2/components/qc/ANY/any/trd-full-qcmn-epn
+      else
+        QC_JSON_TRD=consul://o2/components/qc/ANY/any/trd-full-qcmn-nopulseheight-epn
+      fi
+    fi
     [[ -z "${QC_JSON_PHS:-}" ]] && QC_JSON_PHS=consul://o2/components/qc/ANY/any/phos-raw-clusters-epn
     [[ -z "${QC_JSON_GLO_PRIMVTX:-}" ]] && QC_JSON_GLO_PRIMVTX=consul://o2/components/qc/ANY/any/glo-vtx-qcmn-epn
     [[ -z "${QC_JSON_GLO_ITSTPC:-}" ]] && QC_JSON_GLO_ITSTPC=consul://o2/components/qc/ANY/any/glo-itstpc-mtch-qcmn-epn
@@ -119,7 +125,14 @@ elif [[ -z ${QC_JSON_FROM_OUTSIDE:-} ]]; then
     [[ -z "${QC_JSON_MID:-}" ]] && QC_JSON_MID=$O2DPG_ROOT/DATA/production/qc-sync/mid-digits.json && has_processing_step MID_RECO && QC_JSON_MID=$O2DPG_ROOT/DATA/production/qc-sync/mid.json
     [[ -z "${QC_JSON_CPV:-}" ]] && QC_JSON_CPV=$O2DPG_ROOT/DATA/production/qc-sync/cpv.json
     [[ -z "${QC_JSON_PHS:-}" ]] && QC_JSON_PHS=$O2DPG_ROOT/DATA/production/qc-sync/phs.json
-    [[ -z "${QC_JSON_TRD:-}" ]] && QC_JSON_TRD=$O2DPG_ROOT/DATA/production/qc-sync/trd.json
+    if [[ -z "${QC_JSON_TRD:-}" ]]; then
+      if has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRD; then
+        QC_JSON_TRD=$O2DPG_ROOT/DATA/production/qc-sync/trditstpc.json
+      else
+        QC_JSON_TRD=$O2DPG_ROOT/DATA/production/qc-sync/trd.json
+      fi
+    fi
+
     [[ -z "${QC_JSON_GLO_PRIMVTX:-}" ]] && QC_JSON_GLO_PRIMVTX=$O2DPG_ROOT/DATA/production/qc-sync/glo-vtx-qcmn-epn.json
     [[ -z "${QC_JSON_GLO_ITSTPC:-}" ]] && QC_JSON_GLO_ITSTPC=$O2DPG_ROOT/DATA/production/qc-sync/glo-itstpc-mtch-qcmn-epn.json
     if [[ -z "${QC_JSON_TOF_MATCH:-}" ]]; then
@@ -154,7 +167,13 @@ elif [[ -z ${QC_JSON_FROM_OUTSIDE:-} ]]; then
     fi
     [[ -z "${QC_JSON_CPV:-}" ]] && QC_JSON_CPV=$O2DPG_ROOT/DATA/production/qc-async/cpv.json
     [[ -z "${QC_JSON_PHS:-}" ]] && QC_JSON_PHS=$O2DPG_ROOT/DATA/production/qc-async/phs.json
-    [[ -z "${QC_JSON_TRD:-}" ]] && QC_JSON_TRD=$O2DPG_ROOT/DATA/production/qc-async/trd.json
+    if [[ -z "${QC_JSON_TRD:-}" ]]; then
+      if has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRD; then
+        QC_JSON_TRD=$O2DPG_ROOT/DATA/production/qc-async/trditstpc.json
+      else
+        QC_JSON_TRD=$O2DPG_ROOT/DATA/production/qc-async/trd.json
+      fi
+    fi
     # the following two ($QC_JSON_PRIMVTX and $QC_JSON_ITSTPC) replace $QC_JSON_GLO for async processing
     [[ -z "${QC_JSON_GLO_PRIMVTX:-}" ]] && QC_JSON_GLO_PRIMVTX=$O2DPG_ROOT/DATA/production/qc-async/primvtx.json
     [[ -z "${QC_JSON_GLO_ITSTPC:-}" ]] && QC_JSON_GLO_ITSTPC=$O2DPG_ROOT/DATA/production/qc-async/itstpc.json


martenole (changes requested):
Can you please add the same block also to sync processing running locally and to async processing?

DATA/production/qc-async/trditstpc.json
```diff
@@ -0,0 +1,100 @@
...
+        "active": "true",
+        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource_comment": "no comment",
+        "dataSource": {
+          "type": "direct",
+          "query": "digits:TRD/DIGITS/0;tracklets:TRD/TRACKLETS/0;triggers:TRD/TRKTRGRD/0;rawstats:TRD/RAWSTATS/0"
```
martenole, line 76: Aren't you missing the track matching input here?

DATA/production/qc-workflow.sh
```diff
@@ -87,7 +87,13 @@ elif [[ -z ${QC_JSON_FROM_OUTSIDE:-} ]]; then
...
     fi
     [[ -z "${QC_JSON_CPV:-}" ]] && QC_JSON_CPV=consul://o2/components/qc/ANY/any/cpv-physics-qcmn-epn
-    [[ -z "${QC_JSON_TRD:-}" ]] && QC_JSON_TRD=consul://o2/components/qc/ANY/any/trd-full-qcmn-nopulseheight-epn
+    if [[ -z "${QC_JSON_TRD:-}" ]]; then
+      if has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRD
+        QC_JSON_TRD=consul://o2/components/qc/ANY/any/trd-full-qcmn-epn
+      else
+        QC_JSON_TRD=consul://o2/components/qc/ANY/any/trd-full-qcmn-nopulseheight-epn
+      fi
+    fi
```
martenole, line 96: Ok, so you do it then as MFT with two different json files which partially overlap

martenole (commented):
Not sure about the async QC. But if the CI passes I think it is ok to merge

DATA/production/qc-workflow.sh
```diff
@@ -87,7 +87,13 @@ elif [[ -z ${QC_JSON_FROM_OUTSIDE:-} ]]; then
       fi
     fi
     [[ -z "${QC_JSON_CPV:-}" ]] && QC_JSON_CPV=consul://o2/components/qc/ANY/any/cpv-physics-qcmn-epn
-    [[ -z "${QC_JSON_TRD:-}" ]] && QC_JSON_TRD=consul://o2/components/qc/ANY/any/trd-full-qcmn-nopulseheight-epn
+    if [[ -z "${QC_JSON_TRD:-}" ]]; then
+      if has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRD
```
martenole, line 91:
```suggestion
      if has_detectors_reco ITS TPC TRD && has_detector_matching ITSTPCTRD; then
```
Also in the three other blocks

DATA/production/qc-workflow.sh
bazinski: ja it seemed easier.

martenole:
fullCI is crashing unfortunately with
```
reco_NOGPU.log-o2-dpl-run --session default --severity info --shm-segment-id 0 --shm-segment-size 8000000000  --early-forward-policy noraw --monitoring-backend no-op:// --fairmq-rate-logging 0 --timeframes-rate-limit 24 --timeframes-rate-limit-ipcid 0 -b --run 
reco_NOGPU.log-
reco_NOGPU.log:terminate called after throwing an instance of 'boost::wrapexcept<boost::property_tree::ptree_bad_path>'
reco_NOGPU.log:[ERROR] error while setting up workflow in o2-qc: /sw/slc7_x86-64/boost/v1.75.0-105/include/boost/property_tree/detail/ptree_implementation.hpp(576): Throw in function boost::property_tree::basic_ptree<K, D, C>& boost::property_tree::basic_ptree<Key, Data, KeyCompare>::get_child(const path_type&) [with Key = std::__cxx11::basic_string<char>; Data = std::__cxx11::basic_string<char>; KeyCompare = std::less<std::__cxx11::basic_string<char> >; path_type = boost::property_tree::string_path<std::__cxx11::basic_string<char>, boost::property_tree::id_translator<std::__cxx11::basic_string<char> > >]
reco_NOGPU.log-  what():  No such node (name)
reco_NOGPU.log:[ERROR] Dynamic exception type: boost::wrapexcept<boost::property_tree::ptree_bad_path>
reco_NOGPU.log:[ERROR] std::exception::what: No such node (name)
reco_NOGPU.log-[ERROR] 
killing child 72716
```
Can you please check @bazinski ?

bazinski:
@martenole ja already did a push. It seems I need a holiday ....

DATA/production/qc-workflow.sh
bazinski: thats a complete fail of copy and paste.

DATA/production/qc-async/trditstpc.json
bazinski: indeed

martenole (approved):
Without going through all jsons in detail it looks OK now

DATA/production/qc-async/trditstpc.json
```diff
@@ -0,0 +1,100 @@
...
+        "location": "local",
+        "localMachines": [
+          "epn",
+          "localhost"
+        ],
+        "remoteMachine": "alio2-cr1-qc05.cern.ch",
+        "remotePort": "47745",
+        "mergingMode": "delta",
+        "mergerCycleMultiplier": "2",
+        "localControl": "odc"
```
knopers8, line 93:
```suggestion
        }
```
You don't need all of this in async processing

DATA/production/qc-async/trd.json
```diff
@@ -62,6 +62,35 @@
           "pulseheightpeaklower": "1.0",
           "pulseheightpeakupper": "5.0"
         }
+      },
+      "PulseHeightTrackMatch": {
+        "active": "false",
```
knopers8, line 67: Why add a deactivated task?

DATA/production/qc-async/trd.json
```diff
@@ -62,6 +62,35 @@
...
+        "location": "local",
+        "localMachines": [
+          "epn",
+          "localhost"
+        ],
+        "remoteMachine": "alio2-cr1-qc05.cern.ch",
+        "remotePort": "47745",
+        "mergingMode": "delta",
+        "mergerCycleMultiplier": "2",
+        "localControl": "odc"
```
knopers8, line 93:
```suggestion
        }
```
You don't need all of that in async processing

DATA/production/qc-async/trd.json
```diff
@@ -62,6 +62,35 @@
...
         }
+      },
+      "PulseHeightTrackMatch": {
+        "active": "false",
+        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource_comment": "no comment",
```
knopers8, line 73:
```suggestion
```
You can remove it

DATA/production/qc-async/trditstpc.json
```diff
@@ -0,0 +1,100 @@
...
+    "tasks": {
+      "Digits": {
+        "active": "true",
+        "taskName": "Digits",
+        "className": "o2::quality_control_modules::trd::DigitsTask",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource_comment": "no comment",
```
knopers8, line 34:
```suggestion
```

DATA/production/qc-async/trditstpc.json
```diff
@@ -0,0 +1,100 @@
...
+      },
+      "Tracklets": {
+        "active": "true",
+        "taskName": "Tracklets",
+        "className": "o2::quality_control_modules::trd::TrackletsTask",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource_comment": "no comment",
```
knopers8, line 54:
```suggestion
```

DATA/production/qc-async/trditstpc.json
```diff
@@ -0,0 +1,100 @@
...
+        }
+      },
+      "PulseHeightTrackMatch": {
+        "active": "true",
+        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource_comment": "no comment",
```
knopers8, line 73:
```suggestion
```

DATA/production/qc-sync/trd.json
```diff
@@ -93,22 +103,90 @@
...
+        ]
+      },
+      "PulseHeightTrackMatch": {
+        "active": "false",
+        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
+        "dataSource_comment": "no comment",
```
knopers8, line 135:
```suggestion
```

DATA/production/qc-sync/trditstpc.json
```diff
@@ -0,0 +1,212 @@
...
+            "type": "Task",
+            "name": "RawDataTask",
+            "MOs": [
+              "trackletsperevent"
+            ]
+          }
+        ]
+      },
+      "PulseHeightCheck": {
+        "active": "false",
```
knopers8, line 176: why adding a deactivated check?

DATA/production/qc-sync/trd.json
```diff
@@ -93,22 +103,90 @@
...
+        ],
+        "remoteMachine": "alio2-cr1-qc05.cern.ch",
+        "remotePort": "47745",
+        "mergingMode": "delta",
+        "mergerCycleMultiplier": "2",
+        "localControl": "odc"
+      }
+    },
+    "checks": {
+      "QcCheck": {
```
knopers8, line 159: Perhaps you could use a more applicable name?

DATA/production/qc-sync/trditstpc.json
```diff
@@ -0,0 +1,212 @@
...
+        ],
+        "remoteMachine": "alio2-cr1-qc05.cern.ch",
+        "remotePort": "47745",
+        "mergingMode": "delta",
+        "mergerCycleMultiplier": "2",
+        "localControl": "odc"
+      }
+    },
+    "checks": {
+      "QcCheck": {
```
knopers8, line 159: Perhaps you could use a more applicable name?

DATA/production/qc-sync/trd.json
```diff
@@ -93,22 +103,90 @@
...
+        "mergingMode": "delta",
+        "mergerCycleMultiplier": "2",
+        "localControl": "odc",
+        "mergersPerLayer": [
+          "3",
+          "1"
+        ]
+      },
+      "PulseHeightTrackMatch": {
+        "active": "false",
```
knopers8, line 129: Why deactivated?

DATA/production/qc-async/trd.json
```diff
@@ -23,7 +23,7 @@
       }
     },
     "tasks": {
-      "TRDDigits": {
+      "Digits": {
```
knopers8, line 26:
```suggestion
      "TRDDigits": {
```
This was like this for a reason - to avoid key collisions in merged workflows. Since we combine JSONs for async processing, this will be supressed or will suppress a different task with the name `Digits`.

If you need more reading: https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#merging-multiple-configuration-files-into-one

DATA/production/qc-async/trd.json
```diff
@@ -43,7 +43,7 @@
           "pulseheightpeakupper": "5.0"
         }
       },
-      "TRDTracklets": {
+      "Tracklets": {
```
knopers8, line 46:
```suggestion
      "TRDTracklets": {
```
same comment here

DATA/production/qc-async/trditstpc.json
```diff
@@ -0,0 +1,100 @@
...
+      },
+      "consul": {
+        "url": "alio2-cr1-hv-aliecs:8500"
+      },
+      "conditionDB": {
+        "url": "alio2-cr1-hv-qcdb1.cern.ch:8083"
+      }
+    },
+    "tasks": {
+      "Digits": {
```
knopers8, line 26:
```suggestion
      "TRDDigits": {
```

DATA/production/qc-async/trditstpc.json
```diff
@@ -0,0 +1,100 @@
...
+          "query": "digits:TRD/DIGITS;tracklets:TRD/TRACKLETS;triggers:TRD/TRKTRGRD"
+        },
+       "taskParameters": {
+          "peakregionstart": "7.0",
+          "peakregionend": "20.0",
+          "pulseheightpeaklower": "1.0",
+          "pulseheightpeakupper": "5.0"
+        }
+      },
+      "Tracklets": {
```
knopers8, line 46:
```suggestion
      "TRDTracklets": {
```

DATA/production/qc-async/trditstpc.json
```diff
@@ -0,0 +1,100 @@
...
+          "peakregionstart": "7.0",
+          "peakregionend": "20.0",
+          "pulseheightpeaklower": "1.0",
+          "pulseheightpeakupper": "5.0"
+        }
+      },
+      "PulseHeightTrackMatch": {
+        "active": "true",
+        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
+        "moduleName": "QcTRD",
```
knopers8, line 69: Not sure if we risk overlapping task names in this case, but you may want to consider this
```suggestion
      "TRDPulseHeightTrackMatch": {
        "active": "true",
        "taskName": "PulseHeightTrackMatch",
        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
        "moduleName": "QcTRD",
```

knopers8 (changes requested):
The comments about reverting the change of structure keys (e.g. `TRDTracklets` to `Tracklets`) are especially important and I strongly advise against merging before these are resolved, otherwise some tasks might silently disappear in the processing.

bazinski:
@martenole the idea was to keep it as simple as possible, so jsons are effectively the same but for enabling/disabling the task that requires the matched tracks, and its associated query statements. async is only different by the increase in sampling rates.

DATA/production/qc-async/trd.json
bazinski: simplicity of all the code, so the json files are the same, it makes the checking easier. If its a problem i can remove.

DATA/production/qc-async/trditstpc.json
bazinski: Sure, but it does not break anything to it in there right? It seemed clearer to have all the files almost identical.

DATA/production/qc-async/trd.json
bazinski: as comment above.

DATA/production/qc-async/trd.json
bazinski: removing

DATA/production/qc-async/trditstpc.json
bazinski: removing

DATA/production/qc-async/trd.json
bazinski: Yip I forgot about whether the tasks should have the detector in them, and then seem to have remembered the wrong way around.   fixing.

DATA/production/qc-async/trd.json
bazinski: asbove

DATA/production/qc-async/trditstpc.json
bazinski: as above

DATA/production/qc-async/trd.json
knopers8: Ok, if that is easier for you, then sure, no problem.

DATA/production/qc-async/trditstpc.json
knopers8: It does not break anything, it is just confusing perhaps. But if that makes it easier to you, then no problem.

DATA/production/qc-async/trd.json
bazinski: removed

DATA/production/qc-async/trditstpc.json
bazinski: done

DATA/production/qc-async/trditstpc.json
bazinski: done

DATA/production/qc-async/trditstpc.json
bazinski: but then this should be done for all tasks no ? I thought the name before the { of the task is the taskname ?

DATA/production/qc-async/trditstpc.json
```diff
@@ -0,0 +1,75 @@
...
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "direct",
+          "query": "digits:TRD/DIGITS;tracklets:TRD/TRACKLETS;triggers:TRD/TRKTRGRD"
+        }
+      },
+      "TRDPulseHeightTrackMatch": {
+        "active": "true",
+        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
+        "moduleName": "QcTRD",
```
knopers8, line 61: but here you can put the proper task name. The convention is to avoid detector code in the name, as explained here: https://github.com/AliceO2Group/QualityControl/blob/master/doc/ModulesDevelopment.md#naming-convention
It's not yet applied to sync QC tasks, though.
```suggestion
      "TRDPulseHeightTrackMatch": {
        "active": "true",
        "taskName" : "PulseHeightTrackMatch",
        "className": "o2::quality_control_modules::trd::PulseHeightTrackMatch",
        "moduleName": "QcTRD",
```

DATA/production/qc-async/trditstpc.json
bazinski: done

DATA/production/qc-async/trditstpc.json
bazinski: done

DATA/production/qc-async/trditstpc.json
knopers8: We clarified this on MM, but for the record, the key is taskID, which should be unique throughout all the tasks of all the detectors. `taskName` is what you want e.g. in your QCDB path (`qc/TRD/MO/PulseHeightTrackMatch`), if it's not specified, the taskID is used.

I asked Sean to change it because the naming convention requires not using detector name in the task name. In other words, the 2nd `TRD` would be redundant: `qc/TRD/MO/TRDPulseHeightTrackMatch`

knopers8 (approved):
thanks

PR #1000, opened 2023-04-06
MC: Interaction rate from CTP scalers for anchoring

sawenzel:


chiarazampolli:
Hello @sawenzel , 

Can you resolve the conflicts?

Thanks!

Chiara

noferini:
I found many options missing in the anchoring script of this PR
All these option are not recognized
-eCM 13600 -col pp -gen pythia8 -proc inel -ns 10 -e TGeant4 -j 8

sawenzel:
> > I found many options missing in the anchoring script of this PR All these option are not recognized -eCM 13600 -col pp -gen pythia8 -proc inel -ns 10 -e TGeant4 -j 8
> 
> I managed to recover them by putting at the end of the call (since they are treated as forward args). I still have a question about parser.add_argument("--ccdb-IRate", type=bool, help="whether to try fetching IRate from CCDB/CTP", default=True) what the default is true? Then I don't know how to set it to false

probably `--cddb-IRate False`

sawenzel:
> I found many options missing in the anchoring script of this PR All these option are not recognized -eCM 13600 -col pp -gen pythia8 -proc inel -ns 10 -e TGeant4 -j 8

Ok. Will take a look what goes wrong.

PR #1001, opened 2023-04-06
Tracking parameters depend on IR and B field for 2023

chiarazampolli:
@shahor02 
Needs https://github.com/AliceO2Group/AliceO2/pull/11107

### DIFF
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -156,17 +156,18 @@ else
 fi
 
 # IR
-if [[ -z $RUN_IR ]] || [[ -z $RUN_DURATION ]]; then
+if [[ -z $RUN_IR ]] || [[ -z $RUN_DURATION ]] || [[ -z $RUN_BFIELD ]]; then
   cp $O2DPG_ROOT/DATA/production/common/getIRandDuration.C ./
   echo "In setenv_extra: time used so far = $timeUsed"
   timeStart=`date +%s`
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Hello @shahor02 , 
I have a question, concerning lines:
```
if [[ $BEAMTYPE == "PbPb" || $PERIOD == "MAY" || $PERIOD == "JUN" || $PERIOD == LHC22* ]]; then
  EXTRA_PRIMVTX_TimeMargin="pvertexer.timeMarginVertexTime=1.3"
fi
```
https://github.com/AliceO2Group/O2DPG/blob/3cf0b3395f1f3c2ed310bf2c3cc51415da7e5116/DATA/production/configurations/asyncReco/setenv_extra.sh#L295-L297
I guess we need to change them for 2023. Right? Maybe we should just disable if the data are not from 2021/
Chiara

shahor02 (commented):
Hi @chiarazampolli 
looks fine, for the `timeMarginVertexTime`: I would keep them until all problems with time alignment of different detectors are under the control, so I would simply extend to LHC23*.
Actually, also 2021 would profit from these settings.

chiarazampolli:
> Hi @chiarazampolli looks fine, for the `timeMarginVertexTime`: I would keep them until all problems with time alignment of different detectors are under the control, so I would simply extend to LHC23*. Actually, also 2021 would profit from these settings.

Hello @shahor02 , for now I did not add 2021, I think the period was OCT, but I would need to check. Since it is not urgent, I will do it with more calm another time (now I am in a hurry, only a few minutes to work)

PR #1002, opened 2023-04-07
Add timeout in analysis utilities

njacazio:


njacazio:
@benedikt-voelkel this feature gets useful when testing locally if it's ok with you I would consider having it

benedikt-voelkel:
@njacazio what is the return code if an analysis pipeline times out? I assume it will be different from `0`?

njacazio:
Hi @benedikt-voelkel it should always return 0, e.g. with
`o2-analysis-alice3-qa-singleparticle --aod-file AO2D.root -b --time-limit 1`
I get
`[32291:internal-dpl-aod-reader]: [14:29:21][INFO] Run time exceeds run time limit of 1 seconds. Exiting gracefully...`
and then
```
 echo $?
0
```

njacazio:
@benedikt-voelkel  I would merge this so that we can go on with the reso PR

benedikt-voelkel (approved):

PR #1008, opened 2023-04-13
Possibility to use remote reading + fix for external QC file

chiarazampolli:
@shahor02 , @catalinristea , @davidrohr , @noferini 
This allows to read the CTF files remotely. 
@catalinristea : I am also fixing the issue with passing a local file for QC.

### DIFF
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -23,8 +23,12 @@ fi
 echo "*********************** mode = ${MODE}"
 unset ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow
 if [[ $MODE == "remote" ]]; then
-  export INPUT_FILE_COPY_CMD="\"alien_cp ?src file://?dst\""
-  export ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow="$ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow --remote-regex \"^alien:///alice/data/.+\""
+  if [[ $ALIEN_JDL_REMOTEREADING != 1 ]]; then
+    export INPUT_FILE_COPY_CMD="\"alien_cp ?src file://?dst\""
+    export ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow="$ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow --remote-regex \"^alien:///alice/data/.+\""
+  else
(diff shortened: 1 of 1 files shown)


DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -23,8 +23,12 @@ fi
...
 if [[ $MODE == "remote" ]]; then
-  export INPUT_FILE_COPY_CMD="\"alien_cp ?src file://?dst\""
-  export ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow="$ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow --remote-regex \"^alien:///alice/data/.+\""
+  if [[ $ALIEN_JDL_REMOTEREADING != 1 ]]; then
+    export INPUT_FILE_COPY_CMD="\"alien_cp ?src file://?dst\""
+    export ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow="$ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow --remote-regex \"^alien:///alice/data/.+\""
+  else
+    export INPUT_FILE_COPY_CMD="no-copy"
+  fi
 fi
```
shahor02, line 32: You actually don't need to indicate that the files are remote: default `--remote-regex 
^(alien://|)/alice/data/.+"` covers alien files both with and w/o `alien://` prefix (though for the direct network access to work the `alien` prefix is needed).
So, it is enough to do
```
if [[ $ALIEN_JDL_REMOTEREADING != 1 ]]; then
   export INPUT_FILE_COPY_CMD="no-copy"
fi  
```

chiarazampolli: Sorry, that was a cut/paste mistake.

chiarazampolli: Hello @shahor02 , 
Do you still see issues in this PR, or was it a misunderstanding?
Chiara

shahor02: sorry, missed the comment: right, I had a typo:
```
if [[ $ALIEN_JDL_REMOTEREADING == 1 ]]; then
   export INPUT_FILE_COPY_CMD="no-copy"
fi 
```
should be enough.

chiarazampolli: Ah, ok, now I understand what you mean. I will leave it for now, and in case remove it later.

PR #1009, opened 2023-04-13
Orbits in a TF are 128, instead of 256

chiarazampolli:
@sawenzel .
We should take this number from CCDB, since in 2023 it is even different (32).
@catalinristea , @noferini , FYI.


chiarazampolli:
Merging, then we should implement a more robust way.

PR #1010, opened 2023-04-13
Forgot unbound variable

martenole:
`GEN_TOPO_AUTOSCALE_PROCESSES` is not set by default for our FST using tmux script


martenole:
Trivial, merging

PR #1014, opened 2023-04-14
Make variable known for gen topo even if empty

martenole:
I tested on prod, not on staging. On staging there is no CTP that's why I did not spot this in my test...


martenole:
Tested on the EPNs, merging

PR #1017, opened 2023-04-17
More fixes thanks to unbound variables check

martenole:
if `CALIB_TPC_VDRIFTTGL=1` was set in predefined configurations used by other detectors like HMIPD it was not disabled automatically


martenole:
Tested on the EPNs, merging

PR #1018, opened 2023-04-17
TRD noise run workflow

martenole:


martenole:
@davidrohr could you take a look and see if this would be OK to take as a reference for custom workflows? Then I would update the README a bit.
Instead of putting manually the full workflow as it is usually done e.g. in https://github.com/AliceO2Group/O2DPG/blob/2312e1d044be9594f8543fdf4f7744fdab0817fd/DATA/testing/detectors/TPC/tpc-laser.sh#L57-L83 I use the `add_W` helper and allow for printing the workflow as in dpl-workflow.sh. Also additional arguments/config key values are available like that

PR #1020, opened 2023-04-18
Added params to ITS jsons

iravasen:
Added:

- Param "publishDetailedSummary" which allows to publish more detailed hitmaps for qc async only (=1 for qc async) 
- Param "doNorm" equal to 1 which allows to normalize by n vertices --> no change here wrt what done before but behind doNorm there are now different normalization options. 

cc @chiarazampolli

### DIFF
--- a/DATA/production/qc-async/its.json
+++ b/DATA/production/qc-async/its.json
@@ -42,7 +42,7 @@
           "nThreads": "1",
 	  "nBCbins" : "103",
 	  "dicttimestamp" : "0",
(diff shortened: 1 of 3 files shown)


chiarazampolli (approved):

chiarazampolli:
Fine with me. @knopers8 , do you also want to take a look?

knopers8 (approved):
these are some task-specific arguments which I have no clue about, so fine with me ;)

PR #1021, opened 2023-04-18
Variable was unbound

chiarazampolli:
@davidrohr , @martenole , this goes with https://github.com/AliceO2Group/AliceO2/pull/11187
@catalinristea : this PR and the one in O2 above should fix the issue why we cannot run anchored MC at the moment.

### DIFF
--- a/DATA/common/setenv.sh
+++ b/DATA/common/setenv.sh
@@ -152,6 +152,7 @@ ITSMFT_STROBES=""
 [[ ! -z ${ITS_STROBE:-} ]] && ITSMFT_STROBES+="ITSAlpideParam.roFrameLengthInBC=$ITS_STROBE;"
 [[ ! -z ${MFT_STROBE:-} ]] && ITSMFT_STROBES+="MFTAlpideParam.roFrameLengthInBC=$MFT_STROBE;"
 
+if [[ -z "${SKIP_TPC_CLUSTERSTRACKS_OUTPUT:-}" ]];      then export SKIP_TPC_CLUSTERSTRACKS_OUTPUT=1; fi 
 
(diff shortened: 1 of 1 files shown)


DATA/common/setenv.sh
```diff
@@ -152,6 +152,7 @@ ITSMFT_STROBES=""
 [[ ! -z ${ITS_STROBE:-} ]] && ITSMFT_STROBES+="ITSAlpideParam.roFrameLengthInBC=$ITS_STROBE;"
 [[ ! -z ${MFT_STROBE:-} ]] && ITSMFT_STROBES+="MFTAlpideParam.roFrameLengthInBC=$MFT_STROBE;"
 
+if [[ -z "${SKIP_TPC_CLUSTERSTRACKS_OUTPUT:-}" ]];      then export SKIP_TPC_CLUSTERSTRACKS_OUTPUT=1; fi
```
martenole, line 155:
```suggestion
: ${SKIP_TPC_CLUSTERSTRACKS_OUTPUT:=} 
```
Should it not be set to an empty string by default? That could be achieved the way I wrote it above. But if you can include https://github.com/AliceO2Group/AliceO2/pull/11187 the way I suggested then we would not need this here. Unfortunately I overlooked this variable

martenole (changes requested):

DATA/common/setenv.sh
chiarazampolli: I don't really know what the default should be :-) I thought it was that we *do* skip the TPC_CLUSTERSTRACKS. 
If you like it more, I can take the suggested changes in O2, and close this one. I don't have a strong opinion.

Thanks for the feedback!!

PR #1022, opened 2023-04-19
FT0Calib:hotfix for output proxy DPL channel

afurs:


martenole (approved):
Works fine with the fix, thanks for taking quick action @afurs

PR #1023, opened 2023-04-19
Apply pixel overflow masking to ITS,MFT before skimming

shahor02:


chiarazampolli (approved):

DATA/production/configurations/CTFSkimming/ctf-skim-workflow.sh
```diff
@@ -60,11 +60,21 @@ fi
 [[ "0$ALLOW_MISSING_DET" == "00" ]] && ALLOW_MISSING_DET= || ALLOW_MISSING_DET="--allow-missing-detectors"
 [[ "0$SKIP_SKIMMED_OUT_TF" == "00" ]] && SKIP_SKIMMED_OUT_TF= || SKIP_SKIMMED_OUT_TF="--skip-skimmed-out-tf"
 
-[[ -z $DEF_MARGIN_BWD ]] && DEF_MARGIN_BWD=55 # default backward margin to account for time misalignments
-[[ -z $DEF_MARGIN_FWD ]] && DEF_MARGIN_FWD=55 # default backward margin to account for time misalignments
+[[ -z ${DEF_MARGIN_BWD:-} ]] && DEF_MARGIN_BWD=55 # default backward margin to account for time misalignments
```
chiarazampolli, line 63: I think that @martenole prefers syntax like:
```
: ${DEF_MARGIN_BWD:=55}
```
for defaults.

shahor02: It is fine to assign like that if one wants to use it once, but what about multiple entries? Actually, I see also @martenole using similar 2-stage construct https://github.com/AliceO2Group/AliceO2/blob/89be97b978aaad68ccacfe123879629ae2fbb87e/prodtests/full-system-test/dpl-workflow.sh#L52-L54

chiarazampolli: I don't think I understand what you mean by "multiple entries", sorry. 
For me it is fine to do like you do, I only thought Ole would prefer to stick to one way, when possible and we notice.

martenole: It's true that our formatting is not fully consistent in all script, but thank you for following my favorite syntax :)

PR #1025, opened 2023-04-20
Change the Base Path Files in CCDB writing

fapfap69:


DATA/production/calib/hmp-pedestals-processing.sh
```diff
@@ -72,7 +72,7 @@ then
   WORKFLOW+="--use-files "
 fi
 
-WORKFLOW+="--files-basepath 'HMP/Calib/Pedestals' "
+WORKFLOW+="--files-basepath 'HMP' "
```
chiarazampolli, line 75: Generally, I like to have `Detector/[Calib, Config]/Path3` but if @shahor02 also agrees, it is fine with me.

fapfap69:
Hi Chiara,

the definition of the Base Path is HMP, then in the Run Configuration I shall add one word (Eg. Pedestals or Calib)
and finally we have MEAN_<n> and SIGMA_<n> files...

Like this

	HMP/Pedestals/MEAN_0

or if you prefer

	HMP/Calib/MEAN_0


Antonio

chiarazampolli:
Thanks, merging.

PR #1026, opened 2023-04-20
TPC vdrift for 22c, d, e, f should be taken using a dedicated macro a…

chiarazampolli:
…lso in generalized script

@catalinristea , in case we need to reprocess 22fc, d, e, f with the new scripts.


chiarazampolli:
Merging.

PR #1027, opened 2023-04-20
update MCH pp settings

pillot:
The default parameters for clustering and time offset have been tuned and updated in O2.
Only the tracking parameters still need a special settings for pp, in particular because MCH is not yet aligned.


chiarazampolli (approved):

chiarazampolli:
Hello @pillot , 
Thanks! There is no IR dependence for these settings, right?
Chiara

pillot:
> Thanks! There is no IR dependence for these settings, right?
Hi @chiarazampolli, no, there is not.
Thanks, Philippe

chiarazampolli:
Not testes in CI.

PR #1028, opened 2023-04-21
[AnalysisQC] Disable EvenSelectionQA

benedikt-voelkel:


benedikt-voelkel:
Yes, it is. However only sporadically and in case of O(1000) events per TF it seems. Otherwise for fewer events, it goes through. But there is definitely something we need to understand.
@nburmaso is taking a look asap.

benedikt-voelkel:
@chiarazampolli @gconesab @catalinristea  shall we merge this?
It might be that this problem will not appear on the GRID, but it might (at some point)
I would then re-enable it as soon as it is taken care of

gconesab:
> 

hi @benedikt-voelkel, I would think that if there is a ticket devoted to this issue that states that once fixed revert this commit, I would say that we can proceed with the merge. Cheers

benedikt-voelkel:
Thanks for your feedback! I will open a ticket.

benedikt-voelkel:
The ticket is here: https://alice.its.cern.ch/jira/browse/O2-3703

gconesab:
Unless anyone else has opinions I will merge this later this morning

benedikt-voelkel:
Can we merge this @gconesab ? The analysis fails now not only in one but more simulation scenarios.

PR #1029, opened 2023-04-21
Add labels to overlay plots, few adjustments

benedikt-voelkel:
* by default, legends in overlay plots show "batch_i" and "batch_j"

* with rel-val --labels label1 label2 these can be overwritten

* bit of tidying up


benedikt-voelkel:
This PR will add some labels to the overlay plots as we discussed at some point. Can we merge it @chiarazampolli ?
We will add a few more small features over the coming days as well

PR #1030, opened 2023-04-24
async reco: Don't change DPL_PIPELINE_LENGTH on the EPNs, since 1NUMA workflow will be affected

davidrohr:
@chiarazampolli : On the EPNs with many TFs in flight, we should not mess with the pipeline length.

I am also not sure whether we need this on the GRID, why did you actually set it in the first place? I assume the memory for the metrics might have been quite high otherwise, but I think we don't use them anyway any more.


chiarazampolli:
I don't remember why this was added, we can go back to git and mails, but I guess it was suggested to do so. For me it is ok to remove it if not needed anymore.

PR #1033, opened 2023-04-26
Changing composition of the enhanced HF MC

rbailhac:
Change the composition of the enhanced HF MC (going back to what we have in Run 2).

### DIFF
--- a/MC/run/PWGEM/runHFToDielectrons_pp.sh
+++ b/MC/run/PWGEM/runHFToDielectrons_pp.sh
@@ -21,13 +21,13 @@ NTIMEFRAMES=${NTIMEFRAMES:-1}
 RNDSIG=$((1 + $RANDOM % 100))
 echo $RNDSIG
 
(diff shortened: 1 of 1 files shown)


gconesab (approved):

MC/run/PWGEM/runHFToDielectrons_pp.sh
```diff
@@ -21,13 +21,13 @@ NTIMEFRAMES=${NTIMEFRAMES:-1}
 RNDSIG=$((1 + $RANDOM % 100))
 echo $RNDSIG
 
-if [[ $RNDSIG -ge 0 && $RNDSIG -lt 40 ]];
+if [[ $RNDSIG -ge 0 && $RNDSIG -lt 20 ]];
```
benedikt-voelkel, line 24: More a technicality (and not directly related to you changes): You compare via `$RNDSIG -ge 0`, however, `RNDSIG` will never be `0`.
Like this the probabilities for ccbar and bbar are not exactly the same. Or is this on purpose?

PR #1034, opened 2023-04-26
FT0Calib: changing number of TF per slot for standalone calibration

afurs:


martenole:
Is there a corresponding PR in O2 which should go in parallel with this change?

afurs:
Hi, I guess this one https://github.com/AliceO2Group/AliceO2/pull/11228

chiarazampolli:
Aren't they completely independent? I don't know when the DATA/production/calib/ft0-timeoffset-aggregator.sh changed here is used, but there is no software change needed to approve it.

martenole:
Ah right, this is a standalone calibration and the change in O2 if for normal physics runs, correct? In that case this can of course simply be merged

PR #1038, opened 2023-04-26
Settings to run with full NUMA domain

chiarazampolli:
@davidrohr , can you check?
Pinging also @catalinristea

### DIFF
--- a/DATA/production/configurations/asyncReco/async_pass.sh
+++ b/DATA/production/configurations/asyncReco/async_pass.sh
@@ -340,31 +340,36 @@ if [[ -n "$ALIEN_JDL_USEGPUS" && $ALIEN_JDL_USEGPUS != 0 ]]; then
   export GPUTYPE="HIP"
   export GPUMEMSIZE=$((25 << 30))
   if [[ "0$ASYNC_PASS_NO_OPTIMIZED_DEFAULTS" != "01" ]]; then
-    if [[ $keep -eq 0 ]]; then
-      if [[ $ALIEN_JDL_UNOPTIMIZEDGPUSETTINGS != 1 ]]; then
-	export MULTIPLICITY_PROCESS_tof_matcher=2
-	export MULTIPLICITY_PROCESS_mch_cluster_finder=3
-	export MULTIPLICITY_PROCESS_tpc_entropy_decoder=2
-	export MULTIPLICITY_PROCESS_itstpc_track_matcher=3
-	export MULTIPLICITY_PROCESS_its_tracker=2
+    if [[ "ALIEN_JDL_USEFULLNUMADOMAIN" == 0 ]]; then
+      if [[ $keep -eq 0 ]]; then
+	if [[ $ALIEN_JDL_UNOPTIMIZEDGPUSETTINGS != 1 ]]; then
+	  export OPTIMIZED_PARALLEL_ASYNC=pp_1gpu  # sets the multiplicities to optimized defaults for this configuration (1 job with 1 gpu on EPNs)
+	else
+	  # forcing multiplicities to be 1
+	  export MULTIPLICITY_PROCESS_tof_matcher=1
+	  export MULTIPLICITY_PROCESS_mch_cluster_finder=1
+	  export MULTIPLICITY_PROCESS_tpc_entropy_decoder=1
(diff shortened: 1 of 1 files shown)


DATA/production/configurations/asyncReco/async_pass.sh
```diff
@@ -339,32 +339,45 @@ if [[ -n "$ALIEN_JDL_USEGPUS" && $ALIEN_JDL_USEGPUS != 0 ]]; then
   echo "Enabling GPUS"
   export GPUTYPE="HIP"
   export GPUMEMSIZE=$((25 << 30))
-  if [[ "0$ASYNC_PASS_NO_OPTIMIZED_DEFAULTS" != "01" ]]; then
-    if [[ $keep -eq 0 ]]; then
+  if [[ "ALIEN_JDL_USEFULLNUMADOMAIN" == 0 ]]; then
+    if [[ "0$ASYNC_PASS_NO_OPTIMIZED_DEFAULTS" != "01" ]]; then
```
davidrohr, line 343: Could you leave the check for `"0$ASYNC_PASS_NO_OPTIMIZED_DEFAULTS" != "01"` outside. I use that for my tests, where I set everything externally, and don't want anything to be set by this script. And it is counterintuitive if I have to set `ALIEN_JDL_USEFULLNUMADOMAIN=0` in addition.

DATA/production/configurations/asyncReco/async_pass.sh
```diff
@@ -339,32 +339,45 @@ if [[ -n "$ALIEN_JDL_USEGPUS" && $ALIEN_JDL_USEGPUS != 0 ]]; then
...
+      if [[ $keep -eq 0 ]]; then
+	if [[ $ALIEN_JDL_UNOPTIMIZEDGPUSETTINGS != 1 ]]; then
+	  export MULTIPLICITY_PROCESS_tof_matcher=2
```
davidrohr, line 346: instead of setting this manually, you can now also use `OPTIMIZED_PARALLEL_ASYNC=pp_1gpu`
And below, in the non-GPU settings, instead of setting export OMP_NUM_THREADS=6, we should now use `OPTIMIZED_PARALLEL_ASYNC=pp_8cpu` (for GRID nodes, there is also pp_16cpu for non-1NUMA allocation on the EPNs).

davidrohr (changes requested):
two comments below.
And also this file has again a bogus mixture of spaces and tabs for indentation. Would be good to clean it up.

DATA/production/configurations/asyncReco/async_pass.sh
chiarazampolli: Please, check the new version.

DATA/production/configurations/asyncReco/async_pass.sh
chiarazampolli: Please, check the new version.

DATA/production/configurations/asyncReco/async_pass.sh
```diff
@@ -340,31 +340,40 @@ if [[ -n "$ALIEN_JDL_USEGPUS" && $ALIEN_JDL_USEGPUS != 0 ]]; then
...
+      else
+	export TIMEFRAME_RATE_LIMIT=4
+      fi
       if [[ $ALIEN_JDL_UNOPTIMIZEDGPUSETTINGS != 1 ]]; then
-	export MULTIPLICITY_PROCESS_tof_matcher=2
-	export MULTIPLICITY_PROCESS_mch_cluster_finder=3
-	export MULTIPLICITY_PROCESS_tpc_entropy_decoder=2
-	export MULTIPLICITY_PROCESS_itstpc_track_matcher=3
-	export MULTIPLICITY_PROCESS_its_tracker=2
+	export OMP_NUM_THREADS=8
```
davidrohr, line 360: The OPTIMIZED_PARALLEL_ASYNC also sets the num threads (no longer globally, but per process).
Thus you can remove the OMP_NUM_THREADS here. And the OMP_NUM_THREADS=4 should go above to the "# forcing multiplicities to be 1"

chiarazampolli: done

davidrohr (approved):

PR #1040, opened 2023-04-27
Extended validity of noise calib object in ccdb to 2 years

iravasen:
Hi @shahor02, @mconcas, 
this is a simple modification to the P2 workflow of noise calibration so to have a validity of 2 years for the `NoiseMap` which is pushed to ccdb after a noise calib run. 
As discussed, we don't run a noise calibration very often hence it might well happen that we take data for 6 months with the same list of noisy pixels masked into ITS. In case a new calibration is done and a new object will be pushed to ccdb, you mentioned that the metadata `adjustableEOV = true` will make the proper job in avoiding that the old map (still with a validity of 2 years) will not be considered anymore if a new one appears. 

The modification for the `DeadMap` is instead ongoing in our DCS parser workflow.


martenole:
I believe the `adjustableEOV=true` is not relevant here, since the end of validity of a CCDB object is only changed within a relatively short time span after the object was created. If you do a noise run after about 2 months then this would not be the case. It is used more for the objects which are continuously updated.
But that should also not be needed. If 2 objects are valid then the newer one will be taken automatically. So your change should be perfectly fine.

shahor02:
Hi @ole, the limitation of changing the EoV within 5 min. from the object creation time is only for manual uploads from non-online hosts. For automatic updates from the EPN or FLP there are no limitation: the EoV of the old object will be adjusted to SoV of the new one.

martenole:
Ah that was not clear to me. Thanks for clarifying

PR #1041, opened 2023-04-27
Improved timestamp determination in anchoring

sawenzel:
Anchoring script now determined timestamp for a particular production.

In addition, we need to pass SOR to workflow script since this is still used with HBFUtils.

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -107,6 +107,7 @@ parser.add_argument('--no-tpc-digitchunking', action='store_true', help=argparse
 parser.add_argument('--combine-tpc-clusterization', action='store_true', help=argparse.SUPPRESS) #<--- useful for small productions (pp, low interaction rate, small number of events)
 parser.add_argument('--first-orbit', default=0, type=int, help=argparse.SUPPRESS)  # to set the first orbit number of the run for HBFUtils (only used when anchoring)
                                                             # (consider doing this rather in O2 digitization code directly)
+parser.add_argument('--sor', default=-1, type=int, help=argparse.SUPPRESS) # may pass start of run with this (otherwise it is autodetermined from run number)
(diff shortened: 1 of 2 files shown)


sawenzel:
@noferini

noferini:
@sawenzel ,
it looks fine to me. Can we merge it so that tomorrow we can start to run some tests.
I did run with a similar configuration and I got a reasonable behavior but for one query to ccdb in the grpcreate which was still using SOR time. I didn't have the time to check how the timestamp is set there but if you can quickly check this is an important point (MeanVertex anchoring depends on it)

grpcreate.log:[INFO] ccdb reads http://alice-ccdb.cern.ch//GLO/Calib/MeanVertex/1657458449840/ece16cec-6014-11ed-8420-200114580202 for 1657458449840 (retrieve, agent_id: node083.cluster-1682581652-WxCDVs

For completeness, matbuddownloader uses the right timestamp
matbuddownloader.log:[INFO] ccdb reads http://alice-ccdb.cern.ch/GLO/Param/MatLUT/1636019021295/b9bd65e0-3d53-11ec-8357-2a010e0a09fb for 1657462048426 (retrieve, agent_id: node083.cluster-1682581652-yGJ8

noferini:
I open a PR in O2
https://github.com/AliceO2Group/AliceO2/pull/11248
to allow passing timestamp to grpcreate

sawenzel:
> I open a PR in O2 [AliceO2Group/AliceO2#11248](https://github.com/AliceO2Group/AliceO2/pull/11248) to allow passing timestamp to grpcreate

@noferini : I think this is by design. The GRP is an object which is run dependent but not really time-dependent ... or does it change within a run?

noferini:
The mean vertex is expected to change within the run and it is managed here
bool create_MeanVertexObject(Options const& opts)
{
  // either
  const char* CCDBPATH = "/GLO/Calib/MeanVertex";
  if (opts.vertex == "ccdb") {
    anchor_GRPs(opts, {CCDBPATH});

PR #1043, opened 2023-04-28
Fix a problem in setenv when file limits are unlimited

sawenzel:
There was a crash in setenv.sh when `ulimit -H -n` returned "unlimited". In this case this variable cannot undergo numeric comparison.

This commit fixes this by treating this case separately.

Fixes https://alice.its.cern.ch/jira/browse/O2-3731


davidrohr:
I just opened https://github.com/AliceO2Group/O2DPG/pull/1044 in parallel :)
I'd probably stick to mine since it is shorter

PR #1048, opened 2023-05-01
bug fix for anchoring

noferini:
@sawenzel 
This is a trivial fix I didn't notice before.
I realized it after finding a misbehavior in the output of simulation (EOR used instead of SOR).
This should fix. All the rest looks fine.
I think there is no need to wait for checks


noferini:
@catalinristea ,
I think you can resume the anchoring test production with the new tag (once this PR is merged)

chiarazampolli (approved):

PR #1051, opened 2023-05-03
MCH: temp fix for QC ASYNC

aphecetche:
This is a temporary measure to allow MCH async QC to at least  _not_ crash and produce _some_ histograms that can be used to assess async reco quality.

We still miss the post-processing part though. That will be added in another PR.

@csuire @aferrero2707


aphecetche:
@aferrero2707 the _plan_ is to have the checkers (running on the post-processed histograms) like in sync, yes

aphecetche:
@chiarazampolli sorry I lost track of this PR. Could it be merged ?

PR #1055, opened 2023-05-04
[EMCAL-917] Add analysis-level QA tasks for EMCAL

mfasDa:
Tasks:

(minimal)
- EMCAL event selection
- EMCAL cell monitor

(optional)
- EMCAL cluster monitor
- EMCAL track matching monitor

Including dependencies (i.e. correction tasks).
Track matching and event selection QA will
likely not be portable to the async QC. Several
histos in the cell monitor are too heavy for the
QC framework.

### DIFF
--- a/MC/config/analysis_testing/json/analyses_config.json
+++ b/MC/config/analysis_testing/json/analyses_config.json
@@ -247,6 +247,26 @@
                       "o2-analysis-pid-tpc-full",
                       "o2-analysis-pid-tof-base",
                       "o2-analysis-pid-tof-full"]
+        },
+        {
+            "name": "EMCAL",
+            "enabled": true,
+            "expected_output": ["AnalysisResults.root"],
+            "config" : {
+                "data": "json://${O2DPG_ROOT}/MC/config/analysis_testing/json/analysis-testing-data.json"
+            },
+            "tasks": ["o2-analysis-je-emc-eventselection-qa",
+                      "o2-analysis-je-emc-cellmonitor",
+                      "o2-analysis-je-emcal-correction-task",
+                      "o2-analysis-je-emc-clustermonitor",
+                      "o2-analysis-je-emc-tmmonitor",
+                      "o2-analysis-timestamp",
+                      "o2-analysis-track-propagation",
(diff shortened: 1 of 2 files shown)


mhemmer-cern:
Track Matching config looks good

chiarazampolli:
Seems fine to me, @benedikt-voelkel , can you also check?

mfasDa:
@chiarazampolli was just a protection for the review phase. Now it is set to enabled.

mfasDa:
Hi @benedikt-voelkel, Hi @chiarazampolli,

I needed to add a dependency of a dependency, but the I was able to launch.

Not sure all detectors should handle changes in core wagons themselves, since they might change in time (in our case a new dependency was added to the track propagation wagon) - would be good to handle such wagons at the central level.

Otherwise the PR should be ready for merging.

Cheers

Markus

benedikt-voelkel:
Hej @mfasDa 
thanks for taking care.

It is true, there is some maintenance for the dependencies to be done, but we introduced the possibility of having separate configurations for each analysis on purpose. In that way, none of the changes for one analysis impact the other which in the case of our AnalysisQC is something we want; in particular if we need to react quickly for specific tasks.
And there are various different people who are assigned to the different tasks so they should/will take care.

benedikt-voelkel:
I have a quick question @mfasDa : Would it be possible to add a version for MC as well? It just appeared to me that this is for data only.
Only then, we could reasonably add the analysis task to our nightly testing

mfasDa:
Yes, I will add it in a second step, but could the data version please already be merged so it will be used in upcoming productions? Thanks!

benedikt-voelkel (approved):

chiarazampolli:
Hello @benedikt-voelkel , 
Is it ok for you if I merge? (or you can also do so, in case). Or do you prefer to have first the MC integration?
Chiara

benedikt-voelkel:
I just merged @chiarazampolli 
Just approved from my side earlier to let the merging be done by you or @mfasDa .
But now we have it :)

chiarazampolli:
Great, I did not notice you had already approved! Thanks!

PR #1059, opened 2023-05-06
Light-flavour neutral meson enhanced MC simulations for dielectrons

rbailhac:
Injected pi0, eta, etaprime, rho, omega, phi (and J/Psi) decaying into dielectrons + X for Pb--Pb and pp background events

### DIFF
--- a/MC/config/PWGEM/external/generator/GeneratorCocktailLF.C
+++ b/MC/config/PWGEM/external/generator/GeneratorCocktailLF.C
@@ -1,0 +1,257 @@
+//
+//
+//
+
+R__ADD_INCLUDE_PATH($O2DPG_ROOT/MC/config/PWGDQ/EvtGen)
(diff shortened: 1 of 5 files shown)


rbailhac:
Dear @benedikt-voelkel , @gconesab  and all, I am not 100% sure how to solve the build/O2DPG/sim/o2. It looks like that it is complaining about an include #include "Pythia8/Pythia.h" in ${O2DPG_ROOT}/MC/config/PWGEM/external/generator/GeneratorCocktailLF.C where Pythia.h is not found. But I do not have this include and can not reproduce the error locally on my machine. Should I simply add: 

R__ADD_INCLUDE_PATH($PYTHIA_ROOT/include)

#include "Pythia8/Pythia.h"

on top in GeneratorCocktailLF.C ?

Thanks for the help,

Raphaelle

PR #1064, opened 2023-05-10
Fix flags interference for AOD producer

nburmaso:


chiarazampolli:
Hello @nburmaso , 
It is weird. For me the wf seems fine, I see:
```
o2-aod-producer-workflow --session default_643650_18112 --severity info --shm-segment-id 0 --shm-segment-size 16000000000  --early-forward-policy noraw --monito
ring-backend no-op:// --fairmq-rate-logging 0 --timeframes-rate-limit 3 --timeframes-rate-limit-ipcid 0  --info-sources ITS-TPC,TPC-TRD,ITS-TPC-TRD,TPC-TOF,ITS-
TPC-TOF,TPC-TRD-TOF,ITS-TPC-TRD-TOF,MFT-MCH,MCH-MID,ITS,MFT,TPC,TOF,FT0,MID,EMC,PHS,CPV,ZDC,FDD,HMP,FV0,TRD,MCH,CTP --disable-root-input --aod-writer-keep dangl
ing --aod-writer-resfile "AO2D" --aod-writer-resmode UPDATE --disable-mc --pipeline aod-producer-workflow:1  --aod-writer-maxfilesize 8000 --lpmp-prod-tag "LHC2
2m" --reco-pass "apass2" --aod-producer-workflow "--ctpreadout-create 1" --configKeyValues "NameConf.mDirGeom=/home/zampolli/work/O2/RelVal/CPU_2;NameConf.mDirG
RP=/home/zampolli/work/O2/RelVal/CPU_2;keyval.input_dir=/home/zampolli/work/O2/RelVal/CPU_2;keyval.output_dir=/dev/null;;" | \
```
where all options are present.

chiarazampolli:
Maybe the reason is a combination of the fact that the aod-producer-workflow has only one workflow inside ("aod-producer https://github.com/AliceO2Group/AliceO2/blob/7f4a63c4932b1a554cd94caa09cef68d72504f68/Detectors/AOD/src/aod-producer-workflow.cxx#L67) and of the precedence rules in DPL for options? 

What I know for example is that in case of duplication, the config key values that come last are always having precendence over the others, so if you repeat one, the value that is valid is always the last. For arguments, I think that one cannot repeat the same 2x.

Repeating `--aod-producer-workflow` might tell DPL that from that moment on only he will be given the arguments for this processor, and since this is the only one, the ones before are just ignored? @ktf , can you confirm?



I left it there since it is what was given to us by EMCAL when they asked to have this option added, and I thought that maybe rge wf was build with more processors, but in fact there is only this one, so it is indeed not needed (see https://github.com/AliceO2Group/AliceO2/blob/dev/Detectors/AOD/src/aod-producer-workflow.cxx).

chiarazampolli:
Hello @nburmaso , 
For me it does not work also removing the `--aod-producer-workflow`, this is how the producer is run:
```
o2-aod-producer-workflow --session default_663601_26047 --severity info --shm-segment-id 0 --shm-segment-size 16000000000  --early-forward-policy noraw --monito
ring-backend no-op:// --fairmq-rate-logging 0 --timeframes-rate-limit 3 --timeframes-rate-limit-ipcid 0  --info-sources ITS-TPC,TPC-TRD,ITS-TPC-TRD,TPC-TOF,ITS-
TPC-TOF,TPC-TRD-TOF,ITS-TPC-TRD-TOF,MFT-MCH,MCH-MID,ITS,MFT,TPC,TOF,FT0,MID,EMC,PHS,CPV,ZDC,FDD,HMP,FV0,TRD,MCH,CTP --disable-root-input --aod-writer-keep dangl
ing --aod-writer-resfile "AO2D" --aod-writer-resmode UPDATE --disable-mc --pipeline aod-producer-workflow:1  --aod-writer-maxfilesize 8000 --lpmp-prod-tag "LHC2
2m" --reco-pass "apass2" --ctpreadout-create 1 --configKeyValues "NameConf.mDirGeom=/home/zampolli/work/O2/RelVal/CPU_2;NameConf.mDirGRP=/home/zampolli/work/O2/
RelVal/CPU_2;keyval.input_dir=/home/zampolli/work/O2/RelVal/CPU_2;keyval.output_dir=/dev/null;;" | \
```
Can you compare to yours?

nburmaso:
Dear @chiarazampolli,

In my case, the command for aod-producer is identical to yours. I have just rebuilt 
O2PDPSuite and ran async_pass.sh (see attached, txt is for github) with command `. async_pass.sh o2_ctf_run00523308_orbit0319093916_tf0000000001_epn233.root -rnb 523308 -b pp -m LOCAL -p LHC22m -pa test > test.log`. It produced AOD file with proper metadata. Here I also attach `test.log` and `workflowconfig.log`.

[async_pass.txt](https://github.com/AliceO2Group/O2DPG/files/11448980/async_pass.txt)
[test.log](https://github.com/AliceO2Group/O2DPG/files/11448977/test.log)
[workflowconfig.log](https://github.com/AliceO2Group/O2DPG/files/11448982/workflowconfig.log)

chiarazampolli:
Hello @mfasDa , 
Can you confirm that this change will preserve the correct use of the option? Since this is needed for EMC, I prefer that you check too (even though, as I said, removing it in my tests did not fill the metadata, but maybe I did something wrong).
Chiara

mfasDa:
Hello @chiarazampolli, CTP readout (even emulated) is not only needed of EMCAL but ALL triggered detectors (i.e. TRD and PHOS will need it as well), because it has to be taken into account i.e. in the normalisation whether a detector was read out and unlike the continuously read out detectors which are live for the entire timeframe triggered detectors are only live for a fraction of collisions due to busy time.

The only thing I can tell is that we need to pass the option `--ctp-readout-create 1` needs to be passed to the AODProducerWorkflowSpec for the given list of runs in order to trigger the CTP readout calculation form the trigger status in the detector CTF, as CTP readout was not available in the list of runs. Framework experts will have to tell whether now one can drop the spec in order to access a config option of the spec (as which the option is defined): I learned that in order to set a Config option of the spec I have to specify the spec before for which I want to call the option. The option is not implemented as workflow param (for which one does not need to specify a spec).

Cheers

Markus

PR #1065, opened 2023-05-10
Script that allows to use TPC calibration

chiarazampolli:
The TPC calibration rescaling will be applied only if:
```
if [[ -z $ALIEN_JDL_DONOTAPPLYTPCSPDCALIBRATION ]] || [[ $ALIEN_JDL_DONOTAPPLYTPCSPDCALIBRATION != 1 ]]
```
which means that it will be ON by default. In this case, there will also be no TPCCLUSTERSHIFT.
@shahor02 , if you could check this PR, it would be very good.
I took at starting point the setenv_extra_v1.sh script in O2-3701.
One further remark: in the script we set:
```
TPCEXTRAERR=";GPU_rec_tpc.clusterError2AdditionalY=0.1;GPU_rec_tpc.clusterError2AdditionalZ=0.15;"
```
but according to @wiechula , it should have been:
```
TPCEXTRAERR=";GPU_rec_tpc.clusterError2AdditionalY=0.01;GPU_rec_tpc.clusterError2AdditionalZ=0.0225;"
```
i.e. the square of the values. @wiechula said to keep as in the tests done in O2-3701, but we might want to change it. 

One more comment: the scaling is disabled if we are in MC. But if we are in MC, it can be used provided that ```$ALIEN_JDL_SIMULATEDISTORTIONSINMC == 1```

And one question: what about 22s? and the low IR 22cdef, and 22q? Should I disable the scaling e.g. setting by hand:
```
--corrmap-lumi-inst 0
```
? Then only the reference map would be used.

### DIFF
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -253,16 +253,16 @@ elif [[ $ALIGNLEVEL == 1 ]]; then
   # enabling TPC calibration scaling
   # the default is to use CTP, unless specified differently in the JDL...
   INST_IR_FOR_TPC=${ALIEN_JDL_INSTIRFORTPC-CTP}
-  #...but for 2022 data, where we will rely on different settings depending on the period
+  #...but for 2022 data, where we will rely on different settings depending on the period; note that if ALIEN_JDL_INSTIRFORTPC is set, it has precedence
   if [[ $ALIEN_JDL_LPMANCHORYEAR == "2022" ]]; then
-    INST_IR_FOR_TPC="CTPCCDB"
+    INST_IR_FOR_TPC=${ALIEN_JDL_INSTIRFORTPC-CTPCCDB}
   fi
   if [[ $PERIOD == "LHC22s" ]]; then
-    INST_IR_FOR_TPC=0 # in this way, only TPC/Calib/CorrectionMaps is applied, and we know that for 22s it is the same as TPC/Calib/CorrectionMapsRef
+    INST_IR_FOR_TPC=${ALIEN_JDL_INSTIRFORTPC-0} # in this way, only TPC/Calib/CorrectionMaps is applied, and we know that for 22s it is the same as TPC/Calib/CorrectionMapsRef; note that if ALIEN_JDL_INSTIRFORTPC is set, it has precedence
   fi
-  # in MC, we set it to a negative value to disable completely the corrections
+  # in MC, we set it to a negative value to disable completely the corrections; note that if ALIEN_JDL_INSTIRFORTPC is set, it has precedence
   if [[ $ALIEN_JDL_LPMPRODUCTIONTYPE == "MC" ]]; then
-    INST_IR_FOR_TPC="-1"
+    INST_IR_FOR_TPC=${ALIEN_JDL_INSTIRFORTPC--1}
   fi
 
   # now we set the options
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -249,59 +249,65 @@ elif [[ $ALIGNLEVEL == 1 ]]; then
   [[ -z $ITS_CONFIG || "$ITS_CONFIG" != *"--tracking-mode"* ]] && export ITS_CONFIG+=" --tracking-mode async"
   CUT_MATCH_CHI2=160
   export ITSTPCMATCH="tpcitsMatch.safeMarginTimeCorrErr=10.;tpcitsMatch.cutMatchingChi2=$CUT_MATCH_CHI2;;tpcitsMatch.crudeAbsDiffCut[0]=6;tpcitsMatch.crudeAbsDiffCut[1]=6;tpcitsMatch.crudeAbsDiffCut[2]=0.3;tpcitsMatch.crudeAbsDiffCut[3]=0.3;tpcitsMatch.crudeAbsDiffCut[4]=5;tpcitsMatch.crudeNSigma2Cut[0]=100;tpcitsMatch.crudeNSigma2Cut[1]=100;tpcitsMatch.crudeNSigma2Cut[2]=100;tpcitsMatch.crudeNSigma2Cut[3]=100;tpcitsMatch.crudeNSigma2Cut[4]=100;"
+
   # enabling TPC calibration scaling
-  if [[ -z $ALIEN_JDL_DONOTAPPLYTPCSPDCALIBRATION ]] || [[ $ALIEN_JDL_DONOTAPPLYTPCSPDCALIBRATION != 1 ]]; then
-      echo "Enabling TPC calibration scaling using visible IR = $RUN_IR"
-      echo "Setting TPCCLUSTERTIMESHIFT to 0"
-      TPCCLUSTERTIMESHIFT=0
-      if [[ $ALIEN_JDL_LPMPRODUCTIONTYPE != "MC" ]] || [[ $ALIEN_JDL_SIMULATEDISTORTIONSINMC == 1 ]]; then
-	export TPC_CORR_SCALING="--corrmap-lumi-inst $RUN_IR"
-      fi
-      TRACKTUNETPCINNER="trackTuneParams.tpcCovInnerType=1;trackTuneParams.tpcCovInner[0]=0.01;trackTuneParams.tpcCovInner[1]=1.;trackTuneParams.tpcCovInner[2]=4e-7;trackTuneParams.tpcCovInner[3]=4.e-5;trackTuneParams.tpcCovInner[4]=6.8e-6;"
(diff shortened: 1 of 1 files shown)


shahor02:
Hi @chiarazampolli 

You are right for additive errors, one should use squared one.
I've also started to modify the script but could not test it since my network connection is quite bad.
I had:
```

# To impose using the workflow CTP Lumi for rescaling use ALIEN_JDL_INST_IR_FOR_TPC="CTP"
# To rely on the run IR extracted from the RCT use ALIEN_JDL_INST_IR_FOR_TPC=-1 (or with any negagive int value)
# To impose arbitrary inst. IR use ALIEN_JDL_INST_IR_FOR_TPC=<positive int IR>
# To override the mean man lumi extracted from the map use ALIEN_JDL_MEAN_IR_FOR_TPC=<positive int IR>

if [[ ! -z "${ALIEN_JDL_INST_IR_FOR_TPC:-1}" ]]; then    # inst lumi for scaling is needed
  if [[ $ALIEN_JDL_INST_IR_FOR_TPC > 0 ]]; then # externally imposed IR for scaling
    export TPC_CORR_SCALING+=" --corrmap-lumi-inst $ALIEN_JDL_INST_IR_FOR_TPC "
  else if [[ $ALIEN_JDL_INST_IR_FOR_TPC < 0 ]]; then
    export TPC_CORR_SCALING+=" --corrmap-lumi-inst $$RUN_IR "
  else if [[  $ALIEN_JDL_INST_IR_FOR_TPC == "CTP" ]]; then
    if ! has_detector CTP ; then
      echo "TPC correction with CTP Lumi is requested but CTP is not in the WORKFLOW_DETECTORS=$WORKFLOW_DETECTORS"
      exit 1	
    fi	       
    export TPC_CORR_SCALING+=" --require-ctp-lumi "
  else
    echo "Unknown setting for ALIEN_JDL_INST_IR_FOR_TPC = $ALIEN_JDL_INST_IR_FOR_TPC "
    exit 1
  fi
fi

if [[ ! -z "${ALIEN_JDL_MEAN_IR_FOR_TPC:-}" && $ALIEN_JDL_MEAN_IR_FOR_TPC > 0 ]]; then # externally imposed TPC map mean IR for scaling
  export TPC_CORR_SCALING+=" --corrmap-lumi-mean $ALIEN_JDL_INST_IR_FOR_TPC "
fi
```
Note that for 2023 data we have the CTP lumi stored in the data so one better uses:
```
ALIEN_JDL_INST_IR_FOR_TPC == "CTP"
```
to pass `--require-ctp-lumi`.

I would suppress the unbiasing options, I don't think we will ever use them and if we do, we better use legacy setenv_extra. With them the script becomes too messy.

For the MC: unfortunately, it is not only matter of scaling: since we don't simulated distortions, we should not apply correction either for reconstruction. I am afraid at the moment we should use `condition-remap` to dummy empty maps (rescaling to ~0 IR will also not work since the CorrectionRef is always applied.

For the 22s: we should not rescale but just upload low-IR maps both as TPC/Calib/CorrectionMaps and  TPC/Calib/CorrectionMapsRef.

chiarazampolli:
Hi @shahor02 , 

Thanks for the feedback, I will check tomorrow your suggestion.
So I will: 
- change the values of the additive errors;
- change the behaviour for 2023 (I forgot, indeed)

If you think we'll never use unbiasing, I will remove it. Initially I wanted to create a new script, but maintaining two is not the right way.
For MC: do you mean that we need also some Ref and Corr map with 0 inside, to remap to? To the test CCDB? I think we should avoid using user spaces.
For 22s I can just force that the inst lumi is 0 like I mentioned above, then only the reference will be applied, is this ok? Same for 22cdef and 22q?

shahor02:
Hi @chiarazampolli 

Given that any correction map is better than the "unbiasing" I don't see why we should use the latter.

The convention for the maps currently implemented is:
1) if the scaling coefficient is exactly 0 (== either inst. or mean lumi is <=0) then only TPC/Calib/CorrectionMaps is applied.
2) Otherwise `(CorrectionMaps - CorrectionMapsRef)*scaling + CorrectionMapsRef` is used.
So, for the LHC22s, in order to have only static correction we should either upload the static correction (i.e. low IR) as CorrectionMaps and impose 0. lumi or have the usual high and low IR maps uploaded and use small but non-0 IR (e.g. 1 Hz will effectively mean almost 0 scaling factor hence CorrectionMapsRef applied only. I.e. `ALIEN_JDL_INST_IR_FOR_TPC=1` in my snippet.

For the MC: given that at least the static map is applied, we should make sure it is dummy (0). If you want to avoid private ccdb repositories, I can modify the convention above that with negative IR it enforces correction to 0.

Cheers,
 Ruben

chiarazampolli:
Hello @shahor02 , 
Can you check the present version?

DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -246,45 +247,72 @@ elif [[ $ALIGNLEVEL == 1 ]]; then
...
+    echo "Using CTP inst lumi stored in data"
+    export TPC_CORR_SCALING+=" --require-ctp-lumi "
+  else
+    echo "Unknown setting for INST_IR_FOR_TPC = $INST_IR_FOR_TPC (with ALIEN_JDL_INST_IR_FOR_TPC = $ALIEN_JDL_INST_IR_FOR_TPC)"
+    return 1
   fi
-  export ITSTPCMATCH="tpcitsMatch.safeMarginTimeCorrErr=10.;tpcitsMatch.cutMatchingChi2=$CUT_MATCH_CHI2;;tpcitsMatch.crudeAbsDiffCut[0]=6;tpcitsMatch.crudeAbsDiffCut[1]=6;tpcitsMatch.crudeAbsDiffCut[2]=0.3;tpcitsMatch.crudeAbsDiffCut[3]=0.3;tpcitsMatch.crudeAbsDiffCut[4]=5;tpcitsMatch.crudeNSigma2Cut[0]=100;tpcitsMatch.crudeNSigma2Cut[1]=100;tpcitsMatch.crudeNSigma2Cut[2]=100;tpcitsMatch.crudeNSigma2Cut[3]=100;tpcitsMatch.crudeNSigma2Cut[4]=100;"
+
+  if [[ -n $ALIEN_JDL_MEANIRFORTPC && $ALIEN_JDL_MEANIRFORTPC > 0 ]]; then # externally imposed TPC map mean IR for scaling
+    export TPC_CORR_SCALING+=" --corrmap-lumi-mean $ALIEN_JDL_INST_IR_FOR_TPC "
```
shahor02, line 294: there is a typo here, should be: 
```
    export TPC_CORR_SCALING+=" --corrmap-lumi-mean $ALIEN_JDL_MEANIRFORTPC "
```

DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -246,45 +247,72 @@ elif [[ $ALIGNLEVEL == 1 ]]; then
...
+    echo "We are in period $PERIOD, we need to keep the correction for the TPC cluster time, since no new vdrift was extracted"
+  fi
+
+  TRACKTUNETPCINNER="trackTuneParams.tpcCovInnerType=1;trackTuneParams.tpcCovInner[0]=0.01;trackTuneParams.tpcCovInner[1]=1.;trackTuneParams.tpcCovInner[2]=4e-7;trackTuneParams.tpcCovInner[3]=4.e-5;trackTuneParams.tpcCovInner[4]=6.8e-6;"
+  TRACKTUNETPCOUTER="trackTuneParams.tpcCovOuterType=1;trackTuneParams.tpcCovOuter[0]=0.01;trackTuneParams.tpcCovOuter[1]=1.;trackTuneParams.tpcCovOuter[2]=4e-7;trackTuneParams.tpcCovOuter[3]=4.e-5;trackTuneParams.tpcCovOuter[4]=6.8e-6;"
+
 fi
 
+# adding additional cluster errors
+TPCEXTRAERR=";GPU_rec_tpc.clusterError2AdditionalY=0.1;GPU_rec_tpc.clusterError2AdditionalZ=0.15;"
```
shahor02, line 310: Will you fix the this to squared errors?

shahor02 (commented):
Hi @chiarazampolli 

Your summary is fine, for 2023 pp it is fine to apply a proper scaling with 500kHz / LowIR maps and using CTP lumi from the data, even for low IR runs.

DATA/production/configurations/asyncReco/setenv_extra.sh
chiarazampolli: Yes, sorry, I had copied-pasted. 
Just FYI, when you use an "ALIEN_JDL" variable, all that comes after that should be 1 single word.

DATA/production/configurations/asyncReco/setenv_extra.sh
chiarazampolli: Yes, done

PR #1066, opened 2023-05-11
Fix GEN_TOPO_ENV_AT_RUNTIME handling

martenole:
Adding missing export


davidrohr:
Ah shit, I just checked that it doesn't crash and that the variable is in the command line, but not that it is really set...

martenole:
No problem. I thought it was working fine, since there were no complaints so far. But yesterday CPV had issues on staging with curl requests to CCDB and luckily Sergey tried both the O2DPG master and my fork were I had added that variable hard-coded and saw that it was working only with my fork. So it was clear where to check.

PR #1071, opened 2023-05-16
Preload trivial TPC space-charge corrections

sawenzel:
This is preloading trivial space charge corrections into the CCDB cache so that MC workflows effectively do not account for this effect.

This is needed in order to have consistency between digitization and reconstruction. (Reconstruction was fetching the corrections but digitization does not produce the effect yet).

As soon as TPC digitization is able to account for the effect, this commit should be reverted.


sawenzel:
@noferini @catalinristea

noferini:
Thanks

shahor02:
@sawenzel Note that with https://github.com/AliceO2Group/AliceO2/pull/11336 corrections will be ignored if any of `--corrmap-lumi-inst` or `--corrmap-lumi-mean` options has a negative value.

PR #1072, opened 2023-05-16
Trigger version of enhanced heavy-flavour dielectron MC (no gap)

rbailhac:
Trigger version of enhanced heavy-flavour dielectron MC (still without gap)

### DIFF
--- a/MC/config/PWGEM/external/generator/GeneratorHFToEleFull_EvtGen.C
+++ b/MC/config/PWGEM/external/generator/GeneratorHFToEleFull_EvtGen.C
@@ -1,0 +1,48 @@
+//
+//
+R__ADD_INCLUDE_PATH($O2DPG_ROOT/MC/config/PWGDQ/EvtGen)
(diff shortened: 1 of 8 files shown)


benedikt-voelkel:
@gconesab @rbailhac Did you check that the MC simulation CI passed? It seems it was not yet run before the merge. Just to make sure :)

gconesab:
> @gconesab @rbailhac Did you check that the MC simulation CI passed? It seems it was not yet run before the merge. Just to make sure :)

My fault, confused between AliDPG and O2DPG

rbailhac:
I will need time to run myself the MC simulation CI like here unfortunately. It looks like also that it checked only the first time but then skipped it (although I had a fix). If you prefer you can revert the commit.

benedikt-voelkel:
No, it's fine. If you could just run the test yourself at some point, that would be great. In case a fix is needed, just open another PR I would say.
No problem

PR #1076, opened 2023-05-22
Fix FIT integrated currents calib conditions

martenole:


martenole:
Tested locally, merging

PR #1080, opened 2023-05-25
PWGHF: implement gap triggered MC for OmegacToXiPi

ZFederica:
Implement code for gap triggered MC for OmegaC --> Xi pi


ZFederica:
@fgrosa could you please have a look?

gconesab (approved):

benedikt-voelkel:
@TimoWilken do you know why the sim check is not scheduled?

TimoWilken:
Hi @benedikt-voelkel, it says "started" now. Probably something else was building when this PR was submitted, sorry it's a bit slow sometimes!

fgrosa:
@ZFederica sorry for the late reply, but I was attending LHCP. This is good for me, but I thought that it would be better to use the already defined methods, just adding a couple of arguments to pass the pdg of the hadron to trigger on and the rapidity window of it. In this way, the same method can be used for any triggered hadron and we do not have to define a new method for each hadron. Let me know if you want to implement it, otherwise I will do it. Thanks!

PR #1082, opened 2023-05-25
Revise o2dpg_sim_metrics.py

benedikt-voelkel:
* harmonise resource extraction

  * everything is put into a pandas dataframe
  * derive everything from there instead of re-organising everything first
  * extract time estimates of each task now also from there (previously was relying on the times from <task>.log_time

* pass arguments from o2dpg_sim_workflow.py through so that all of them end up in pipeline_metric_<pid>.log for later usage

* 3 sub-commands:

  * history: produce plots of 1. resources vs. iterations 2. resources of tasks grouped into categories (bar and pie chart) 3. resources distribution of tasks within digi and reco category (pie chart) 4. a stacked bar chart (one bar per iteration) showing the composition of resources by tasks

  * compare: plot resources per task based on a feature e.g. eCM if various different pipeline_metric with different eCM values are given, resources are put into a comparison plot for different eCM values

  * influx: make a file that can be pushed to InfluxDB resource needs per category (as also extracted by history command)

### DIFF
--- a/MC/bin/o2_dpg_workflow_runner.py
+++ b/MC/bin/o2_dpg_workflow_runner.py
@@ -28,6 +28,9 @@ import argparse
 import psutil
 max_system_mem=psutil.virtual_memory().total
 
+sys.path.append(os.path.join(os.path.dirname(__file__), '.', 'o2dpg_workflow_utils'))
+from o2dpg_workflow_utils import read_workflow
+
 # defining command line options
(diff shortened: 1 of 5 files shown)


benedikt-voelkel:
Examples:
Resources per task (here USS)
uss_tasks

Resources per task in a given category (here digi)
cpu_tasks_digi

If multiple pipeline_metric files are given, overlay resources vs. iteration
pss_vs_iterations

Or stack this per pipeline_metric to see the composition of tasks' resources per iteration (very dense)
uss_history_stacked

sawenzel:
Very nice!

PR #1084, opened 2023-05-26
Some minor changes

sawenzel:
* allow to disable distortion prefetching
* avoiding a problem with broken pandas

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -27,7 +27,11 @@ import json
 import itertools
 import time
 import requests, re
-import pandas as pd
+pandas_available = True
+try:
+    import pandas as pd
+except (ImportError, ValueError):  # ARM architecture has problems with pandas + numpy
+    pandas_available = False
 
 sys.path.append(join(dirname(__file__), '.', 'o2dpg_workflow_utils'))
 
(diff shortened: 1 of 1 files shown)


sawenzel:
@noferini @catalinristea @chiarazampolli : With this, I was able to run a job where TPC distortion corrections are applied in digitization + reco. What needs to be done in the job script is the following:

```
export O2DPG_ENABLE_TPC_DISTORTIONS=ON
SCFile=$PWD/distortions_5kG_lowIR.root # file needs to be downloaded
export O2DPG_TPC_DIGIT_EXTRA=" --distortionType 2 --readSpaceCharge ${SCFile} "
# create workflow as usual
${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow.py ...
```

chiarazampolli:
Hello @sawenzel , 
If possible, it would be very good to do such that we don't need to change the steering script, but we can do everything in the JDL. E.g.:

- O2DPG_ENABLE_TPC_DISTORTIONS should become ALIEN_JDL_O2DPGENABLETPCDISTORTIONS (or better: use ALIEN_JDL_O2DPGENABLETPCDISTORTIONS if defined, otherwise O2DPG_ENABLE_TPC_DISTORTIONS)  
- O2DPG_TPC_DIGIT_EXTRA should become ALIEN_JDL_O2DPGTPCDIGITEXTRA (or better: use ALIEN_JDL_O2DPGTPCDIGITEXTRA if defined, otherwise O2DPG_TPC_DIGIT_EXTRA)
- SCFile should become ALIEN_JDL_SCFILE

In this way, we can set everything from the JDL. @catalinristea , what do you think? Or course we can also change the steering script, but it is more difficult to maintain, in my opinion, when you have more tests, while we anyway need a JDL per production.

Chiara

PR #1086, opened 2023-05-26
Fix IR for low IR, for scaling of TPC maps

chiarazampolli:
@catalinristea


chiarazampolli:
Needed for production, merging.

PR #1087, opened 2023-05-28
Always add meta info (default empty dict)

benedikt-voelkel:
Add also an option to convert pipeline_metric to pandas.DataFrame and dump to JSON for later analysis

### DIFF
--- a/MC/bin/o2dpg_workflow_utils.py
+++ b/MC/bin/o2dpg_workflow_utils.py
@@ -122,9 +122,8 @@ def dump_workflow(workflow, filename, meta=None):
     # make the final dict to be dumped
     to_dump = {"stages": to_dump}
     filename = make_workflow_filename(filename)
-    if meta:
-        to_dump["meta"] = meta
-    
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
@chiarazampolli @catalinristea This fixes an issue where newly introduced meta information are not properly handled e.g. in the case of AnalysisQC. Unfortunately, that broke the AnalysisQC.

gconesab (approved):

chiarazampolli:
Hello @benedikt-voelkel ,

thanks. For my information, which metadata are these?

chiara

benedikt-voelkel:
@chiarazampolli 
For the normal simulation run, all command line options of `o2dpg_sim_workflow.py` are now added to the workflow file and will as well end up in the logged `pipeline_metric_<pid>.log` for later inspection (https://github.com/AliceO2Group/O2DPG/pull/1082).
For AnalysisQC there is so far nothing, it just puts an empty block into the workflow file in that case. But that missing block lead to an issue that was fixed by this PR.

PR #1094, opened 2023-06-02
Forcing MC reco to not apply any correction only if we do not apply distortions in sim

chiarazampolli:
@sawenzel , can you try?
@shahor02 , @catalinristea , @noferini , FYI


sawenzel:
Yes. Works as it should.

PR #1097, opened 2023-06-05
Plot and influx update

benedikt-voelkel:
* add average values to tasks' bar plots (while bar plots show the max as usual)

* add per-task info to file to be sent to influx


chiarazampolli (approved):

chiarazampolli:
Maybe @sawenzel  wants to check too.

PR #1101, opened 2023-06-07
Script for producing overlay plots

DanielSamitz:
Some things still to fix:

- TH2
- use again TRatioPlot if somehow possible
- checks if all files/objects exist
- maximal number of histograms per plot
- return values in python script

Should work also with TTree, but not tested yet

### DIFF
--- a/RelVal/PlotOverlays.C
+++ b/RelVal/PlotOverlays.C
@@ -1,0 +1,151 @@
+#include <string>
+#include <vector>
+#include <filesystem>
+
+// overlay 1D histograms
+void overlay1D(std::vector<TH1*> hVec, std::vector<std::string> labelVec, TLegend* legend, std::string const& outputDir)
+{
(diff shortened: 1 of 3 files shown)


sawenzel:
@benedikt-voelkel : Could you please review?

benedikt-voelkel (approved):
Looks good to me.
As @DanielSamitz indicated, there are still a few small things to do but they are mostly minor. Most important, they do not corrupt RelVal in any way.

DanielSamitz:
The only thing that affects RelVal is that the TH1 ratio plots are now not produced with TRatioPlot, but are "handmade". Should in principle be the same, I will do further testing to make sure that the differences are just in the aesthetics of the plots

benedikt-voelkel:
Yes, sure. But that is fine and I think we could even leave it that way in the future.

PR #1105, opened 2023-06-14
Strangeness tracking in o2dpg_sim_workflow

fmazzasc:
Follows https://alice.its.cern.ch/jira/browse/O2-3830 
 @benedikt-voelkel , what do you think?

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -1271,7 +1271,7 @@ for tf in range(1, NTIMEFRAMES + 1):
 
    #strangeness tracking
    if not args.no_strangeness_tracking:
-      STRACKINGtask = createTask(name='stracking_'+str(tf), needs=[SVFINDERtask['name']], tf=tf, cwd=timeframeworkdir, lab=["RECO"], cpu=NWORKERS, mem='5000')
+      STRACKINGtask = createTask(name='stracking_'+str(tf), needs=[SVFINDERtask['name']], tf=tf, cwd=timeframeworkdir, lab=["RECO"], mem='5000')
       STRACKINGtask['cmd'] = '${O2_ROOT}/bin/o2-strangeness-tracking-workflow '
       STRACKINGtask['cmd'] += ('',' --disable-mc')[args.no_mc_labels]
       STRACKINGtask['cmd'] += getDPL_global_options(bigshm=True) + putConfigValuesNew(['strtracker'], {"NameConf.mDirMatLUT" : ".."})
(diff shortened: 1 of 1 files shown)


MC/bin/o2dpg_sim_workflow.py
```diff
@@ -1268,12 +1269,23 @@ def addQCPerTF(taskName, needs, readerCommand, configFilePath, objectsFile=''):
    SVFINDERtask['cmd'] += ' --vertexing-sources ' + svfinder_sources + (' --combine-source-devices','')[args.no_combine_dpl_devices]
    workflow['stages'].append(SVFINDERtask)
 
+   #strangeness tracking
+   if not args.no_strangeness_tracking:
+      STRACKINGtask = createTask(name='stracking_'+str(tf), needs=[SVFINDERtask['name']], tf=tf, cwd=timeframeworkdir, lab=["RECO"], cpu=NWORKERS, mem='5000')
+      STRACKINGtask['cmd'] = '${O2_ROOT}/bin/o2-strangeness-tracking-workflow '
+      STRACKINGtask['cmd'] += ('',' --disable-mc')[args.no_mc_labels]
+      STRACKINGtask['cmd'] += getDPL_global_options(bigshm=True) + putConfigValuesNew(['strtracker'], {"NameConf.mDirMatLUT" : ".."})
+      workflow['stages'].append(STRACKINGtask)
```
sawenzel, line 1278: this looks correct and is a good start. Do you think this could be directly concatenated to svfinder (as a DPL pipe)? This might be more efficient but we need to see the CPU/mem requirements first.

sawenzel: `cpu=NWORKERS` means that this workflow is ideal parallel and it expected to use all the cores. Is that the case?

fmazzasc: Hi @sawenzel , yes it can be concatenated to the svfinder. How to do that here? For the CPU/mem requirements I attach the pipeline_metrics file. Is that what you need for monitoring the memory?

[pipeline_metric_1154352.log](https://github.com/AliceO2Group/O2DPG/files/11755932/pipeline_metric_1154352.log)

benedikt-voelkel: Thanks a lot for the feedback.
Unfortunately, the metric file is not really representative with 1 signal and 1 background event, simulated with 20 workers.

But I will add something to JIRA.

fmazzasc: Thanks! I also have repeated the same exercise w/ 100 events of signal and bkg and 1 worker. 
[pipeline_metric_1838464.log](https://github.com/AliceO2Group/O2DPG/files/11757310/pipeline_metric_1838464.log)

PR #1108, opened 2023-06-16
PWGDQ: Add config for MC with JPsi (fwdy) - HF (midy) correlations

fgrosa:
@lucamicheletti93 this is the config for the MC simulation with c-cbar and Jpsi at forward rapidity triggers, as well as forced charm-hadron decays


gconesab (approved):

lucamicheletti93:
@fgrosa Thanks a lot! For me these settings are fine.

PR #1109, opened 2023-06-19
Adjust skimming settings for grid jobs

chiarazampolli:
@shahor02 , as discussed.

### DIFF
--- a/DATA/production/configurations/CTFSkimming/ctf-skim-workflow.sh
+++ b/DATA/production/configurations/CTFSkimming/ctf-skim-workflow.sh
@@ -18,11 +18,16 @@ source $GEN_TOPO_MYDIR/getCommonArgs.sh || { echo "getCommonArgs.sh failed" 1>&2
 TIMEFRAME_RATE_LIMIT=2
 : ${NUMAID:="0"}
 
-if [[ ${ALIEN_JDL_CPUCORES:-} == 8 ]]; then
-  export MULTIPLICITY_PROCESS_tpc_entropy_decoder=2
-  export MULTIPLICITY_PROCESS_tpc_entropy_encoder=2
+if [[ ${ALIEN_JDL_CPUCORES:-} == 8 ]] || [[ ${ALIEN_JDL_CPUCORES:-} == 16 ]]; then # 16 is used for jobs on EPNs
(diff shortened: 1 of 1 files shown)


chiarazampolli:
@shahor02 , do you have any comment, or can i merge?

DATA/production/configurations/CTFSkimming/ctf-skim-workflow.sh
```diff
@@ -18,12 +18,15 @@ source $GEN_TOPO_MYDIR/getCommonArgs.sh || { echo "getCommonArgs.sh failed" 1>&2
 TIMEFRAME_RATE_LIMIT=2
 : ${NUMAID:="0"}
 
-if [[ ${ALIEN_JDL_CPUCORES:-} == 8 ]]; then
-  export MULTIPLICITY_PROCESS_tpc_entropy_decoder=2
-  export MULTIPLICITY_PROCESS_tpc_entropy_encoder=2
+if [[ ${ALIEN_JDL_CPUCORES:-} == 8 ]] || [[ ${ALIEN_JDL_CPUCORES:-} == 16 ]; then # 16 is used for jobs on EPNs
+  export MULTIPLICITY_PROCESS_tpc_entropy_decoder=3
+  export MULTIPLICITY_PROCESS_tpc_entropy_encoder=3
+  SHMSIZE=16000000000
```
shahor02, line 24: With 16 you can probably go to twice hight mult.

chiarazampolli: For EPNs, this is a kind of trick. there we have 64 cores (with the current configuration). This is just to be sure that we run the same configuration. 
If we want a more dedicated setting for EPNs, we might use a different env var, also (in case this changes in the future).

shahor02: OK, then it is fine. Though, for experimenting I would foresee a possibility of explicit setting of the MULTIPLICITY_PROCESS_tpc_entropy_decoder from the JDL

chiarazampolli: Ok, I will add this.

chiarazampolli:
Merging.

PR #1115, opened 2023-06-24
MCH: add async post-processing

aferrero2707:
The PR adds the post-processing and checkers for the Digits and Preclusters QC tasks.

The code is the same that is used for the online processing. The PP tasks and the checkers should process the output of the ASYNC QC for each run, producing additional plots in the `MCH/Digits` and `MCH/Preclusters` folders, as well as QualityObjects with the check results in `MCH/QO`

### DIFF
--- a/DATA/production/o2dpg_qc_postproc_workflow.py
+++ b/DATA/production/o2dpg_qc_postproc_workflow.py
@@ -79,6 +79,7 @@ def QC_postprocessing_workflow(runNumber, periodName, passName, qcdbUrl):
   ## The list of QC Post-processing workflows, add the new ones below
   add_QC_postprocessing('example', 'json://${O2DPG_ROOT}/DATA/production/qc-postproc-async/example.json', needs=[], runSpecific=False, periodSpecific=False, passSpecific=True)
   add_QC_postprocessing('EMC', 'json://${O2DPG_ROOT}/DATA/production/qc-postproc-async/emc.json', needs=[], runSpecific=False, periodSpecific=True, passSpecific=True)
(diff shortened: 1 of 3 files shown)


aferrero2707:
@aphecetche @knopers8 could you please review the changes? I have tested them locally and all seems to work as expected, but I have no experience with the ASYNC processing.Thanks!

aphecetche:
LGTM

knopers8 (commented):
Sorry, I missed it before. It look OK, i would only propose to remove "Check" in Check names, as mentioned in the naming convention:
https://github.com/AliceO2Group/QualityControl/blob/master/doc/ModulesDevelopment.md#naming-convention

DATA/production/o2dpg_qc_postproc_workflow.py
```diff
@@ -79,6 +79,7 @@ def add_QC_postprocessing(taskName, qcConfigPath, needs, runSpecific, periodSpec
   ## The list of QC Post-processing workflows, add the new ones below
   add_QC_postprocessing('example', 'json://${O2DPG_ROOT}/DATA/production/qc-postproc-async/example.json', needs=[], runSpecific=False, periodSpecific=False, passSpecific=True)
   add_QC_postprocessing('EMC', 'json://${O2DPG_ROOT}/DATA/production/qc-postproc-async/emc.json', needs=[], runSpecific=False, periodSpecific=True, passSpecific=True)
+  add_QC_postprocessing('MCH', 'json://${O2DPG_ROOT}/DATA/production/qc-postproc-async/mch.json', needs=[], runSpecific=True, periodSpecific=True, passSpecific=True)
```
knopers8, line 82: Am I correct to understand that you expect to produce an output for each run separately? I.e. you are not making a trend across runs?

PR #1116, opened 2023-06-24
Settings for EPN processing with full NUMA but only CPUs

chiarazampolli:


DATA/production/configurations/asyncReco/async_pass.sh
```diff
@@ -382,15 +382,21 @@ else
   # the optimized settings for the 8 core GRID queue without GPU are
   # (overwriting the values above)
   #
-  if [[ "0$ASYNC_PASS_NO_OPTIMIZED_DEFAULTS" != "01" ]]; then
-    export TIMEFRAME_RATE_LIMIT=3
-    if (( $(echo "$RUN_IR > 800000" | bc -l) )); then
-      export TIMEFRAME_RATE_LIMIT=1
-    elif (( $(echo "$RUN_IR < 50000" | bc -l) )); then
-      export TIMEFRAME_RATE_LIMIT=6
+  if [[ "$ALIEN_JDL_CPUEPN" != 1 ]]; then
```
davidrohr, line 385: I'd probably use a clearer variable name here indicating it is 1numa workflow

chiarazampolli: Thanks for the comment, I will change it. 
First I'll see if the job works fine (I don't know why it should not, but...)

chiarazampolli: @davidrohr , can you check now?

davidrohr (commented):
I would leave `0$ASYNC_PASS_NO_OPTIMIZED_DEFAULTS` as outermost if, since it should prevent setting of any optimized settings. Otherwise it seems fine.

davidrohr (approved):

chiarazampolli:
Thanks @davidrohr , I merged since this is not tested in the CI anyway.

PR #1118, opened 2023-06-26
PWGHF: update values in .ini and update generator

ZFederica:
- update argument values in .ini file for Xic and Omegac generator
- update HF generator to trigger either on quark or on hadron

Hello @fgrosa and @ddobrigk, what do you think about the change in the generator?

### DIFF
--- a/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
+++ b/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
@@ -69,7 +69,7 @@ protected:
 
   bool selectEvent(const Pythia8::Event& event)
   {
-    bool isGoodAtPartonLevel = (mHadronPdg != 0) ? true : false, isGoodAtHadronLevel = (mHadronPdg != 0) ? false : true;
+    bool isGoodAtPartonLevel = false, isGoodAtHadronLevel = (mHadronPdg != 0) ? false : true;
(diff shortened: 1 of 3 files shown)


fgrosa:
> * update argument values in .ini file for Xic and Omegac generator
> 
>     * update HF generator to trigger either on quark or on hadron
> 
> 
> Hello @fgrosa and @ddobrigk, what do you think about the change in the generator?

Hi @ZFederica is there a reason to remove the trigger on the quark? If you want both charm and beauty I think that the best would be to have two productions, one with a trigger on beauty quarks + Xic and the other one with a trigger on charm quarks + Xic (otherwise anyway you will have a charm dominated sample and to get a decent statistics for beauty you should simulate much more charm than needed). In general, if there is a good point for having it, it would be good to have an option to turn the trigger on the quark off and not to have it always off when you require an hadron.

ZFederica:
Ok, I removed the change in the generator, now in this PR there is only an update of some values in my .ini files

benedikt-voelkel:
I assume this PR is ready for review/merge, right @ZFederica? However you are saying 

> This is a temporary solution to have this option available, further studies are needed to propose a final version.

so just to be sure: Is there anything else expected here?

ZFederica:
Hi @benedikt-voelkel, yes this PR is ready to be merged, thanks!

benedikt-voelkel (approved):

PR #1119, opened 2023-06-27
Adding a log file with the number of files produced by the skimming

chiarazampolli:
@catalinristea , @mpuccio , @noferini : this extra file will contain the number of skimmed CTFs at the end of the processing. It is not an info important per se, but we can then trigger the validation of this job on the presence of this file instead of o2_ctf*.root, since we might have none of those when the bcRanges selects nothing in the current job.
Let me know if you have feedback!


chiarazampolli:
Hello @mpuccio , @catalinristea , @noferini , 
Do you have any comment about this PR?
Cheers,
Chiara

noferini:
Hi @chiarazampolli ,
no comment. It looks quite straightforward

noferini (approved):

mpuccio:
LGTM, thanks!!

chiarazampolli:
Thanks for the feedback! Merging.

PR #1125, opened 2023-07-04
LHC22c, d, e appear as JUN

chiarazampolli:
@noferini , @catalinristea


chiarazampolli:
@mfasDa , @bazinski , @martenole , I believe that 22cde (which are defined as JUN...) were not correct before. Can you please check this change? We need to merge it ASAP for apass4 of these periods.

Thanks,

Chiara

mfasDa:
Indeed I see no readout trigger in LHC22d which means the CTP emulation was not on.

I hope there won't be any future sample with the name "JUN" ...

martenole (approved):
Sure, if LHCcde appear as JUN this is certainly correct

chiarazampolli:
Thanks for the feedback, merged.

PR #1129, opened 2023-07-07
Modify injector to deal with different PDGs

lbariogl:
The list of PDG codes to be injected is provided as an argument.
Tested locally on injected nuclei and hypernuclei.


gconesab (approved):

lbariogl:
@gconesab @benedikt-voelkel there is something strange in the `sim/o2` tests. The errors here reported could be fake (it says that the function does not exist, but I used it). Is it possible to check, please?

MC/config/PWGLF/ini/GeneratorLFHyperNucleiPbPbGap.ini
```diff
@@ -0,0 +1,7 @@
+[GeneratorExternal]
+fileName=${O2DPG_ROOT}/MC/config/PWGLF/pythia8/generator_pythia8_longlived.C
```
benedikt-voelkel, line 2: This also seems to be the wrong filename still, right?

benedikt-voelkel:
You can also test locally btw.

Say your O2DPG development directory is the ${O2DPG_DEV} and you have loaded the environment:
```bash
O2DPG_TEST_REPO_DIR=${O2DPG_DEV} O2DPG_TEST_HASH_BASE="HEAD~2" ${O2DPG_ROOT}/test/run_generator_tests.sh
```

Note, that you can say how many commits to go back with `O2DPG_TEST_HASH_BASE="HEAD~2" ` If you only want to test the current commit, you can omit this in which case it will go one commit back.
If you have unstaged changes, it compare to `HEAD`

The output is dumped at `o2dpg_tests` by default. If you put `--keep-artifacts`, it will keep all the ROOT files and simulation output in addition to the log files

MC/config/PWGLF/ini/GeneratorLFHyperNucleiPbPbGap.ini
lbariogl: Yes, your are right

lbariogl:
@benedikt-voelkel the errors that I get do not seem to be related to my generators. Locally, on macOS, all tests passed. Is it possible to check, please?

MC/config/PWGLF/ini/tests/GeneratorLFHyperNucleiPbPbGap.C
```diff
@@ -0,0 +1,65 @@
+int External()
+{
+  std::string path{"o2sim_Kine.root"};
+  int numberOfInjectedSignalsPerEvent{5};
+  std::vector<int> injectedPDGs = {1000010020, -1000010020
```
benedikt-voelkel, line 5: there are commas missing

benedikt-voelkel (changes requested):

benedikt-voelkel:
I ran the test locally myself and in fact, it fails as well (seems the same problem as in the CI).

I checked the full log file and it appears that your test macro might have some problems. I started a review (see above) and there are for instance some commas missing.

benedikt-voelkel:
I might update the documentation also a bit. Maybe there are some information missing or something is not clear: https://aliceo2group.github.io/simulation/docs/generators/generatorconfig.html#run-the-test-locally

lbariogl:
Instructions are clear. Tried tests with previous commit on a linux machine and I got the same error as here. Do not understand why there were no errors on macOS. Tried tests with the new commits on linux and now it works:
```All generator tests successful``` (the three that have been changed)

benedikt-voelkel:
Perfect, thanks a lot @lbariogl 
I will wait for the CI to pass and once done, I will merge your developments

benedikt-voelkel (approved):

PR #1133, opened 2023-07-10
[RelVal] Modularies, update, fix

benedikt-voelkel:
* modularise python code
  * entrypoint: o2dpg_release_validation.py
  * general utils: o2dpg_release_validation_utils.py
  * plotting: o2dpg_release_validation_plot.py
  * variables: o2dpg_release_validation_variables.py

* introduce classes RelVal:  reflecting a release validation and handling testing, reading, writing and interfacing to known objects, metrics etx Metric: reflect a metric aka some computation of a number based on 2 compared obejcts Result: reflecting some comparison result of a metric to some threshold, mean+-std region etc.

* update sub-command print allows to simply print information for further processing e.g. if one wants the object names where the interpretation is BAD for the chi2 metric print --path rel_val --object-names --enable-metrics chi2 \ --interpretations BAD

* metrics are all defined in ReleaseValidationMetrics.C to add a metric, more or less 2 line are required to do so

* propagate reasons for why comparison (aka metric calculation) might have failed

* now agnostic about metric names, python simply takes what it finds in the output of ReleaseValidation.C

* adjusted RelVal test

* RelVal for TProfile: do not derive and save ProjectionX but only keep the TProfil in ExtractAndFlatten Calculation of metrics on derived x-projection lead to strange values However, calculation of metrics on TProfile itself works just fine

* ExtractAndFlatten for TEfficiency delete pointers make sure, cloned object is not assigned to a directory, otherwise it will be written twice (==> "efficiency" > 1)

* fix for Kolmogorov Returns 0 if there are no bin errors in either of the histograms. So changed to return non-comaprable

* improve error code propagation between ROOT macros and Python

* remove file size comparison not used and not part of RelVal scope


benedikt-voelkel:
From my side, this is good to be merged.
In particular, it keeps the same basic UI of the `rel-val`, `compare` and `inspect` commands. Only specific arguments have changed.
The output is the same as before and all the metrics and their calculations are the same (modulo bug fixes).
@chiarazampolli @noferini what do you think?
Otherwise we can wait until after the discussion in the coming WP12 Meeting

gconesab (approved):

PR #1141, opened 2023-07-18
Add standardized ways how workflows can embed QC JSONs in the DDS XML

davidrohr:


DATA/production/calib/its-threshold-processing.sh
```diff
@@ -42,6 +43,7 @@ do
 done
 if workflow_has_parameter QC && has_detector_qc ITS; then
   WORKFLOW+="o2-qc --config consul-json://alio2-cr1-hv-con01.cern.ch:8500/o2/components/qc/ANY/any/its-qc-calibration --local --host epn -b $ARGS_ALL | "
+  add_QC_from_consul "/o2/components/qc/ANY/any/its-qc-calibration" "--local --host epn -b"
```
knopers8, line 46: I think I don't understand something here: doesn't it add QC twice to the same workflow?

davidrohr: Thx for spotting! This is obviously a typo....
It still worked since if you have the same workflow twice, DPL takes the latter one...
Fixed here: https://github.com/AliceO2Group/O2DPG/pull/1144

PR #933, opened 2023-03-09
Next-gen gap-triggered

ddobrigk:
Next version of gap-triggering. Changes: 
* now based on 2 differently configured PYTHIA instances inside generator: enables use of userhooks and very different settings for triggered events versus min-bias ones
* greatly simplified code logic 
* now transports all partonic history to the event record too, keeping full mother-daughter relationships from original encapsulated event generators

PR #1142, opened 2023-07-18
Add triggered in general purpose injector generator

njacazio:
based on https://github.com/AliceO2Group/O2DPG/pull/933


njacazio:
@BongHwi could you also have a look for the resonance part?

ChiaraDeMartin95 (commented):
Dear @njacazio, we did the QC for strangeness on a MC sample enriched in omegas with this method and we observe the expected reconstruction efficiency. I think we can proceed and merge this commit.

njacazio:
@benedikt-voelkel 

here we have the gap and indeed we should not have particles for transport in even 1, how could we avoid breaking the CI?
Thanks!

```
1-### Testing /O2DPG/MC/config/PWGLF/ini/GeneratorLFStrangeness.ini with generator External ###
2-
3-Processing /sw/SOURCES/O2DPG-sim-tests/1.0/0/../../../O2DPG/1142-slc7_x86-64/0/test/common/kine_tests/test_generic_kine.C...
4-Event 1 has no particles marked to be transported. Make sure they are marked correctly (see https://aliceo2group.github.io/simulation/docs/generators/).
5:(int) 1
```

benedikt-voelkel (approved):

PR #1143, opened 2023-07-19
TPC laser calib: don't force decoder type

davidrohr:


wiechula:
@davidrohr , I would prefer not to merge this, since this will break laser processing with all versions due to missing trigger handling in your decoder.

davidrohr:
We don't need to merge it now, but we need to do something, since your decoder cannot decode dlbzs.
As you call my decoder page hy page, can you simply handle the triggers of the page and then call my decoder?

Kind Regards
David Rohr

wiechula:
That's what I'm trying to implement, yes.

PR #1152, opened 2023-07-24
TRD PulseHeight input binding

maroozm:
Adding input for `TRD/PULSEHEIGHT` 
https://github.com/AliceO2Group/QualityControl/pull/1886

### DIFF
--- a/DATA/production/qc-sync/trditstpc.json
+++ b/DATA/production/qc-sync/trditstpc.json
@@ -161,7 +161,7 @@
       "id": "trdall",
       "active": "true",
       "machines": [],
-      "query": "digits:TRD/DIGITS/0;tracklets:TRD/TRACKLETS/0;triggers:TRD/TRKTRGRD/0;rawstats:TRD/RAWSTATS/0;tracks:TRD/MATCH_ITSTPC;trigrectrk:TRD/TRGREC_ITSTPC;phValues:TRD/PULSEHEIGHT",
+      "query": "digits:TRD/DIGITS/0;tracklets:TRD/TRACKLETS/0;triggers:TRD/TRKTRGRD/0;rawstats:TRD/RAWSTATS/0;phValues:TRD/PULSEHEIGHT",
       "samplingConditions": [
         {
           "condition": "random",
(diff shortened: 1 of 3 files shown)


martenole:
Thanks, locally the CI works fine for me with this change. But this is only because the async QC is not enabled in our FST. I overlooked this. In order to not find the asynchronous processing crashing at some point could you please also update https://github.com/AliceO2Group/O2DPG/blob/e2d63096aa294e2bee8f13b94d19002b8bc8ad09/DATA/production/qc-async/trditstpc.json#L68 ?
After that we can merge this and your changes to QC

maroozm:
> Thanks, locally the CI works fine for me with this change. But this is only because the async QC is not enabled in our FST. I overlooked this. In order to not find the asynchronous processing crashing at some point could you please also update
> 
> https://github.com/AliceO2Group/O2DPG/blob/e2d63096aa294e2bee8f13b94d19002b8bc8ad09/DATA/production/qc-async/trditstpc.json#L68
> 
> ?
> After that we can merge this and your changes to QC

done :)

DATA/production/qc-sync/trditstpc.json
```diff
@@ -161,7 +161,7 @@
       "id": "trdall",
       "active": "true",
       "machines": [],
-      "query": "digits:TRD/DIGITS/0;tracklets:TRD/TRACKLETS/0;triggers:TRD/TRKTRGRD/0;rawstats:TRD/RAWSTATS/0;tracks:TRD/MATCH_ITSTPC;trigrectrk:TRD/TRGREC_ITSTPC",
+      "query": "digits:TRD/DIGITS/0;tracklets:TRD/TRACKLETS/0;triggers:TRD/TRKTRGRD/0;rawstats:TRD/RAWSTATS/0;tracks:TRD/MATCH_ITSTPC;trigrectrk:TRD/TRGREC_ITSTPC;phValues:TRD/PULSEHEIGHT",
```
martenole, line 164: @maroozm  almost. Here in this line you define the input needed for the PH track match task only. I believe that only needs the phValues and everything else can be removed

maroozm: ohh. yes. i was thinking about any issue it could create because our code is not yet in any QC update.

maroozm: hmm. done :)

martenole (approved):
Thanks! Looks good and tested locally

PR #1153, opened 2023-07-26
Fix data type configuration for trackQC.

mfaggin:


mfaggin:
This PR is to solve an issue that I spotted in the AnalysisQC for the LHC23f apass1, caused by the fact that the configurable `produceFBextendedTable` moved from `bool` to `int` in this PR: https://github.com/AliceO2Group/O2Physics/pull/2930

benedikt-voelkel (approved):

mfaggin:
> Thank you! So should we actually do it for all configurations that use `track-selection`? Seems so, right? In the meantime I will merge this one and open another PR to update potential other configurations.

Hi @benedikt-voelkel , this issue in principle affect all workflows that have the `track-selection` as dependency. The Configurable `produceFBextendedTable` is now an `int`, and the safest thing to do is to set it to `-1` (auto-detection of the table)

PR #1159, opened 2023-08-01
set ft0-eff with energy/coll_system

noferini:
I changed the anchoring script to parse eCM and col to set ft0-eff.
Then the parameters are forwarded to the sim script

Note that in our scripts these two parameters have to be moved from
remainingargs -> baseargs
once this change is in, as shown also in this PR (and --ft0-eff flag should be no longer used)

Please check also syntax since I couldn't test the change


noferini:
Adding @catalinristea

noferini:
I tested the code in all available conditions
pp@13.6 TeV
pp@900 GeV
PbPb

It works as expected. We can merge the PR

chiarazampolli (approved):

PR #1161, opened 2023-08-03
StrangenessTracking called from the SVertexer following O2 PR-11713

shahor02:
@sawenzel @chiarazampolli O2 [PR-11713](https://github.com/AliceO2Group/AliceO2/pull/11713) should fix excessive memory and TBuffer overflow problems in SVertexing we see in PbPb, but the strangeness tracking in this case must be invoked directly from the svertexer. So, this PR should go to the production release only with the corresponding O2 [PR-11713](https://github.com/AliceO2Group/AliceO2/pull/11713) and O2Physics [PR-3259](https://github.com/AliceO2Group/O2Physics/pull/3259)

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -1273,19 +1273,17 @@ for tf in range(1, NTIMEFRAMES + 1):
    # Take None as default, we only add more if nothing from anchorConfig
    svfinder_sources = anchorConfig.get("o2-secondary-vertexing-workflow-options",{}).get("vertexing-sources", None)
    if not svfinder_sources:
-       svfinder_sources = "ITS,ITS-TPC,TPC-TRD,TPC-TOF,ITS-TPC-TRD,ITS-TPC-TOF,ITS-TPC-TRD-TOF"
-       if isActive("MID"):
-           svfinder_sources += ",MID"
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
@shahor02 thanks for this PR!
It seems there are some leftovers like `aodneeds += [ STRACKINGtask['name'] ]` in the workflow.
Let me know and I can take care as well.

shahor02:
@benedikt-voelkel , thanks for spotting! The https://github.com/AliceO2Group/AliceO2/pull/11713 works with current O2Physics.
The  https://github.com/AliceO2Group/O2Physics/pull/3259 is just suppressing using the obsolete V0 constructor (still functional) which I had to preserve in the PR11713 to not break O2Physics PRs CIs, which are built against the previous nightly rather than O2 dev.
Then this constructor will be removed by https://github.com/AliceO2Group/AliceO2/pull/11722.

So, the stable async branch can use PR-11713 only but perhaps it is better to cherry-pick the full set: https://github.com/AliceO2Group/O2Physics/pull/3259 and https://github.com/AliceO2Group/AliceO2/pull/11722 + https://github.com/AliceO2Group/AliceO2/pull/11725 (this latter reshuffles some directories to overcome the inability of our cmake to rebuild ROOT dictionaries dependencies when the dependencies of 2 libraries are swapped. Strictly speaking, it is not needed if O2 is built from scratch).

benedikt-voelkel:
Great thanks!
For the cherry picking I think @chiarazampolli and @noferini should have a look as well.

When the CI is through, I think it should be good to be merged. Any objections anyone?

benedikt-voelkel (approved):

PR #1180, opened 2023-08-21
ITS: push to ccdb test for synthetic/staging calibration scans

iravasen:
Hi @davidrohr, this is to include the feature you suggested in your email few days ago. 
I renamed RUNTYPE to RUNTYPE_ITS to avoid having a variable with the same name of the one used for the real run type set in ECS.


davidrohr (commented):
Thx for the PR:
Just to comment, will you set RUNTTYPE_ITS manually, or where will it come from?
There is no RUNTYPE_ITS variable coming from ECS AFAIK, so it would need to be set via extra env variable?

iravasen:
No, it is just an "O2DPG internal" variable which comes from the standalone-calibration.desc and which is used only to distinguish the various calibration workflows which are defined.

PR #1182, opened 2023-08-21
GenTopo updates

davidrohr:


martenole:
@davidrohr these changes are already at P2, right? But with this we should recreate the GenTopo RPMs and install that to have it permanent?

PR #1183, opened 2023-08-22
Move export of CTP offset outside of dataset block

mfasDa:
Make sure it is applied for any dataset where it is requested.


chiarazampolli (approved):

chiarazampolli:
Just as a comment: 2022 pp data (900 GeV and 13.6 TeV) did not show issue since the shift was applied at analysis.

PR #1187, opened 2023-08-23
WIP: Use --assets in EPN topology merger instead of sed

lkrcal:


davidrohr (approved):
looks good, not merging yet since marked as WIP

PR #1189, opened 2023-08-23
[EMCAL-1037] Include EMCAL BC moniotoring in async QC

mfasDa:


mfasDa:
Needs https://github.com/AliceO2Group/QualityControl/pull/1939

mfasDa:
BC monitoring relevant both in sync and async QC:
- sync QC (configurations in consul): Determine whether EMCAL data and EMCAL triggers in the CTP are in the same BC, determine shifts if present
- async QC (here): Check whether EMCAL data and EMCAL triggers are in the same BC at AOD level (i.e. in case corrections need to be applied)

mfasDa:
Config change tested locally on CTFs from pp this year.

PR #1195, opened 2023-08-27
TRD QC json structure changed

martenole:
Don't merge yet. The json files are not yet in production consul, only on staging. But want to see already now if the CI if OK or not.
@davidrohr or @chiarazampolli I see that the global QC for ITSTPC and PRIMVTX have some additional protection to not be included if the required reco is not configured:
https://github.com/AliceO2Group/O2DPG/blob/91bbc79a2800b24470f5f8672967cb03971dd9cf/DATA/production/qc-workflow.sh#L229-L230
But from what I can see in that case also the `has_matching_qc ITSTPC` should already be false and therefore the check not be needed. Do you remember why this protection was added? Or could it be removed?


davidrohr:
> Don't merge yet. The json files are not yet in production consul, only on staging. But want to see already now if the CI if OK or not. @davidrohr or @chiarazampolli I see that the global QC for ITSTPC and PRIMVTX have some additional protection to not be included if the required reco is not configured:
> 
> https://github.com/AliceO2Group/O2DPG/blob/91bbc79a2800b24470f5f8672967cb03971dd9cf/DATA/production/qc-workflow.sh#L229-L230
> 
> But from what I can see in that case also the `has_matching_qc ITSTPC` should already be false and therefore the check not be needed. Do you remember why this protection was added? Or could it be removed?

Not fully sure any more why it is needed. From what I see, the matching_qc is derrived from WORKFLOW_DETECTORS_MATCHING, and also in dpl-workflow.sh we check `has_detectors_reco ITS TPC && has_detector_matching ITSTPC`. I assume the WORKFLOW_DETECTORS_MATCHING is just set depending on runtype, etc, but does not include the detectors actually present in the run. Thus, `has_matching_qc ITSTPC` can be true even if one of the detectors is missing. It should be interpreted as, do the matching and QC if all inputs are available.
So I think it should stay.

Does it pose any problem?

martenole:
Hi @davidrohr and @chiarazampolli this would be ready for review. Please ignore changes to TRD json files and take a look only at `qc-workflow.sh`. I am allowing to change on the fly during topology generation to change individual json keys in oder to avoid having to use completely different json files depending on which detector is included in the run or not.
This override can be set from outside via `QC_DETECTOR_CONFIG_OVERRIDE` or disabled via `DISABLE_QC_DETECTOR_CONFIG_OVERRIDE=1`.
For the TRD we could even use just a single json file which runs on the QC merger node where all tasks are enabled by default. Some are then simply idling. But on the EPNs we have to disable the tasks which would otherwise try to subscribe to input which is not available and that would fail.

davidrohr (commented):
This looks in principle ok to me.
For the PRIMVTX and ITSTPC matching, I see that you have moved up the checks to the setting of the JSON file names, instead of below for adding the files.

I am actually not sure what is better.
The original approach to have it below had it in one place, which is somewhat cleaner, but it would not allow to have different requirement for online / sync / async, though I am not sure if we need it.

Similarly, for adding the additional QC inputs for the tasks, do we want to set that per online / sync / async, or only once globally?

martenole:
> This looks in principle ok to me. For the PRIMVTX and ITSTPC matching, I see that you have moved up the checks to the setting of the JSON file names, instead of below for adding the files.
> 
> I am actually not sure what is better. The original approach to have it below had it in one place, which is somewhat cleaner, but it would not allow to have different requirement for online / sync / async, though I am not sure if we need it.

Ah right. I had at some point also added `QC_JSON_GLO_ITSTPCTRD` since that would also be automatically supported. And I thought that if we eventually add more I did not want to repeat the check for every task to keep the script as short as possible. Also other detectors e.g. ZDC check their constraints already when setting the `QC_JSON_` variable and not below in the detector QC loop. So while we indeed repeat that check then 3 times its still only always one line for the QC_JSON and the check in my opinion makes sense where one sets the QC_JSON file in the first place.

> Similarly, for adding the additional QC inputs for the tasks, do we want to set that per online / sync / async, or only once globally?

That is a good point and I fear one has to decide from setting to setting. The ones I have added now are fine for all, but maybe some others would like to change only online. I'd say when the need arises I would add up to 4 blocks in the `DETECTOR_CONFIG_OVERRIDE` paragraph. One for global settings and one for online / sync / async. Does that make sense?

davidrohr:
Sounds good to me. I'd prefer to keep as many of these alterations global, and if we need really different behavior for sync/async, we can add an if to the block. There should principally never be a difference for online/sync I would say.

martenole:
Cool! Then I'll merge this as soon as I have updated the TRD json file in CONSUL

davidrohr (approved):

PR #1200, opened 2023-08-29
TRD fix for labelling error in async qc

bazinski:
Add a parameter to the async qc to stop a pile of messages about labels.
This fixes older versions of qc, and newer ones are fixed via : qc pr  #1951


bazinski:
Ah then @martenole I will just delete this pr, did not see #1195

PR #1201, opened 2023-08-29
Add dedicated staging QC global config

martenole:
as discussed in https://alice.its.cern.ch/jira/browse/O2-4069 so far staging could use the same configuration as production, since the database was not used by the worker nodes. Soon the qc task will interface with bookkeeping, so we need to distinguish between staging and prod. Since more than a single key will change I believe its cleaner to have a second json for staging if fine with @davidrohr 

Will merge only after https://github.com/AliceO2Group/O2DPG/pull/1195

### DIFF
--- a/DATA/production/qc-sync/qc-global-epn-staging.json
+++ b/DATA/production/qc-sync/qc-global-epn-staging.json
@@ -1,0 +1,36 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "alio2-cr1-hv-mvs00:8083",
+        "username": "not_applicable",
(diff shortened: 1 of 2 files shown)


davidrohr:
> as discussed in https://alice.its.cern.ch/jira/browse/O2-4069 so far staging could use the same configuration as production, since the database was not used by the worker nodes. Soon the qc task will interface with bookkeeping, so we need to distinguish between staging and prod. Since more than a single key will change I believe its cleaner to have a second json for staging if fine with @davidrohr
> 
> Will merge only after #1195

fine with me

martenole:
Not tested in any case, so merging

PR #1203, opened 2023-08-30
remove full path for o2-aod-merger

noferini:


noferini:
Hi @chiarazampolli , I removed the full path. I think that is not needed and it create the problem with aod-merge
Let's see if the check is passed

PR #1204, opened 2023-08-30
Set default topo merger commands to use installed Python package

lkrcal:


DATA/tools/epn/gen_topo.sh
```diff
@@ -55,7 +51,7 @@ else
   fi
 fi
 # Run stage 2 of GenTopo, which does the PDP part, still from hardcoded updatable RPM path
-/opt/alisw/el8/GenTopo/bin/gen_topo_o2dpg.sh
+/home/lkrcal/alice/O2DPG/DATA/tools/epn/gen_topo_o2dpg.sh
```
davidrohr, line 54: this should not be changed

DATA/tools/epn/gen_topo.sh
```diff
@@ -26,17 +26,13 @@ fi
...
-# Command for topology generation
-if [[ -z "$GEN_TOPO_ODC_EPN_TOPO_CMD" ]]; then
-  export GEN_TOPO_ODC_EPN_TOPO_CMD='/home/lkrcal/epn/topogen/.venv/bin/python3 /home/lkrcal/epn/topogen/topo_merger.py'
-fi
+# Command for topology merging
+export GEN_TOPO_ODC_EPN_TOPO_CMD='/usr/local/bin/epn-topo-merger'
+
 # Command for postprocessing of topology generation after topology caching
-export GEN_TOPO_ODC_EPN_TOPO_POST_CACHING_CMD='/home/lkrcal/epn/topogen/.venv/bin/python3 /home/lkrcal/epn/topogen/topo_resource_alloc.py'
+export GEN_TOPO_ODC_EPN_TOPO_POST_CACHING_CMD='/usr/local/bin/epn-topo-alloc'
```
davidrohr, line 33: Could you please keep the `if [[ -z...` and we should have an `if [[ -z...` for GEN_TOPO_ODC_EPN_TOPO_POST_CACHING_CMD as well.

davidrohr (changes requested):
@martenole : could you help @lkrcal to test this? I'll be unavailable Thursday and Friday, I could help in the tests on Monday again.

davidrohr:
Yes, change in parse makes sense to me.

DATA/tools/epn/gen_topo.sh
lkrcal: I was hoping that since there was no defaulting for GEN_TOPO_ODC_EPN_TOPO_POST_CACHING_CMD, we are finally getting rid of all these env variables. :-)

martenole (approved):

PR #1219, opened 2023-09-06
Enable TRD QC tracking task by default

martenole:
@chiarazampolli this should be what is needed to have at least some tracking QC for async. But I still need to test on staging if this does not break something when there is no TPC-TRD matching enabled

### DIFF
--- a/DATA/production/qc-async/trd.json
+++ b/DATA/production/qc-async/trd.json
@@ -60,7 +60,7 @@
         }
       },
       "Tracking": {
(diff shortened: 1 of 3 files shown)


chiarazampolli:
Hello @martenole , 
Thanks! Should I wait to merge?
Cheers,
Chiara

martenole:
I wanted to make sure that the FST passes and I also tried this on staging now and its fine. Merging

PR #1220, opened 2023-09-07
Update analysis QC for the K0, format json

njacazio:


MC/config/analysis_testing/json/analysis-testing-data.json
```diff
@@ -1,4 +1,50 @@
 {
+  "EMCClusterMonitorTask": {
```
njacazio, line 2: @benedikt-voelkel I formatted the config json as it is the only way to have a meaningful comparison wrt to the one produced by the DPL workflow and to check if it's correctly up 2 date

benedikt-voelkel: That's fine and I think good to do so every now and then. So thanks!

benedikt-voelkel (approved):

PR #1225, opened 2023-09-13
Tpc laser

rmunzer:
Add new option to fix issue with vdrift fitting.


davidrohr:
Seems ok, squash-merging. @rmunzer : perhaps it would make sense if you rebased your branch on O2DPG/master instead of regularly merging. This PR has 22 commits, and most of them are more or less empty merges.

PR #1226, opened 2023-09-13
Make autoset_convertes more flexible

wiechula:
* allow checking files on alien
* allow using also input AOD lists


wiechula:
@njacazio , the present implementation of the autoset_converters is rather restrictive to only be able to use one local AOD as input. In this PR I extended it to allow for files on alien and also input lists of AODs.
Please let me know if this is fine for you.

benedikt-voelkel:
This looks fine to me. Thanks @wiechula

benedikt-voelkel (approved):

PR #1228, opened 2023-09-14
Update its-noise-processing.sh - bug fix

iravasen:


iravasen:
I added also a noise calib workflow which allow to push NoiseMaps to ccdb-test so to not pollute production-ccdb in case tests are needed.

PR #1229, opened 2023-09-14
Add TPC-qa to PIDFull task

giovannimalfattore:


mfaggin:
@chiarazampolli @noferini @benedikt-voelkel this PR adds the PID analysis on AO2Ds in the AnalysisQC. Thanks a lot @giovannimalfattore for the quick implementation. @benedikt-voelkel let us know if everything is fine with you, also considering the mail I sent you this morning.

Thank you.

chiarazampolli:
Hello @benedikt-voelkel , 
Will you take care of this PR?
Cheers,
Chiara

benedikt-voelkel:
Yes @chiarazampolli 

Of course I cannot judge the physics configuration for the analysis but I have one remark:
You are still using the default configuration. But, if you want to be independent, you could think about having your own config files @giovannimalfattore . Just like this one for instance: https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/analysis_testing/json/analysis-testing-TPCSkimming.json

In that case you need to set the paths in the overall configuration correctly: https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/analysis_testing/json/analyses_config.json#L81-L82

Just let me know how you would like to proceed.

Another question: Did you check how much longer the `PIDFull` analysis takes? Does TOC QA introduce significant overhead or is it negligible?

giovannimalfattore:
Hi @***@***.***>,

I did not check the time increase due to the TPC QA  addition cause, as I tested locally. It has taken quite a  time, it already did before adding the tpc qa. Is there a more precise way I can check this amount? To be fair, I’d had to, as TPC QA usually is quite hungry of resources on HL.

By the way, I could try to change the configuration as you suggested.
I just noticed that the PIDFull task was already ready to export all the TOF QA plots, and as my PID macro need only the TPC QA and TOF QA output (as Sign vs pT vs NSigma histograms), possibly in the same output file, I thought that my small change was enough.

Cheers,
Giovanni

From: benedikt-voelkel ***@***.***>
Reply to: AliceO2Group/O2DPG ***@***.***>
Date: Friday, 15 September 2023 at 14:38
To: AliceO2Group/O2DPG ***@***.***>
Cc: Giovanni Malfattore ***@***.***>, Mention ***@***.***>
Subject: Re: [AliceO2Group/O2DPG] Add TPC-qa to PIDFull task (PR #1229)


Yes @chiarazampolli<https://github.com/chiarazampolli>

Of course I cannot judge the physics configuration for the analysis but I have one remark:
You are still using the default configuration. But, if you want to be independent, you could think about having your own config files @giovannimalfattore<https://github.com/giovannimalfattore> . Just like this one for instance: https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/analysis_testing/json/analysis-testing-TPCSkimming.json

In that case you need to set the paths in the overall configuration correctly: https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/analysis_testing/json/analyses_config.json#L81-L82

Just let me know how you would like to proceed.

Another question: Did you check how much longer the PIDFull analysis takes? Does TOC QA introduce significant overhead or is it negligible?

benedikt-voelkel:
Thanks!
Surely, your small change to the config is in principle enough.
My suggestion was because changes on the default configuration might interfere with what you want to have in the configuration.
That I why I suggested to have a dedicated config JSON for PIDFull from now on.

For the potential time overhead: I will test your PR to see If it changes the runtime significantly and all report back here.

benedikt-voelkel:
@giovannimalfattore @chiarazampolli 
I can see that for this single task the additional runtime is around 10-20%.
Given that we have almost 10 different analyses in the AnalysisQC, the effective impact on the entire AnalysisQC should be fine.

But that makes me think that we should benchmark the AnalysisQC maybe at some point to have a better feeling, how changes in the configurations changes the resource needs. I will put this on my list.
But that is another discussion.

benedikt-voelkel:
@TimoWilken 
It seems to me that the CIs are not starting for some reason. Could you maybe check?
Thanks a lot!

TimoWilken:
Hi @benedikt-voelkel, sorry, someone needs to approve the PR before the check will start! This is because @giovannimalfattore is a first-time contributor to this repo.

benedikt-voelkel (approved):

benedikt-voelkel:
Hello @giovannimalfattore 
The CI went through, so from the technical point of view, I'd say that your PR is ready to be merged.
Just let me know how you would like to go about the configuration JSONs. If you don't want to have specific ones for your analysis task, we can go ahead and merge right away.

giovannimalfattore:
@benedikt-voelkel Hi, for me it's already perfectly working, so as the time is limited I would go as it is now

PR #1234, opened 2023-09-19
use FT0 in TOF calibs

noferini:
This allows to use FT0 BC for producing TOF calib infos.
It will be effective once 
https://github.com/AliceO2Group/AliceO2/pull/11887
is in


noferini:
Now, we can merge!

chiarazampolli (approved):

PR #1235, opened 2023-09-20
MID: match recent changes in consul

dstocco:
The QC of the synchronous reconstruction in consul has evolved since a while. This PR is meant to use similar parameters in the async_qc.

### DIFF
--- a/DATA/production/qc-async/mid.json
+++ b/DATA/production/qc-async/mid.json
@@ -34,7 +34,7 @@
         "dataSource": {
           "type": "direct",
           "query": "digits:MID/DATA;digits_rof:MID/DATAROF",
-          "query_comment" : "100% sampling"
+          "query_comment": "100% sampling"
         }
       },
       "MIDClusters": {
@@ -48,7 +48,7 @@
         "dataSource": {
           "type": "direct",
           "query": "clusters:MID/TRACKCLUSTERS;clusterrofs:MID/TRCLUSROFS",
-          "query_comment" : "100% sampling"
+          "query_comment": "100% sampling"
         }
       },
       "MIDTracks": {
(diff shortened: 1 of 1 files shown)


davidrohr:
@chiarazampolli : could you check?

chiarazampolli:
Hello @dstocco , @davidrohr , 
I don't know if these changes make sense in async. Do the checkers work, for example? Pinging @JianLIUhep , @knopers8 .
Chiara

DATA/production/qc-async/mid.json
```diff
@@ -31,10 +31,12 @@
         "detectorName": "MID",
         "cycleDurationSeconds": "60",
         "maxNumberCycles": "-1",
+        "mergerCycleMultiplier": "5",
+        "resetAfterCycles": "1",
```
knopers8, line 35: Indeed this does not make sense in async, it will have no effect. What would you like to achieve by this setting?

knopers8:
Checks work in async QC, they are applied after the merging is done, but before the object upload.

DATA/production/qc-async/mid.json
dstocco: Well, the main purpose of the change was to re-activate one check with the same parameters that we use in sync. I have to say I'm not expert of QC, so I simply copy-pasted the `mergerCycleMultiplier` and `resetAfterCycles`. But If they make no sense in async I will remove them

knopers8 (approved):

chiarazampolli:
Merging, since approved by @knopers8 .

chiarazampolli (approved):

chiarazampolli:
@dstocco : if you need this in async for PbPb, you should add it to:
https://codimd.web.cern.ch/1TXyVM79RkuMkumAwz6xqg#

PR #1236, opened 2023-09-21
Add HF injected jet-jet sim scriptx

aimeric-landou:


gconesab (commented):
I see you are using pT hard bins. I though the strategy for now was to use weighted spectra, has this changed now?

aimeric-landou:
It didn't change, I forgot to add the -weightPow argument. Thank you for noticing! I only managed to get very few events in my local test so I didn't catch it then.

gconesab (commented):
Do we have in place the testing scripts for the CI?

aimeric-landou:
I do not know
Let's put this on hold and we can discuss it when you are back.

benedikt-voelkel:
In the scripts I can see that https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/PWGHF/ini/GeneratorHF_bbbar.ini and https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/PWGHF/ini/GeneratorHF_ccbar.ini are used.

Since these are not changed themselves, the `ini` files are assumed to be fine.
Hence, there will be no test running for the `ini` files.

From my side, the PR is therefore fine and I will approve.

benedikt-voelkel (approved):

PR #1246, opened 2023-09-27
include non-prompt psi(2S) and non-prompt Jpsi in the same production cycle

ffionda:


gconesab (approved):

gconesab (approved):

ffionda:
Ciao @gconesab, thanks. I added the testing macros for the two ini files added in this commit. However, there are also some fixes needed for some of the previously commited ini files. Should I add the corresponding testing macros for those as well ? 
Thanks, Fiorella

benedikt-voelkel:
Hej @ffionda 
You can add any other related `ini` file here I would say. And yes, in that case, the corresponding test macros should all bee there, otherwise the PR will fail.
Which/why macros are tested is explained here: https://aliceo2group.github.io/simulation/docs/generators/generatorconfig.html

gconesab:
Hi @ffionda, do I merge or you want to add something else?

ffionda:
Hi @gconesab, I think you can merge this PR since it is needed for making a MC production request. For older DQ ini files (not related to this PR) I can prepare the corresponding test macros and commit them in a new PR. Thank you, Fiorella

PR #1252, opened 2023-09-29
remove ft0 usage from TOF calib

noferini:
I found that TOF calib infos when using FT0 (recently introduced) are not as expected and cannot be used for calibration.
For the moment we cannot use it. Restore to previous behavior.


chiarazampolli (approved):

chiarazampolli:
Merging, since it is not tested in CI.

noferini:
Thanks!

PR #1253, opened 2023-09-29
Switch EvselQA to PbPb configuration

nburmaso:
Passing correct flags for EventSelectionQA for Pb-Pb runs

### DIFF
--- a/MC/config/analysis_testing/json/analysis-testing-data.json
+++ b/MC/config/analysis_testing/json/analysis-testing-data.json
@@ -124,6 +124,7 @@
   },
   "event-selection-qa-task": {
     "isMC": "false",
+    "isLowFlux" : "false",
     "processRun2": "false",
     "processRun3": "true"
   },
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
Hej @nburmaso 
I think, we should have configurations for PbPb and pp at the same time right? Because AnalysisQC should be runnable for both collision systems at any time.
That would require a bit of work in the AnalysisQC steering scripts which I could do.

nburmaso:
Hi @benedikt-voelkel!

Of course it would be better to have two configurations. Should we wait for the changes in the steering scripts and then add two json files in a different PR?

benedikt-voelkel:
It is currently under discussion in DPG. I will get back here with news, hopefully this afternoon.

nburmaso:
Ok, thank you, Benedikt :)

benedikt-voelkel:
Closing, accounted for by #1255

PR #1258, opened 2023-10-04
FV0: enable amp vs time histos for each channel in async QC

andreasmolander:


chiarazampolli:
Hello @andreasmolander , @JianLIUhep , 
Was this change tested? How larger does the QC file become?
Cheers,
Chiara

JianLIUhep:
Hi @chiarazampolli, @andreasmolander,  I think Andreas tested it and the file size of the newly introduced objects is about 250kB but I do not know how it was done. What I would like to check with Andreas is if the test was carried out with a realistic data rate. Please note empty histograms can be dramatically compressed. Andreas, maybe you can run qc locally with `--local-batch` so that you can check the histograms and the size. 
I quickly calculated the RAM usage for those objects and it should be around 135MB. I think it should be fine on GRID. 
Thank you.
Cheers,
Jian

andreasmolander:
Hi @chiarazampolli @JianLIUhep, I was too quick with that estimate, that is not realistic, hold on a while and I'll check properly.

andreasmolander:
Hi @chiarazampolli @JianLIUhep, sorry for the delay. I tested this with 1000 CTFs from run 543950 (<INEL> (Hz) from ZDC - start of run = 27561.3). With the new histograms the filesize for FV0 QC objects increases from 3.9 MB to 5.6 MB (30% increase). Is this kind of test sufficient? I don't know how it translates to the full async QC processing, but let me know if I can/should test it differently. Thanks!

JianLIUhep:
Hi @andreasmolander, thanks for your tests. I think it should be fine for the 30% increase. Currently, the final QC file size is about 120MB. Adding 3MB more is not a problem. 
@chiarazampolli, I would give the green light to merge this PR. 

Thank you.

chiarazampolli (approved):

chiarazampolli:
Ok, approved and merged, without CI since it is not tested there.
Thanks!

PR #1261, opened 2023-10-05
Changes to async processing

martenole:
@chiarazampolli the newly added TRD calibrations gain and t0 were simply not disabled by default in async... I have added the possibility to do the TRD gain calibration via setting `ALIEN_JDL_DOTRDGAINCALIB=1`. It would be good if this would be done, since we cannot run the gain calibration online since RC did not want to port the fix before HI started.

Also changes for the TPC calibration are in this PR, but in a different commit


chiarazampolli:
Ciao @martenole , 

Thanks! So this is what was missing. Very good, we should just remember that when an online calib is added, also this script is changed. 

I will merge your changes now. 

Cheers,

Chiara

PR #1262, opened 2023-10-05
Enable TRD gain calib ROOT output

martenole:
Hi @chiarazampolli 
sorry, I did not think this through. If we run the gain calibration offline we need to enable the ROOT output, so that we can prepare the manual upload of calibration objects afterwards.
The additional file we'll need to keep is `trd_calibgain.root`.
Cheers,
Ole

### DIFF
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -484,6 +484,7 @@ if [[ $ADD_CALIB == "1" ]]; then
     export SVERTEXING_SOURCES=none # disable secondary vertexing
   fi
   if [[ $ALIEN_JDL_DOTRDGAINCALIB == 1 ]]; then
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Hello @martenole ,

Let's check tomorrow,  the root output is enabled in the main script.

Cheers,

Chiara

martenole:
Hi @chiarazampolli 
but how is the root output enabled? Usually we have `DISABLE_ROOT_OUTPUT=--disable-root-output`  and if we set that to an empty string it means all workflows don't get the `--disable-root-output` option and that way the output is enabled.
For the TRD gain calibration the argument is `--enable-root-output` and this has to be added specifically in order to write the file.
Cheers,
Ole

chiarazampolli:
Hello @martenole , 
We do it here:
https://github.com/AliceO2Group/O2DPG/blob/114105608e15eb5fa5f32fa90c4ab7453582ff15/DATA/production/configurations/asyncReco/async_pass.sh#L237-L276
If, though, TRD needs a different way to enable the root output, than this PR is fine. If you can confirm, I will then merge it.
Cheers,
Chiara

chiarazampolli (approved):

PR #1263, opened 2023-10-06
extend TOF time tolerance in cpass0

noferini:


DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -408,6 +408,10 @@ has_detector FT0 && export ARGS_EXTRA_PROCESS_o2_tpcits_match_workflow="$ARGS_EX
 export ARGS_EXTRA_PROCESS_o2_tof_matcher_workflow="$ARGS_EXTRA_PROCESS_o2_tof_matcher_workflow --output-type matching-info,calib-info --enable-dia --use-fit"
 export CONFIG_EXTRA_PROCESS_o2_tof_matcher_workflow+=";$ITSEXTRAERR;$TRACKTUNETPC;$VDRIFTPARAMOPTION;"
 
+if [ $ALIEN_JDL_LPMPASSNAME == "cpass0" ]; then
```
chiarazampolli, line 411: Ciao @noferini , 
Do you want this only for cpass0? Otherwise you should use something like:
```
if [[ -n $ALIEN_JDL_LPMCPASSMODE ]] && [[ $ALIEN_JDL_LPMCPASSMODE != "-1" ]]; then
  SETTING_ROOT_OUTPUT+="ENABLE_ROOT_OUTPUT_o2_tof_matcher_workflow= "
fi
```

chiarazampolli: Note also that we use `[[ ]]`, instead of single brackets.

noferini: I would keep only for cpass0
For any physics pass I would prefer to get empty matching histos in case of problems

noferini: Let's assume for the moment that extra tolerance are needed only for cpass0 (cpass1 should be in any case better for TPC calibs)
 "[" -> "[[" DONE

DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -408,6 +408,10 @@ has_detector FT0 && export ARGS_EXTRA_PROCESS_o2_tpcits_match_workflow="$ARGS_EX
 export ARGS_EXTRA_PROCESS_o2_tof_matcher_workflow="$ARGS_EXTRA_PROCESS_o2_tof_matcher_workflow --output-type matching-info,calib-info --enable-dia --use-fit"
 export CONFIG_EXTRA_PROCESS_o2_tof_matcher_workflow+=";$ITSEXTRAERR;$TRACKTUNETPC;$VDRIFTPARAMOPTION;"
 
+if [[ $ALIEN_JDL_LPMPASSNAME == "cpass0" ]]; then
+   CONFIG_EXTRA_PROCESS_o2_tof_matcher_workflow+="MatchTOF.nsigmaTimeCut=6;"
```
chiarazampolli, line 412: Can you also add a `;` before `MatchTOF....`?

noferini: Is it needed?
It is there two lines above

noferini: $VDRIFTPARAMOPTION;"

chiarazampolli: I know, but I always add one when there are vars with `+=` in order to avoid messing up passing the arguments. So if you could do it...

noferini: done

chiarazampolli: Thanks!

chiarazampolli (approved):

PR #1270, opened 2023-10-09
Tpc update qc

rmunzer:


DATA/production/qc-async/tpc.json
```diff
@@ -120,6 +120,20 @@
         }
       ],
       "blocking": "false"
+    },
+    {
+      "id": "tpc-tracks",
```
chiarazampolli, line 125: This is not used anywhere, you should change also above, where you have "direct", right? Same for sync json.

davidrohr:
@chiarazampolli : this PR is not yet final, it is not working. Robert will fix it tonight. Then it should indeed go to async reco.

chiarazampolli:
It is probably not working because it is still using "direct", even if the filter is active... 
Please, add the PR to https://codimd.web.cern.ch/1TXyVM79RkuMkumAwz6xqg?both, when ready.

chiarazampolli (approved):

PR #1278, opened 2023-10-13
[AnalysisQC] Enable common args per analysis

benedikt-voelkel:
* use for instance --add-common-args EMCAL-shm-segment-size 2500000000 which will be forwarded as a common argument and appended to the analysis pipeline of EMCAL analysis as --shm-segment-size 2500000000

### DIFF
--- a/MC/analysis_testing/analysis_test.sh
+++ b/MC/analysis_testing/analysis_test.sh
@@ -24,6 +24,7 @@ NTF=$(find ./ -name "tf*" -type d | wc | awk '//{print $1}')
 include_disabled=
 testanalysis=
 aod=
+add_common_args=
 
 if [[ "${#}" == "1" ]] ; then
     # make it backward-compatible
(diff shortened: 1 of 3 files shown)


benedikt-voelkel:
@chiarazampolli 
These are the prerequisites to tackle some of the EMCAL analysis failures in the alibi nightlies.|
For instance, what we need is to increase the SHM for this analysis in certain simulation tasks (many/large events).

In general it gives the flexibility to modify common arguments for single tasks by passing
`--add-common-args <ana-name>-readers 2`
which would change to `--readers 2` instead of using the default `1` for the corresponding analysis `<ana-name>`

Of course, one can give multiple arguments for different analyses like

`--add-common-args <ana-name>-readers 2 <ana-name-other>-shm-segment-size 4000000000 ...`

chiarazampolli (approved):

PR #1279, opened 2023-10-13
Possibility to disable strangeness tracking and cascades from JDL

chiarazampolli:
@shahor02 , @catalinristea


chiarazampolli:
Merged, not tested in CI anyway

PR #1280, opened 2023-10-15
Use alt lumi (TCE) in absence of main one (ZDC)

shahor02:


DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -315,7 +315,14 @@ elif [[ $ALIGNLEVEL == 1 ]]; then
 
   if [[ $ALIEN_JDL_LPMANCHORYEAR == "2023" ]] && [[ $BEAMTYPE == "PbPb" ]]; then
     unset TPC_CORR_SCALING
-    export TPC_CORR_SCALING="--ctp-lumi-factor 2.414 --require-ctp-lumi"
+    if has_detector ZDC ; then
```
chiarazampolli, line 318: Is `has_detector` really enough? That checks only if we want to process ZDC, not if it was read out, right? See https://github.com/AliceO2Group/O2DPG/blob/32be41c3e123b41ecf98148d987f846df2f854a6/DATA/common/gen_topo_helper_functions.sh#L7-L10

shahor02: Sure, but don't you define `WORKFLOW_DETECTORS` according to the detectors present in the readout?
Otherwise, the only way will be to read the GRPECS beforehand.

chiarazampolli: No, we set to everyone. And yes, i would have read GRP...

Cheers,

Chiara

shahor02: ok, will implement this later, when have more time, unless it is urgent to process these runs.

chiarazampolli: It is not urgent, I can also do it tomorrow or so. We should just pay attention that data without ZDC do not go to CPass0, @catalinristea , @noferini , @JianLIUhep (please, inform also Elena).

chiarazampolli:
Hello @shahor02 , 
I think this can be closed, superseeded by https://github.com/AliceO2Group/O2DPG/pull/1289/files
Chiara

PR #1285, opened 2023-10-19
Update Resonance Injector

BongHwi:
Add Triggered injection
Split energy 
Use json inputfile

### DIFF
--- a/MC/config/PWGLF/ini/tests/GeneratorLF_ResonancesTriggered_pp1360.C
+++ b/MC/config/PWGLF/ini/tests/GeneratorLF_ResonancesTriggered_pp1360.C
@@ -1,0 +1,160 @@
+int External()
+{
+    std::string path{"o2sim_Kine.root"};
+    int numberOfInjectedSignalsPerEvent{10};
+    std::vector<int> injectedPDGs = {
+        313,     // K0*0
+        -313,    // K0*0bar
+        323,     // K*+-
+        -323,    // K*bar+-
+        333,     // phi
+        9010221, // f_0(980)
+        113,     // rho(770)0
+        213,     // rho(770)+
+        -213,    // rho(770)bar-
+        3224,    // Sigma(1385)+
+        -3224,   // Sigma(1385)bar-
+        3114,    // Sigma(1385)-
+        -3114,   // Sigma(1385)bar+
+        3124,    // Lambda(1520)0
+        -3124,   // Lambda(1520)0bar
+        3324,    // Xi(1530)0
+        -3324,   // Xi(1530)0bar
+        10323,   // K1(1270)+
+        -10323   // K1(1270)-bar
+    };
+    // TODO: add decay daughters
+    // 9030221, // f_0(1500)
+    // 10331,   // f_0(1710)
+    // 123314,  // Xi(1820)-
+    // -123314, // Xi(1820)+
+    // 123324,  // Xi(1820)0
+    // -123324  // Xi(1820)0bar
+    std::vector<std::vector<int>> decayDaughters = {
+        {321, -211},    // K0*0
+        {-321, 211},  // K0*0bar
+        {311, 211},    // K*+-
+        {-311, -211},  // K*bar+-
+        {321, 321},    // phi
+        {211, -211},    // f_0(980)
+        {211, -211},    // rho(770)0
+        {211, 111},    // rho(770)+
+        {-211, 111},  // rho(770)bar-
+        {3122, 211},   // Sigma(1385)+
+        {-3122, -211}, // Sigma(1385)bar-
+        {3122, -211},   // Sigma(1385)-
+        {-3122, 211}, // Sigma(1385)bar+
+        {2212, -321},   // Lambda(1520)0
+        {-2212, 321}, // Lambda(1520)0bar
+        {3312, 211},   // Xi(1530)0
+        {-3312, -211}, // Xi(1530)0bar
+        {321, 211},   // K1(1270)+
+        {-321, -211}  // K1(1270)-bar
+    };
+    // TODO: add decay daughters
+    // {211, 211},         // f_0(1500)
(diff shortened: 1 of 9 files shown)


MC/config/PWGLF/ini/tests/GeneratorLF_ResonancesTriggered_pp1360.C
```diff
@@ -0,0 +1,160 @@
+int External()
+{
+    std::string path{"o2sim_Kine.root"};
```
njacazio, line 3: If this should be the same, maybe we can use a symbolic link?

BongHwi: I think this is same for all test scripts,
Maybe to @benedikt-voelkel ?

benedikt-voelkel: Sure, if some test macros are identical, you can add symbolic links to minimise redundancy. Of course you then need to commit the links as well and Git will take care of it.

BongHwi: Okay, I changed the type of them. Can we resolve this?

benedikt-voelkel: ok, let's see if the tests work or if there is anything failing with the symlinks.
Let's resolve this once the CI is through or when we know that I have to do something on the CI side

BongHwi:
It looks like the test took too long time...

benedikt-voelkel:
It seems stuck indeed.
@TimoWilken can you see where the problem is? I saw the test hanging before. Is there an issue with the test scripts itself, can I do something to improve?

BongHwi:
The test has been cancelled but not activated again.

BongHwi:
Can we retrigger the test?

TimoWilken:
Hm, this check has re-run twice already in the last 24 hours, and both times it took 10 hours, timed out and was killed.

No other checks in O2DPG (or elsewhere) seem to have that problem, so I'd guess something is wrong with this PR...

BongHwi:
Then, it may relate to the symbolic link we updated the first time.

BongHwi:
It still fails the test... I can't find the clue in the build detail.
```
/sw/BUILD/a34bed5ccbaae4d53136350aac63ee30bf575ef4/O2DPG-sim-tests/o2dpg-sim_tests /sw/BUILD/a34bed5ccbaae4d53136350aac63ee30bf575ef4/O2DPG-sim-tests
++ O2DPG_TEST_GENERATOR_EXITCODE=0
++ O2DPG_TEST_REPO_DIR=/sw/SOURCES/O2DPG-sim-tests/1.0/0/../../../O2DPG/1285-slc7_x86-64/0
++ /sw/slc7_x86-64/O2DPG/1285-slc7_x86-64-local1/test/run_generator_tests.sh
## End sw/BUILD/O2DPG-sim-tests-latest/log
```

Do you have any suggestions?

benedikt-voelkel:
I will run it locally and report back here

benedikt-voelkel:
In my local test I see
```bash
Pythia::next(): 192000 events have been generated 

 Pythia::next(): 193000 events have been generated 

 Pythia::next(): 194000 events have been generated 

 Pythia::next(): 195000 events have been generated 

```
and so on...
It seems it takes forever for some reason. 
Is it correct that you ask Pythia to generate events until you find interesting ones?

BongHwi:
No, I've not applied any option to limit the number of events generated by Pythia.
I think it's the same as I used before.
Did we update the method on the testing part?

benedikt-voelkel:
No, nothing has changed on that side.
I assume it gets caught in a very long loop while trying to find an event that satisfies some conditions: https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/PWGLF/pythia8/generator_pythia8_LF.C#L178-L181

benedikt-voelkel:
You could of course also try to run the tests locally by following this: https://aliceo2group.github.io/simulation/docs/generators/generatorconfig.html#run-the-test-locally

BongHwi:
Okay, thanks a lot, I'll try!

TimoWilken:
Hi @BongHwi, hi @benedikt-voelkel, as of this morning, the CI check for this PR still takes 10 hours (after which it times out and is stopped). This is holding up other pull requests scheduled on the same machine, so I've marked this PR as draft for the time being, to stop it from being re-tested.

Please mark the PR as "ready for review" whenever you think the issue might be fixed! Then it will be tested again.

TimoWilken:
Thank you @benedikt-voelkel! No worries. Maybe we should reduce the overall timeout from 10h to ~4-5h in CI as well, since very few PRs take longer than ~2h30m in normal operation... An extra timeout in O2DPG of a few mins (or however long you deem sensible) would definitely help as well, though!

BongHwi:
Yes, we found that the triggered generation is impossible for the resonance. This made an infinite loop on the production test.
Therefore, we removed the Triggered macros and tests.

Since we introduced the gap production, the production test could fail, but it's not a real issue. It's because we add/test 19 resonances and 100 events is too small for testing all of this injection. It would be great if we had an option to increase the number of events produced for the test.

benedikt-voelkel:
Good news: After the latest commit the CI works fine.
Bad news: The failure this time is genuine

BongHwi:
Yes, the failure was related to the issue that I mentioned.
So I updated the test logic to only check the fraction of the injected events while checking the injection status as the output.

benedikt-voelkel:
No everything looks good. Is there anything to come? Otherwise I will merge it.
Thanks for your patience!

BongHwi:
Yes, please! @benedikt-voelkel

benedikt-voelkel (approved):

PR #1289, opened 2023-10-23
If ZDC is not there, scale with FT0

chiarazampolli:
@shahor02 .
@catalinristea , if this gets merged, then it is needed to process runs from 2023 PbPb without ZDC.

### DIFF
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -206,13 +206,24 @@ if [[ -z $RUN_IR ]] || [[ -z $RUN_DURATION ]] || [[ -z $RUN_BFIELD ]]; then
   export RUN_IR=`cat IR.txt`
   export RUN_DURATION=`cat Duration.txt`
   export RUN_BFIELD=`cat BField.txt`
+  export RUN_DETECTOR_LIST=`cat DetList.txt`
 fi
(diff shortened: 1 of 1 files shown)


DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -206,13 +206,23 @@ if [[ -z $RUN_IR ]] || [[ -z $RUN_DURATION ]] || [[ -z $RUN_BFIELD ]]; then
...
   RUN_IR=1
 fi
-echo "Duration of current run ($RUNNUMBER) = $RUN_DURATION"
+
+# Let's check if ZDC is in the detector list; this is needed for TPC dist correction scaling in PbPb 2023
+SCALE_WITH_ZDC=1
+isZDCinDataTaking=`echo $RUN_DETECTOR_LIST | grep ZDC`
+isFT0inDataTaking=`echo $RUN_DETECTOR_LIST | grep FT0`
+[[ -z $isZDCinDataTaking ]] && SCALE_WITH_ZDC=0
+[[ -z $isFT0inDataTaking ]] && SCALE_WITH_FT0=0
```
shahor02, line 225: looks fine, but should not `SCALE_WITH_FT0` be declared?

chiarazampolli: Yes, I forgot. Thanks!

PR #1290, opened 2023-10-23
Small fix to heavy-quark rapidity check for HF generator

fcatalan92:
Previously the requirement of having a heavy quark with rapidity within the defined window could be satisfied by another parton not composing the heavy-quark pair.

I tested the change generating 20000 b-bbar events with the old and new behavior and requiring a rapidity window between -1 and 1 (the seed is also the same for the two productions).  

The rapidity vs pt distribution for the old rapidity check
RapDistr_oldLogic

The same for the new rapidity check
RapDistr_newLogic

Plotting the difference between the old and the new distributions (old-new), there is an improvement: events with b quarks at very low pT and large rapidity are removed, while new ones with larger pT and closer to the rapidity window are generated.
image

@fgrosa @fcolamar @cterrevo @ZFederica @zhangbiao-phy @mfaggin The difference w.r.t. the previous code is small, but I would merge this since there is an improvement of the generation process.

### DIFF
--- a/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
+++ b/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
@@ -88,12 +88,18 @@ protected:
         auto daughterList = event[iPart].daughterList();
         bool hasQ = false, hasQbar = false, atSelectedY = false;
         for (auto iDau : daughterList) {
-          if (event[iDau].id() == mQuarkPdg)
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
@TimoWilken There was a timeout during the build. Could we re-trigger the test?

fcatalan92:
@TimoWilken Could you restart the test please?

fcatalan92:
@gconesab @sawenzel @benedikt-voelkel @chiarazampolli Could you merge the PR please?

PR #1288, opened 2023-10-23
Update dpgsim

njacazio:

PR #1291, opened 2023-10-23
dpgsim with compatibility with old O2

njacazio:


njacazio:
this should deliver what promised in https://github.com/AliceO2Group/O2DPG/pull/1288

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -1290,7 +1295,10 @@ def addQCPerTF(taskName, needs, readerCommand, configFilePath, objectsFile=''):
    SVFINDERtask['cmd'] += ' --vertexing-sources ' + svfinder_sources + (' --combine-source-devices','')[args.no_combine_dpl_devices]
    # strangeness tracking is now called from the secondary vertexer
    if not args.with_strangeness_tracking:
-      SVFINDERtask['cmd'] += ' --disable-strangeness-tracker'
+      from subprocess import run
+      data = run(SVFINDERtask['cmd'].split(" ")[0] + " --help", capture_output=True, shell=True, text=True)
+      if "disable-strangeness-tracker" in data.stdout:
+         SVFINDERtask['cmd'] += ' --disable-strangeness-tracker'
```
benedikt-voelkel, line 1301: How much is that needed?
Could you not simply move to an older O2DPG version?
It should be anyway safer because other things might have changed as well

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -83,6 +83,7 @@
 parser.add_argument('-colBkg',help='embedding background collision system', default='PbPb')
 
 parser.add_argument('-e',help='simengine', default='TGeant4')
+parser.add_argument('--noGeant',help='prohibits any Geant transport/physics', action='store_true')
```
benedikt-voelkel, line 86: I think it is in principle fine although this option seems to be added because you would like to do some studies? In such a case, all following steps do not make sense.
But if this is a global option here, we might want to include some more logic and not even define the digitisation and reco tasks.
What do you think @sawenzel ?

benedikt-voelkel (changes requested):

MC/bin/o2dpg_sim_workflow.py
njacazio: but then the injection scheme that we plan to use would not be available right?

benedikt-voelkel: Ah, you mean your new devs.
So you mean you would need recent O2DPG but definitely and older O2?
Why can you not use the current O2 versions?

njacazio: Hi @benedikt-voelkel apologies, I was away from work yesterday. Resuming now. We would like to have the most recent injection scheme anchored to the data produced this year with the older O2 tags. We could also ship the injection scheme outside of O2DPG

benedikt-voelkel: To do very custom things, it is also possible (to the best of my knowledge), to have a tarball of `O2DPG` that can be used on the GRID.
There you could have your very own tweaks if needed.
@catalinristea would that be a way to go or is this approach only used in very exceptional cases?

njacazio:
We go via `.tgz`

benedikt-voelkel (approved):

benedikt-voelkel:
I merged that now as I was made aware that this PR is required for some top-priority resource estimations.
We need to follow up on how we incorporate things like this more generically --> TODO list

chiarazampolli:
Thanks, @benedikt-voelkel !

MC/bin/o2dpg_sim_workflow.py
sawenzel: It seems that this approach breaks the GRID MC processing (for some reason it hangs). Therefore, I am in favour of taking back this change. In any case, it is super hard to keep the evolution of O2DPG **not coupled** to O2. 

In this particular case, one could also simply correct the generated workflow.json with some awk/sed command ... or use as @benedikt-voelkel says use custom tarbals for special studies.

PR #1292, opened 2023-10-23
add configuration files for fwd-y productions + add other fixes

ffionda:


benedikt-voelkel:
ok, now it is merged. I am a bit puzzled, it seems no tests were run really.
But probably we will see during another iteration

PR #1294, opened 2023-10-23
Add generator info for each event + add alternate injection of charm and beauty

fcatalan92:
@fgrosa @fcolamar @cterrevo @ZFederica @zhangbiao-phy @mfaggin

### DIFF
--- a/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
+++ b/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
@@ -30,6 +30,14 @@ public:
   ///  Destructor
   ~GeneratorPythia8GapTriggeredHF() = default;
 
+  Bool_t Init() override
+  {
+    addSubGenerator(0, "Minimum bias");
+    addSubGenerator(4, "Charm injected");
+    addSubGenerator(5, "Beauty injected");
+    return o2::eventgen::GeneratorPythia8::Init();
+  }
+
   void addTriggerOnHadron(int hadPdg) { mHadronPdg = hadPdg; };
   void setQuarkTrigger (bool doNoQuarkTrigger) { mDoNoQuarkTrigger = doNoQuarkTrigger; };
   void setQuarkRapidity(float yMin, float yMax)
@@ -56,12 +64,14 @@ protected:
           genOk = selectEvent(mPythia.event);
         }        
       }
+      notifySubGenerator(mQuarkPdg);
     } else {
       // Generate minimum-bias event
       bool genOk = false;
       while (!genOk) {
         genOk = GeneratorPythia8::generateEvent();
       }
+      notifySubGenerator(0);
     }
 
     mGeneratedEvents++;
--- a/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
+++ b/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
@@ -4,6 +4,7 @@
 #include "TRandom.h"
 
 #include <string>
+#include <vector>
 
 using namespace Pythia8;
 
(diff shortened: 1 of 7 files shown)


gconesab (approved):

MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
```diff
@@ -25,11 +25,20 @@ public:
     mHadRapidityMin = -1.5;
     mHadRapidityMax = -1.5;
     mDoNoQuarkTrigger = false;
+    mDoAltInjection = false;
   }
 
   ///  Destructor
   ~GeneratorPythia8GapTriggeredHF() = default;
 
+  Bool_t Init() override
```
fgrosa, line 34:
```suggestion
  bool Init() override
```

MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
```diff
@@ -154,16 +174,32 @@ FairGenerator *GeneratorPythia8GapTriggeredCharm(int inputTriggerRatio, float yQ
...
+    myGen->setQuarkTrigger(doNoQuarkTrigger);
+    myGen->addTriggerOnHadron(pdgCodeCharmHadron);
+    myGen->setHadronRapidity(yHadronMin, yHadronMax);
+  }
+  return myGen;
+}
+
+// Charm and beauty enriched (with same ratio)
+FairGenerator *GeneratorPythia8GapTriggeredCharmAndBeauty(int inputTriggerRatio, float yQuarkMin=-1.5, float yQuarkMax=1.5, float yHadronMin=-1.5, float yHadronMax=1.5, int pdgCodeCharmHadron=0, bool doNoQuarkTrigger=false) {
+  auto myGen = new GeneratorPythia8GapTriggeredHF(inputTriggerRatio, 4);
```
fgrosa, line 193: I don't particularly like the fact that we set the quark pdg to the charm only if we alternate: could we maybe use another code here, or maybe pass a vector of pdg codes (size 1 if no alternated, size 2 if alternated).

fgrosa (changes requested):
Thanks @fcatalan92! I just have a couple of minor comments, see below

MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
fcatalan92: Hi @fgrosa, I didn't get the point here. The new generator I added has no way for the user to change/mess with the alternate injection since it is not exposed in the arguments. The implementation is bare bone but also simple and effective.

Can you clarify or do you have some code suggestions to propose?

fgrosa: The code is indeed correct, I was commenting about the clarity of it: when you construct the generator you set the quark pdg code to 4, so it seems that only charm is set. Keep in mind that a user could directly use the `GeneratorPythia8GapTriggeredHF` and not using the predefined functions that return the generator configured. Hence, I was suggesting either to replace the `mPdgCode` initialisation to a different value (e.g. could be `9=4+5`) to be more explicit, or with a vector of integers (initialised to `{4}` or `{5}` in case of a single flavour and `{4, 5}` in case of charm + beauty. This only requires then a small change (i.e. you have to define an independent variable that is used for the actual quark pdg event by event).

fcatalan92: Ok, I implemented your suggestion

fgrosa: Thanks @fcatalan92!

fcatalan92:
@fgrosa @fcolamar @cterrevo @ZFederica @zhangbiao-phy @mfaggin Please have a look. 
I added also some `ini` files, the configuration is temporary and I will merge it as it is. We will update it  after the planned checks .

zhangbiao-phy:
> @fgrosa @fcolamar @cterrevo @ZFederica @zhangbiao-phy @mfaggin Please have a look. I added also some `ini` files, the configuration is temporary and I would merge it as it is. We will update it after the planned checks.

Hi @fcatalan92, Thanks for this huge work! I have reviewed all the changes, and they are all acceptable to me. Further checks will begin at this stage.

fcolamar:
Ciao @fcatalan92, thanks a lot for this update! I had a look at the modifications (except the new test macros), and evertything looks fine to me.
Fabio

fcatalan92:
@gconesab @sawenzel @benedikt-voelkel @chiarazampolli Could you merge the PR please?

PR #1296, opened 2023-10-24
To enable DPL REPORT PROCESSING from JDL

chiarazampolli:
@shahor02 , @ktf , @davidrohr , @catalinristea


chiarazampolli:
Merged, we want to enable reporting for timing in some test jobs.

PR #1304, opened 2023-10-27
Add track-extra-converter dependency.

mfaggin:
Extra dependency needed for AO2Ds produced with O2 tags not including this PR: [12087](https://github.com/AliceO2Group/AliceO2/pull/12087).

See also comment from David Chinellato in the O2 Analysis Announcements Mattermost channel on 27th October 2023: [link](https://mattermost.web.cern.ch/alice/pl/4esghoynzjyxtqnsgmgsb91dqe)

### DIFF
--- a/MC/analysis_testing/o2dpg_analysis_test_workflow.py
+++ b/MC/analysis_testing/o2dpg_analysis_test_workflow.py
@@ -189,6 +189,7 @@ def get_additional_workflows(input_aod):
         found_O2collision_001 = False
         found_O2zdc_001 = False
         found_O2bc_001 = False
+        found_O2trackextra_001 = False
         for i in froot.GetListOfKeys():
             if "DF_" not in i.GetName():
                 continue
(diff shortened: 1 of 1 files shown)


MC/analysis_testing/o2dpg_analysis_test_workflow.py
```diff
@@ -202,12 +203,16 @@ def get_additional_workflows(input_aod):
                     found_O2zdc_001 = True
                 if "O2bc_001" in j.GetName():
                     found_O2bc_001 = True
+                if "O2trackextra" in j.GetName():
+                    found_O2trackextra_001 = True
```
benedikt-voelkel, line 207: Is this the same logic as for the other tables?
I see you are checking for `O2trackextra` and the you set `found_O2trackextra_001` to `True`.
In the other cases it was looking for the table with `_001` and then setting the corresponding flag.

Could you please explain? The logic is not quite clear to me.

mfaggin: Hi @benedikt-voelkel , indeed I miswrote the table to be checked. If you see in the O2 PR that I linked, we have to check `O2trackextra_001`

MC/analysis_testing/o2dpg_analysis_test_workflow.py
```diff
@@ -202,12 +203,16 @@ def get_additional_workflows(input_aod):
...
                     found_O2bc_001 = True
+                if "O2trackextra_001" in j.GetName():
+                    found_O2trackextra_001 = True 
             if not found_O2collision_001:
                 additional_workflows.append("o2-analysis-collision-converter --doNotSwap")
             if not found_O2zdc_001:
                 additional_workflows.append("o2-analysis-zdc-converter")
             if not found_O2bc_001:
                 additional_workflows.append("o2-analysis-bc-converter")
+            if found_O2trackextra_001:
```
benedikt-voelkel, line 214: And I guess now we need a `not` here as well, right?
Just to be sure :)

mfaggin: .... of course..... thank you very much Benedikt, sorry for the mess

benedikt-voelkel: no worries, thanks for the quick action!

benedikt-voelkel (approved):

PR #1308, opened 2023-10-30
5 threads for sec vtx in case of PbPb with 2 TF in flight

chiarazampolli:
@catalinristea , @shahor02


chiarazampolli:
Not tested, merging.

PR #1314, opened 2023-11-03
MUON: improve SYNC QC for global forward matched tracks

aferrero2707:
Currently the SYNC QC for matched forward tracks is included in the MCH configuration, and limited to MCH-MID tracks. The PR introduces separate QC configuration files for forward matched tracks in SYNC processing, for three different combinations:
- MCH-MID matching (MFT not included)
- MFT-MCH matching (MID not included)
- full MFT-MCH-MID matching: in this case three set of plots are produced, corresponding to the possible matching combinations (MFT-MCH, MCH-MID and MFT-MCH-MID)

### DIFF
--- a/DATA/production/qc-sync/glo-mchmid-mtch-qcmn-epn.json
+++ b/DATA/production/qc-sync/glo-mchmid-mtch-qcmn-epn.json
@@ -1,0 +1,66 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "ali-qcdb.cern.ch:8083",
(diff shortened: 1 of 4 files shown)


davidrohr:
@aferrero2707 : Could you add the respective JSON files to the `qc-sync` folder of O2DPG for the full system test, and add them to the qc-workflow.sh also in case `EPNSYNCMODE=0` `SYNCMODE=1`?

aferrero2707:
@davidrohr done! I have also changed the file names to use the same pattern as for ITSTPC matched tracks.

PR #1315, opened 2023-11-05
Change default setting of mult. axis for trackQC.

mfaggin:
Setting values to the new defaults as in https://github.com/AliceO2Group/O2Physics/commit/505f13ee032701d54664a63580f0bb9a9a371c5a

### DIFF
--- a/MC/config/analysis_testing/json/EventSelectionQA/pbpb/analysis-testing-data.json
+++ b/MC/config/analysis_testing/json/EventSelectionQA/pbpb/analysis-testing-data.json
@@ -296,13 +296,6 @@
         "50"
       ]
     },
-    "binsTrackMultiplcity": {
-      "values": [
-        "5000",
-        "0",
-        "5000"
-      ]
-    },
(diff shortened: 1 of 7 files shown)


benedikt-voelkel:
Would it be a reasonable alternative to remove these specific settings from the config so that the default is take automatically? Then, in the future, people would rather implement specific values if needed.
What do you think @mfaggin ?

mfaggin:
> Would it be a reasonable alternative to remove these specific settings from the config so that the default is take automatically? Then, in the future, people would rather implement specific values if needed. What do you think @mfaggin ?

Hi @benedikt-voelkel , indeed this is how I was thinking these json filesto behave. For some reasons, instead, those default values were copy-pasted. I'll change in the next commit.

benedikt-voelkel:
Great, thank you!
Yes, I agree - whenever there are defaults to be used - we should remove these key-value pairs in the JSONs so that in the future, people edit it only in case they need specific values.

benedikt-voelkel (approved):

PR #1317, opened 2023-11-06
Fixing condition to setup env for EPNs

chiarazampolli:
@catalinristea , anything else missing?


DATA/production/workflow-multiplicities.sh
```diff
@@ -100,7 +100,7 @@ if [[ ! -z ${OPTIMIZED_PARALLEL_ASYNC:-} ]]; then
   elif [[ $OPTIMIZED_PARALLEL_ASYNC == "PbPb_4gpu" ]]; then
     [[ -z $TIMEFRAME_RATE_LIMIT ]] && TIMEFRAME_RATE_LIMIT=20
     [[ -z $SHMSIZE ]] && SHMSIZE=128000000000 # SHM_LIMIT 3/4
-    [[ -z $TIMEFRAME_SHM_LIMIT ]] && TIMEFRAME_SHM_LIMIT=$(($SHMSIZE / 3))
+    [[ -z ${TIMEFRAME_SHM_LIMIT+x} ]] && TIMEFRAME_SHM_LIMIT=$(($SHMSIZE / 3))
```
chiarazampolli, line 103: Hello @davidrohr , 
Please check. Without this, in async we get:
```
/home/zampolli/SOFT/alibuild/ali-o2-dev1/sw/ubuntu2004_x86-64/O2DPG/master-local2/DATA/production/workflow-multiplicities.sh: line 105: TIMEFRAME_SHM_LIMIT: unbound variable
```

chiarazampolli: Hello @davidrohr , 
Is this ok for you?
Chiara

davidrohr: I would prefer `${TIMEFRAME_SHM_LIMIT:-}` to be in line with the other checks for optional variables.
And actually, I think we should do the same for all the checks for `SHMSIZE` and `TIMEFRAME_RATE_LIMIT` in this file. Currently it works in the async reco only since you set it explicitly, but setting the defaults this way would not work.
Can you do this in your PR, or shall I do it?

chiarazampolli: Actually, that is how I implemented the first time, but then it was not working. I believe that I was using the wrong branch by mistake, so let me retry. 
I can then add the others too.

DATA/production/configurations/asyncReco/async_pass.sh
```diff
@@ -353,7 +353,7 @@ fi
 echo "SETTING_ROOT_OUTPUT = $SETTING_ROOT_OUTPUT"
 
 # Enabling GPUs
-if [[ -n "$ALIEN_JDL_USEGPUS" && $ALIEN_JDL_USEGPUS != 0 ]]; then
+if [[ -n "$ALIEN_JDL_USEGPUS" && $ALIEN_JDL_USEGPUS != 0 ]] || [[ $ALIEN_JDL_CPUCORES == 64 ]] ; then   # the fact that we are on EPNs is dictated by the USEGPU or the number of CORES
```
davidrohr, line 356: In principle OK now, but I have 3 comments here:
1. why do we run on the EPNs without GPUs? That will be less efficient. But OK, I agree we should be able to do it for tests.
2. On the MI100 EPNs, I believe we have more than 64 cores in the JDL? So thius case should also be covered here?
3. Don't we have the site name or so as env variable? Would be better to check that instead of such technicalities, which can change, and which at some point might also appear on other sites.

chiarazampolli: You are right. This is wrong, since on EPNs we might not want to use GPUs. 
The reason for this PR was that the reco was submitted on EPNs without the USEGPUS set, and I thought that the idea was now to not have it anymore, since the number of cores defines if we are on EPNs or not, and this we need to have for ML accounting. 
If we want to change the logic, and avoid using "USEGPUS", then we need to refurbish a few things. @catalinristea , what do you prefer?
Chiara

catalinristea: Ciao! Not using "USEGPUS" was accidental. Let's not overcomplicate and leave the code as it is. Thanks!

Yours, Catalin.

chiarazampolli: Done.

PR #1328, opened 2023-11-13
ITS tracker: release tracklets/cells selection for Pb-Pb tests

mconcas:
@chiarazampolli : in run: `544116  @38342 Hz`, these thresholds are beyond the measured distribution.
For instance, the attached image is the dist for `cellsPerClusters`-only, since `trackletPerClusters` never exceeds its threshold.
It is however the case for higher IR, therefore I set the two cuts such as I know they should work.

This should be ready for the tests. I left non-fatal behaviour to allow the processing of the logs to double-check, as I understood this is not yet in production.

CC-ing: @mpuccio @shahor02 

Screenshot 2023-11-13 at 17 49 43


chiarazampolli:
Hello @mconcas , 
Thanks! So now we test like this, and then we will add the FATAL?
Chiara

chiarazampolli:
Perfect, let's do like this for this test.
Thanks!
Chiara

PR #1330, opened 2023-11-16
TPC timeseries: Adding option to store unbinned data

matthias-kleiner:


matthias-kleiner:
Hello @chiarazampolli ,
do you think it is fine like this? The unbinned data can be created with `ALIEN_JDL_ENABLEUNBINNEDTIMESERIES` and the sampling set via `ALIEN_JDL_SAMPLINGFACTORTIMESERIES`

chiarazampolli (commented):
Hello @matthias-kleiner , 

It seems fine to me. This means that you will either have unbinned (which I think is what you usually call "skimmed") or the time series as we have now. Not both. Will the merging that @catalinristea  implemented work in both cases in the same way? 

Chiara

wiechula (approved):

matthias-kleiner:
> Hello @matthias-kleiner ,
> 
> It seems fine to me. This means that you will either have unbinned (which I think is what you usually call "skimmed") or the time series as we have now. Not both. Will the merging that @catalinristea implemented work in both cases in the same way?
> 
> Chiara

No, you will get both output trees. The `--enable-unbinned-root-output` enables only additional output. The original time series will always be produced. For the merging of the new output tree, we would have to implement the same mechanism as for the timeseries

chiarazampolli:
> > Hello @matthias-kleiner ,
> 
> > 
> 
> > It seems fine to me. This means that you will either have unbinned (which I think is what you usually call "skimmed") or the time series as we have now. Not both. Will the merging that @catalinristea implemented work in both cases in the same way?
> 
> > 
> 
> > Chiara
> 
> 
> 
> No, you will get both output trees. The `--enable-unbinned-root-output` enables only additional output. The original time series will always be produced. For the merging of the new output tree, we would have to implement the same mechanism as for the timeseries

Ok, at the moment it is a simple TFileMerger, we can add at the same time the merging of this file.

chiarazampolli (approved):

PR #1333, opened 2023-11-17
Improvements for anchoring

sawenzel:
* improvements in anchoring service (coll system, energy, interaction rate)
* protection agains zero or wrong interaction rates

### DIFF
--- a/MC/bin/o2dpg_sim_workflow_anchored.py
+++ b/MC/bin/o2dpg_sim_workflow_anchored.py
@@ -206,7 +206,14 @@ def retrieve_GRP(ccdbreader, timestamp):
     ts, grp = ccdbreader.fetch(grp_path, "o2::parameters::GRPObject", timestamp = timestamp)
     return grp
 
-def retrieve_MinBias_CTPScaler_Rate(ccdbreader, timestamp, run_number, finaltime, ft0_eff):
+def retrieve_GRPLHCIF(ccdbreader, timestamp):
+    """
+    retrieves the GRPLHCIF object for a given time stamp
+    """
+    _, grplhcif = ccdbreader.fetch("GLO/Config/GRPLHCIF", "o2::parameters::GRPLHCIFData", timestamp = timestamp)
+    return grplhcif
+
+def retrieve_MinBias_CTPScaler_Rate(ccdbreader, timestamp, run_number, finaltime, ft0_eff, NBunches, ColSystem):
     """
     retrieves the CTP scalers object for a given timestamp and run_number
     and calculates the interation rate to be applied in Monte Carlo digitizers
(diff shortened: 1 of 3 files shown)


MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -215,16 +222,26 @@ def retrieve_MinBias_CTPScaler_Rate(ccdbreader, timestamp, run_number, finaltime
...
+      #if ColSystem == "PbPb":
+      #  rate.first = rate.first / 28.
+      #  rate.second = rate.second / 28.
+
+      print("Global rate " + str(rate.first) + " local rate " + str(rate.second))
+      if rate.first >= 0:
+        # calculate true rate (input from Chiara Zampolli) using number of bunches
+        coll_bunches = NBunches
+        mu = - math.log(1. - rate.second / 11245 / coll_bunches) / ft0_eff
+        finalRate = coll_bunches * mu * 11245
```
sawenzel, line 243: @chiarazampolli @noferini : This is better than before. Class 25 seems to get some numbers for PbPb runs. However, we need to carefully review the other numbers and formulas. For the moment the interaction rate seems larger than typical PbPb rates.

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -306,26 +321,47 @@ def main():
...
+    ColSystem = ""
+    if A1 == 82 and A2 == 82:
+      ColSystem = "PbPb"
+    elif A1 == 1 and A2 == 1:
+      ColSystem = "pp"
+    else:
+      print ("Unknown collision system ... exiting")
+      exit (1)
+
+    print ("Collision system ", ColSystem)
```
sawenzel, line 343: @catalinristea : You no longer need to pass collysion system and energy from the outside

chiarazampolli: Hello @sawenzel , 
Should we already foresee pPb? Even if it is not clear when it will happen in Run 3 (when also OO will take place).

MC/bin/o2dpg_sim_workflow_anchored.py
chiarazampolli: Hello @sawenzel , 
Can you give us some examples?

sawenzel: run 544508 : Interaction rate 262397

chiarazampolli: We can take a look. What are you comparing this to?
The IRs in the spreadsheet are integrated over some time (depending on what you use), so some difference is expected since PbPb is not levelled, but it should be reasonable. Unless you are comparing to smth else than the spreadsheet. 
Note that the get-run-parameter (or similar) in O2 should be also updated for PbPb.

noferini: I guess that the rate is obtained assuming a ft0_eff=4 (corr factor for ft0).
If for ZDC we expect a scale factor 28 (do I remember correctly?) you should find an IR 7 time larger (it seems the case).
It could be that we need only to adjust the trigger efficiency in our scripts

sawenzel: @noferini : Might well be. I would like to delegate this to you or someone else.

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -306,26 +321,47 @@ def main():
...
+           if eCM < 1000:
              effT0 = 0.68
-           elif args.eCM < 6000:
+           elif eCM < 6000:
              effT0 = 0.737
            else:
              effT0 = 0.759
-         elif args.col == "PbPb":
+         elif ColSystem == "PbPb":
            effT0 = 4.0
```
noferini, line 360: If we move to ZDC here you have to set 28

MC/bin/o2dpg_sim_workflow_anchored.py
noferini: I added a comment to the line you have to change

MC/bin/o2dpg_sim_workflow_anchored.py
noferini: effT0 = 28.0

MC/bin/o2dpg_sim_workflow_anchored.py
chiarazampolli: The factor 28 is explained here by Roman:

https://alice.its.cern.ch/jira/browse/O2-4315?focusedId=308708&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-308708

PR #1336, opened 2023-11-20
ITS Tracking: enable multi-ROF tracking, disable redundant parameters

mconcas:
Behaviour on the parameters changed in https://github.com/AliceO2Group/AliceO2/pull/12256


mconcas:
> Hello Matteo, For now I would not merge, because we want tests without the multi-ROF tracking, if you agree. Chiara

I understood we decided to test three scenarios with and without multi-rof, to create some references for the comparisons. 
If the tests you refer to are related to the corresponding combinations with no multi-rof, of course this should not be merged. 

Please let me know if I missed something.

PR #1340, opened 2023-11-23
new anchoredMC.sh for nowadays productions

catalinristea:
Ciao,

Please review and amend if found necessary.

Thanks, Catalin.


MC/run/ANCHOR/anchorMC.sh
```diff
@@ -0,0 +1,181 @@
...
+TIMESTAMP=`grep "Determined timestamp to be" timestampsampling_${RUNNUMBER}.log | awk '//{print $6}'`
+echo "TIMESTAMP IS ${TIMESTAMP}"
+
+# -- PREFETCH CCDB OBJECTS TO DISC      --
+# (make sure the right objects at the right timestamp are fetched
+#  until https://alice.its.cern.ch/jira/browse/O2-2852 is fixed)
+export ALICEO2_CCDB_LOCALCACHE=$PWD/.ccdb
+[ ! -d .ccdb ] && mkdir .ccdb
+
+CCDBOBJECTS="/CTP/Calib/OrbitReset /GLO/Config/GRPMagField/ /GLO/Config/GRPLHCIF /ITS/Calib/DeadMap /ITS/Calib/NoiseMap /ITS/Calib/ClusterDictionary /TPC/Calib/PadGainFull /TPC/Calib/TopologyGain /TPC/Calib/TimeGain /TPC/Calib/PadGainResidual /TPC/Config/FEEPad /TOF/Calib/Diagnostic /TOF/Calib/LHCphase /TOF/Calib/FEELIGHT /TOF/Calib/ChannelCalib /MFT/Calib/DeadMap /MFT/Calib/NoiseMap /MFT/Calib/ClusterDictionary /FT0/Calibration/ChannelTimeOffset /FV0/Calibration/ChannelTimeOffset /GLO/GRP/BunchFilling"
```
chiarazampolli, line 131: Hello @sawenzel , @catalinristea , 
Do we still need these?

chiarazampolli: Note also that the list contains non-existing objects, 
```
[ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch///FT0/Calibration/ChannelTimeOffset/1686209727558/
[ERROR] Unable to find object /FT0/Calibration/ChannelTimeOffset/1686209727558, Aborting
[ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch///FV0/Calibration/ChannelTimeOffset/1686209727558/
[ERROR] Unable to find object /FV0/Calibration/ChannelTimeOffset/1686209727558, Aborting
[ERROR] Requested resource does not exist: http://alice-ccdb.cern.ch///GLO/GRP/BunchFilling/1686209727558/
[ERROR] Unable to find object /GLO/GRP/BunchFilling/1686209727558, Aborting
```

sawenzel: These are no longer needed.

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -0,0 +1,181 @@
...
+# get the async script (we need to modify it)
+# the script location can be configured with a JDL option
+
+ALIEN_JDL_LPMPRODUCTIONTAG_KEEP=$ALIEN_JDL_LPMPRODUCTIONTAG
+echo "Substituting ALIEN_JDL_LPMPRODUCTIONTAG=$ALIEN_JDL_LPMPRODUCTIONTAG with ALIEN_JDL_LPMANCHORPRODUCTION=$ALIEN_JDL_LPMANCHORPRODUCTION for simulating reco pass..."
+ALIEN_JDL_LPMPRODUCTIONTAG=$ALIEN_JDL_LPMANCHORPRODUCTION
+
+# ZDC causes issues for sim
+#export ALIEN_JDL_WORKFLOWDETECTORS=ITS,TPC,TOF,FV0,FT0,FDD,MID,MFT,MCH,TRD,EMC,PHS,CPV,HMP,ZDC,CTP
+export ALIEN_JDL_WORKFLOWDETECTORS=ITS,TPC,TOF,FV0,FT0,FDD,MID,MFT,MCH,TRD,EMC,PHS,CPV,HMP,CTP
```
sawenzel, line 34: shouldn't this list be dynamic and come from the data async scripts?
O2DPG already determines the list of detectors (somewhat) dynamically

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -0,0 +1,181 @@
...
+if [[ ! -f setenv_extra.sh ]]; then
+    # cp -v ${DPGRECO%/*}/setenv_extra.sh .
+    cp -v $O2DPG_ROOT/DATA/production/configurations/asyncReco/setenv_extra.sh .
+fi
+
+echo "[INFO alien_async_pass.sh] Setting up DPGRECO to ${DPGRECO}"
+
+#settings that are MC-specific
+sed -i 's/GPU_global.dEdxUseFullGainMap=1;GPU_global.dEdxDisableResidualGainMap=1/GPU_global.dEdxSplineTopologyCorrFile=splines_for_dedx_V1_MC_iter0_PP.root;GPU_global.dEdxDisableTopologyPol=1;GPU_global.dEdxDisableGainMap=1;GPU_global.dEdxDisableResidualGainMap=1;GPU_global.dEdxDisableResidualGain=1/' setenv_extra.sh
+### ???
```
sawenzel, line 55: @chiarazampolli : is this still needed?

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -0,0 +1,181 @@
...
+CCDBOBJECTS_IDEAL_MC="ITS/Calib/Align MFT/Calib/Align"
+TIMESTAMP_IDEAL_MC=1
+${O2_ROOT}/bin/o2-ccdb-downloadccdbfile --host http://alice-ccdb.cern.ch/ -p ${CCDBOBJECTS_IDEAL_MC} -d .ccdb --timestamp ${TIMESTAMP_IDEAL_MC}
+if [ ! "$?" == "0" ]; then
+  echo "Problem during CCDB prefetching of ${CCDBOBJECTS_IDEAL_MC}. Exiting."
+  exit 1
+fi
+
+# ${O2_ROOT}/bin/o2-create-aligned-geometry-workflow --configKeyValues "HBFUtils.startTime=${TIMESTAMP}" --condition-remap=file://${ALICEO2_CCDB_LOCALCACHE}=ITS/Calib/Align,MFT/Calib/Align -b 
+echo "run with echo in pipe" | ${O2_ROOT}/bin/o2-create-aligned-geometry-workflow --configKeyValues "HBFUtils.startTime=${TIMESTAMP}" --condition-remap=file://${ALICEO2_CCDB_LOCALCACHE}=ITS/Calib/Align,MFT/Calib/Align -b
```
sawenzel, line 149: what is this doing??

MC/run/ANCHOR/anchorMC.sh
chiarazampolli: I don't know, maybe @wiechula does? Otherwise I need to check.

MC/run/ANCHOR/anchorMC.sh
chiarazampolli: To avoid overlaps in ITS and MFT. What was observed is that with overlaps hits are missing, see https://indico.cern.ch/event/1339452/contributions/5638180/attachments/2748989/4784254/SCD_11_2023_ITSreport_mconcas.pdf

MC/run/ANCHOR/anchorMC.sh
noferini: I think this is a way to replace data TPC dEdx parameterizations (splines) wtht different ones tuned on MC.
Energy loss in MC is not exactly the same as in data  and gain calibration are not yet ready for MC (if I remember correctly)

wiechula: Is the setenv_extra.sh the default one from asyncReco https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/configurations/asyncReco/setenv_extra.sh ?

benedikt-voelkel: Yes, should be: https://github.com/AliceO2Group/O2DPG/pull/1340/files#diff-ec62e7d374a5f942616197c7eb143f6ad3faba56988ed36c8f0f0e5b49f936e1R48

MC/run/ANCHOR/anchorMC.sh
sawenzel: but I think this can be simply done by pre-fetching trivial CCDB alignment objects for ITS and MFT. I am wondering more why we need to call a special workflow executable for this.

chiarazampolli:
Hello @catalinristea , 
Do you think this PR could be updated with the comments above? I think they have not been taken, yet.
Then we could review it again.
Chiara

benedikt-voelkel:
Are there any news on this?
It would be super helpful if that script could eventually go into O2DPG so there is no need of maintaining local versions.
In addition, it could help a lot to ensure more consistency across different SW builds, e.g. when we have 2023 vs. 2022 simulation SW setups.

MC/run/ANCHOR/anchorMC.sh
catalinristea: Done, thanks!

catalinristea:
Ciao @benedikt-voelkel, all

I've just pushed changes from your tests
https://gitlab.cern.ch/o2simulationtools/alice-dpg-tests/-/blob/main/reco-2022/20231122/anchorMC811.sh?ref_type=heads
I did just a small change: since SETENV is a keyword for (t)csh shells, I've renamed it to DPGSETENV, in case one is using any of those. Please have a look to make sure I haven't done mistakes.

Apart from these last minute changes, for what concerns me, this PR can be merged.

Thank you!

Yours, Catalin.

catalinristea:
...made consistent "_alt" for base/remaining args with explicit set of energy & system.

Thanks,
Catalin.

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -0,0 +1,186 @@
...
+    chmod +x async_pass.sh
+    DPGRECO=./async_pass.sh
+else
+    cp -v $DPGRECO .
+fi
+
+if [[ ! -f setenv_extra.sh ]]; then
+    DPGSETENV=${DPGRECO%/*}/setenv_extra.sh
+    [[ -f ${DPGSETENV} ]] && cp ${DPGSETENV} . || cp ${DPGSETENV} .
+fi
```
benedikt-voelkel, line 50: The logic should be changed to something like
```bash
[[ ! -f setenv_extra.sh ]] && cp ${DPGSETENV} .
```
If `setenv_extra.sh` is there already, nothing needs to be done.
If not, it needs to be copied.

MC/run/ANCHOR/anchorMC.sh
benedikt-voelkel: The original comment should still be finally clarified before merging.

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -0,0 +1,186 @@
...
+CYCLE=${CYCLE:-0}
+# let SEED=$SPLITID+$CYCLE*$PRODSPLIT
+SEED=${ALIEN_PROC_ID}
+
+# create workflow
+# THIS NEEDS TO COME FROM OUTSIDE
+# echo "$" | awk -F' -- ' '{print $1, $3}'
+
+baseargs="-tf ${NTIMEFRAMES} --split-id ${SPLITID} --prod-split ${PRODSPLIT} --cycle ${CYCLE} --run-number ${RUNNUMBER}"
+baseargs_alt="-col pp -eCM 13600 -tf ${NTIMEFRAMES} --split-id ${SPLITID} --prod-split ${PRODSPLIT} --cycle ${CYCLE} --run-number ${RUNNUMBER}"
```
benedikt-voelkel, line 107: This should not be there. I had it in order to test various different versions with the same `anchorMC.sh`

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -0,0 +1,186 @@
...
+
+# THIS NEEDS TO COME FROM OUTSIDE
+remainingargs="-gen pythia8 -proc heavy_ion -seed ${SEED} -ns ${NSIGEVENTS} --include-local-qc --pregenCollContext"
+
+remainingargs="${remainingargs} -e ${SIMENGINE} -j ${NWORKERS}"
+remainingargs="${remainingargs} -productionTag ${ALIEN_JDL_LPMPRODUCTIONTAG:-alibi_anchorTest_tmp}"
+remainingargs="${remainingargs} --anchor-config config-json.json"
+remainingargs_alt="${remainingargs} --anchor-config config-json.json -col pp -eCM 13600"
```
benedikt-voelkel, line 115: See my comment above.

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -0,0 +1,186 @@
...
+
+echo "baseargs: ${baseargs}"
+echo "remainingargs: ${remainingargs}"
+              
+# query CCDB has changed, w/o "_"
+${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow_anchored.py ${baseargs} -- ${remainingargs} &> timestampsampling_${RUNNUMBER}.log
+if [ "$?" != "0" ] ; then
+    ${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow_anchored.py ${baseargs_alt} -- ${remainingargs_alt} &> timestampsampling_${RUNNUMBER}.log
+    [ "$?" != "0" ] && echo "Problem during anchor timestamp sampling " && exit 1
+fi
```
benedikt-voelkel, line 125: See my comment above.

benedikt-voelkel (changes requested):

benedikt-voelkel:
@catalinristea @chiarazampolli @noferini @sawenzel 
I rebased the branch and added some updates. Let me know what you think.
This is the commit message:
```
* prevent from sourcing the script

* fix to find correct setenv_extra.sh

* stop immediately if running async_pass.sh fails
  --> removed the condiction on "NO_MC"

* add some comments

* remove test and development-related code
```

After a final test we should be able to proceed with merging it.

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -0,0 +1,192 @@
...
+TIMESTAMP_IDEAL_MC=1
+${O2_ROOT}/bin/o2-ccdb-downloadccdbfile --host http://alice-ccdb.cern.ch/ -p ${CCDBOBJECTS_IDEAL_MC} -d .ccdb --timestamp ${TIMESTAMP_IDEAL_MC}
+if [ ! "$?" == "0" ]; then
+  echo "Problem during CCDB prefetching of ${CCDBOBJECTS_IDEAL_MC}. Exiting."
+  exit 1
+fi
+
+echo "run with echo in pipe" | ${O2_ROOT}/bin/o2-create-aligned-geometry-workflow --configKeyValues "HBFUtils.startTime=${TIMESTAMP}" --condition-remap=file://${ALICEO2_CCDB_LOCALCACHE}=ITS/Calib/Align,MFT/Calib/Align -b
+mkdir -p $ALICEO2_CCDB_LOCALCACHE/GLO/Config/GeometryAligned
+ln -s -f $PWD/o2sim_geometry-aligned.root $ALICEO2_CCDB_LOCALCACHE/GLO/Config/GeometryAligned/snapshot.root
```
benedikt-voelkel, line 164: I defined `ALICEO2_CCDB_LOCALCACHE` above explicitly now.
The reasons why it might have worked before
1. `ALICEO2_CCDB_LOCALCACHE` was always set in the parent stop of this script
1. It actually failed but the workflow could still run, downloading everything when it was needed
1. It actually failed but the workflow could still run, files are not actually needed

benedikt-voelkel:
I added a few more updates.
Now testing it, `pregencollcontext` complains with
```
the argument ('2932420022') for option '--first-orbit' is invalid
```
First of all, is that a reasonable number? Sorry for my ignorance

If so, we might need to change https://github.com/AliceO2Group/AliceO2/blob/dev/Steer/src/CollisionContextTool.cxx#L48 to `unsigned int` as it is then also used here: https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/common/include/CommonDataFormat/InteractionRecord.h#L44

Also the orbit limit is much more than just a simple `int` would hold: https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/CTP/include/DataFormatsCTP/Scalers.h#L139

@sawenzel @chiarazampolli

chiarazampolli:
Hello @benedikt-voelkel , 

Yes, the max orbit should be unsigned int, see https://github.com/AliceO2Group/AliceO2/blob/652c819713a734b6a415ccda9e677ed562fa5620/Common/Constants/include/CommonConstants/LHCConstants.h#L38
```
constexpr unsigned int MaxNOrbits = 0xffffffff;
```
1 orbit is ~88.9 um, so the number above gives 260'764'266 s, which is anyway too large. Where does it come from?
Who is printing the error?

chiarazampolli:
Apologies, I missed a conversion factor in https://github.com/AliceO2Group/O2DPG/pull/1340#issuecomment-1900492008: the first orbit 2932420022 gives:

2932420022 * 88.9 um = 260'692.1399558 s ~ 3days

So it is reasonable. But we should change the type to unsigned int. 

As discussed with @benedikt-voelkel , the test was run with 22o, for which the anchored MC 23k2b/c did not fail, because we did not use the precoll context.

I believe that what also changed between 2022 and 2023 is that from some point on the orbit reset was issued at every start of fill, and a fill never lasts more that 1 day. So in 2023 we never got out of the int range.

Chiara

benedikt-voelkel:
PR in O2 is here: https://github.com/AliceO2Group/AliceO2/pull/12565

chiarazampolli:
Hello @benedikt-voelkel , 
Thanks! I think we can anyway merge this, since the problem with the IR is there anyway.
Chiara

lietava:
@chiarazampolli : clarifying the orbit reset 22 versus 23 with Anton Jusko in email you sent.

benedikt-voelkel:
Let's wait for this one to be merged: https://github.com/AliceO2Group/O2DPG/pull/1416. Then, we can really merge this one here as well.

benedikt-voelkel (approved):

PR #1344, opened 2023-11-24
Allow dpgsim ro run without geant

njacazio:


MC/bin/o2dpg_sim_workflow.py
```diff
@@ -679,6 +682,8 @@ def getDPL_global_options(bigshm=False, ccdbbackend=True):
                   + ' -o '      + signalprefix   + ' '    + embeddinto                                \
                   + ('', ' --timestamp ' + str(args.timestamp))[args.timestamp!=-1] + ' --run ' + str(args.run)   \
                   + ' --vertexMode kCCDB'
+   if args.noGeant:
+       SGNtask['cmd'] += ' --noGeant'
```
benedikt-voelkel, line 686: Further down the line, if this option would be set, no digi or any other task should be added to the workflow I would say.

sawenzel: I agree. If O2DPG is just used for event generation, then the rest of the processing should be skipped. 
However, in that case ... the user could also just call `o2-sim -g extern ... --noGeant` directly in the script. What would be the additional benefit of using O2DPG (apart from being able to use the workflow_runner ... which would be a fair argument)?

njacazio:
I missed the comment, apologies

PR #1346, opened 2023-11-27
Svertexer configuration changes

ddobrigk:


ddobrigk:
@chiarazampolli @shahor02

DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -427,6 +427,12 @@ fi
 if [[ $ALIEN_JDL_DISABLECASCADES == 1 ]]; then
   export ARGS_EXTRA_PROCESS_o2_secondary_vertexing_workflow+=" --disable-cascade-finder  "
 fi
+# allow usage of TPC-only in svertexer (default: do not)
+if [[ $ALIEN_JDL_USETPCONLY == 1 ]]; then
+  export CONFIG_EXTRA_PROCESS_o2_secondary_vertexing_workflow+=";svertexer.mExcludeTPCtracks=false"
+else 
+  export CONFIG_EXTRA_PROCESS_o2_secondary_vertexing_workflow+=";svertexer.mExcludeTPCtracks=true"
```
shahor02, line 434: @ddobrigk `svertexer.mExcludeTPCtracks` is false by default, I would use your previous version
```
if [[ $ALIEN_JDL_EXCLUDETPCONLY == 1 ]]; then
  export CONFIG_EXTRA_PROCESS_o2_secondary_vertexing_workflow+=";svertexer.mExcludeTPCtracks=true"
fi
```

ddobrigk: Hi Ruben, the point is @chiarazampolli wanted to have the default operation be *true*, without any additional parameter being set (which would be the case in my previous version). If we want that, we either do as I propose or we simply change the default mExcludeTPCtracks in the SVertexerParams.h to true. Any way you and Chiara prefer is fine with me :-D

shahor02: Well, as I wrote by email, we hope to have TPC tracks used by default (which is already the case on the `svertexer.mExcludeTPCtracks` default value level). If the point of Chiara to minimize amount of JDL variable used by default, then we should define this JDL parameter only when we want to disable them, i.e. your previous version.

shahor02:
removed `draft` tag by mistake but leaving it as it is ...

DATA/production/configurations/asyncReco/setenv_extra.sh
chiarazampolli: Hello @shahor02 , @ddobrigk , 

The point is that the the JDL should be able to change things only for testing purposes, so we should leave as default what we want to be our default. This means that for me the default you want in this moment is `svertexer.mExcludeTPCtracks=true"`. 

I will now approve, and then you should decide what you want that becomes our default. The PbPb apass1 and all future reconstructions will run with that without extra JDL variables, which are difficult to maintain. If this means changing the logic, we should do it before Friday.

Cheers,

Chiara

chiarazampolli (approved):

DATA/production/configurations/asyncReco/setenv_extra.sh
chiarazampolli: Hello @shahor02 , 
I understand, but till Felix is ready, the default should be "true". 
When Felix development is ready, and validated, we should change here.

DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -427,6 +427,12 @@ fi
 if [[ $ALIEN_JDL_DISABLECASCADES == 1 ]]; then
   export ARGS_EXTRA_PROCESS_o2_secondary_vertexing_workflow+=" --disable-cascade-finder  "
 fi
+# allow usage of TPC-only in svertexer (default: do not)
+if [[ $ALIEN_JDL_USETPCONLY == 1 ]]; then
```
chiarazampolli, line 431: @ddobrigk , can you make this env var more explicit? E.g. "ALIEN_JDL_USETPCONLYFORV0S" or so?

DATA/production/configurations/asyncReco/setenv_extra.sh
chiarazampolli: My understanding is that till Felix changes are ready, the default is "true", so I am not changing for testing purposes. 
Anyway, I believe that this is a minor thing...

chiarazampolli (approved):

chiarazampolli:
Merging, so that I can tag.

PR #1349, opened 2023-11-28
fix HMP async QC json

chiarazampolli:
@gvolpe79 , @JianLIUhep , @noferini , @catalinristea


chiarazampolli:
Merging, after discussion with @gvolpe79

PR #1350, opened 2023-11-28
Add configs for HF/D2H MC productions triggered on Xic and Omegac

fcatalan92:
@fgrosa @fcolamar @cterrevo @ZFederica @zhangbiao-phy @mfaggin 
I added the remaining `ini` files and test macros for the D2H MC productions. I did not add specific `.cfg` files to decrease the entropy, instead I updated the existing ones.

During the local tests, I noticed that generating the right Pythia event requires a substantial amount of time (not sure if this has impact on the full simulation or not). We should have a look at improving this later on.


ZFederica:
Hi @fcatalan92, thanks for implementing these files! I just have two question, for my understanding:
- didn't we decide to include bothOmegac0->Omega pi and Omegac0->Xi pi? (probably we already discussed this, but I can't remember)
- what is the reason for setting the xic0 BR to this specific value (0.0143)? Do we include also other decay channels for the xic0?

fcatalan92:
Hello @ZFederica
> * didn't we decide to include both Omegac0->Omega pi and Omegac0->Xi pi? (probably we already discussed this, but I can't remember)

From the mail exchange I got that we were going for the Omegac0->Omega pi channel. We could split half/half between Omega pi and Xi pi maybe. Is it fine for you (and the others of course)?

> * what is the reason for setting the xic0 BR to this specific value (0.0143)? Do we include also other decay channels for the xic0?

No particular reason, I set the proper value just because I had the PDG open. This has no effect in the end since we force the decay channel.

benedikt-voelkel:
I will wait with merging this PR until you give me green light. It seems that there is still some discussion ongoing that might need to be resolved if I understand correctly?!

ZFederica:
@fcatalan92 thanks for the clarification. Yes, as for Omegac0 having half/half between Omega pi and Xi pi would be fine

fcatalan92:
> I will wait with merging this PR until you give me green light. It seems that there is still some discussion ongoing that might need to be resolved if I understand correctly?!
Hi @benedikt-voelkel, we agreed on how to proceed. You can merge when tests are green.

benedikt-voelkel (approved):

fcatalan92:
@sawenzel @chiarazampolli @gconesab @benedikt-voelkel Could you merge please?

gconesab (approved):

PR #1352, opened 2023-11-29
[EMCAL-1054] Move EMCAL to digitizer group serving as CTP input

mfasDa:


mfasDa:
Local tests using pythia run 529006 (one timeframe), from BC QC task

Triggers read out by EMCAL:
EMCReadout

EMC min bias-accept accepted by CTP
EMCMinBiasTriggers

The difference between the two are triggers rejected due to FT0, indicating that it gets important to get the 2-stage CTP simulation in, even in min. bias mode.

For the moment I would consider the test as validated and would ask to merge the PR.

PR #1356, opened 2023-11-30
[EMCAL-755] Update EMCAL tasks in MC

mfasDa:
- Add cluster QC
- Rename config to `emc-reco-tasks.json`


mfasDa:
For the moment adding only the cluster task. The BC task would also be appreciated but this is blocked by https://github.com/AliceO2Group/AliceO2/pull/12364

mfasDa:
In order to load the geometry I need to access via the GRPGeometryHelper, otherwise I get problems with the CCDB access of the geometry via the QC CCDB interface. Needs the corresponding PR in QualityControl mentioned above.

mfasDa:
Since the update in the QC is tagged now I would like to ask to merge the PR, in order to be able to include the cluster task in simulations. A second iteration with the BC task will follow.

sawenzel (approved):

PR #1358, opened 2023-11-30
Align SOR and first orbit with CTPScalers

benedikt-voelkel:


benedikt-voelkel:
@sawenzel @chiarazampolli @noferini 
This is a proposal to go together with https://github.com/AliceO2Group/AliceO2/pull/12368

benedikt-voelkel:
See discussion at https://github.com/AliceO2Group/AliceO2/pull/12368

In short:
* Usually `SOX > SOR`
* From `RCT/Info?RunInformation` we get the `SOR` which is often smaller than what we have in the first `CTPScalerRecordO2::epochTime`
In order to not be out of bounds, let's take the maximum `max(for_scalers, GLOparams["SOR"]`, same for first orbit

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -311,14 +319,29 @@ def main():
...
+    ctp_scalers = retrieve_CTPScalers(ccdbreader, timestamp, args.run_number)
+    if ctp_scalers is None:
+      print(f"ERROR: Cannot retrive scalers for run number {args.run_number} and timestamp {timestamp}")
+      exit (1)
+
+    first_orbit_scalers = ctp_scalers.getOrbitLimit().first
+    # this comes in microseconds
+    sor_scalers = ctp_scalers.getTimeLimit().first
+
+    final_first_orbit = max(first_orbit_scalers, GLOparams["FirstOrbit"]
```
noferini, line 334: ")" missing at the end of the line

noferini: line 334

benedikt-voelkel: fixed

chiarazampolli:
Hello @benedikt-voelkel ,
I propose to merge this PR, and fix the issue in another, what do you think?
Is there a way to define if the production is anchored? Maybe from the run number? So that if it is not, we have a default of 650 kHz for pp and maybe 25 kHz PbPb?
@noferini .
Chiara

benedikt-voelkel:
I guess we can check. In case of an anchored run, `o2dpg_sim_workflow.py` gets an `--anchored-config` from the outside: https://github.com/AliceO2Group/O2DPG/blob/master/MC/bin/o2dpg_sim_workflow.py#L53

If that is empty, we could assume an unanchored run. But I would make it more robust since *in principle* anyone could pass such a config from the outside without attempting a real anchored simulation but just to pass their custom configuration.

@chiarazampolli From my side this would be fine to be merged.
@noferini Did you do a final test with your scripts? You saw that previously that you had problems towards the end of `SPLITID`s. So if that works now with this development, it would give us some good feedback.

PR #1359, opened 2023-12-01
[EMCAL-1054] Enable BC task in simulations

mfasDa:
Include EMCAL BC QC in simulations in case
CTP is included.

### DIFF
--- a/DATA/production/qc-async/emc_PbPb.json
+++ b/DATA/production/qc-async/emc_PbPb.json
@@ -100,7 +100,7 @@
           "query": "emcal-triggers:EMC/CELLSTRGR/0;ctp-digits:CTP/DIGITS;ctp-config:CTP/CONFIG/0?lifetime=condition&ccdb-path=CTP/Config/Config&ccdb-run-dependent=true"
         },
         "taskParameters": {
-          "AliasMB" : "CMTVXTSC"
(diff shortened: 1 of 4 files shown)


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/12364, therefore handled as separate PR.

mfasDa:
Results from local tests shown in https://github.com/AliceO2Group/O2DPG/pull/1352

mfasDa:
Needs https://github.com/AliceO2Group/QualityControl/pull/2060 in order to provide different trigger class names for min. bias triggers, or to allow an or of 2 min. bias trigger classes at the same time.

Reasons:
- Trigger class names were changing in time, we need to be able to allow different classes and match the with the corresponding trigger configuration of that run
- In pp we used 2 trigger classes as min. bias trigger (C0TVXTSC,C0TVXTCE), we need to be able to accept triggers if either of them fires.

chiarazampolli (approved):

chiarazampolli:
Hello @benedikt-voelkel , @noferini , 
The CI is full of 
```
8-Background generator seed:  825338745
9-Timeframe 1 seed:  825338746
10-o2dpg_sim_workflow: Error! Interaction rate not >0 !!!
11:[FATAL]: O2DPG_TEST Workflow creation failed
```
I am merging this one now. 
Chiara

PR #1363, opened 2023-12-04
TPC residual extraction with ITS-TPC tracks only

martenole:
Hi @chiarazampolli 
for the release validation I understood from Alex that the TPC residuals will be extracted for one full run, right? In that case it would be good to only use ITS-TPC tracks for the map, because otherwise we will create a bias to tracks matched to TRD/TOF which leads to an efficiency loss in some regions (PHOS hole).
Sorry for the late update.


chiarazampolli:
Hello @martenole , 
This will then be used for every production, is it ok? 
Chiara

chiarazampolli:
As discussed in MM, it is ok that this is now the default. Merging.

PR #1364, opened 2023-12-05
Fix ft0reco crash

sawenzel:
Recent disabling of FT0 calibrations in the reco workflow internally disables the CCDB fetcher. In result, the option --condition-not-before vanishes and the workflow crashes when the option is given on the command line.

This is a generic problem with "dynamic" options.

For now fixing, by taking away this option for FT0 reconstruction.


sawenzel:
Opened a JIRA ticket https://alice.its.cern.ch/jira/browse/O2-4467 concerning the underlying problem in O2

PR #1365, opened 2023-12-05
fix ft0 ctp input for pp case

noferini:
Dear @benedikt-voelkel , @chiarazampolli , @sawenzel , @shahor02 , @gconesab,
this is to recover the right behaviour in pp anchored MC (temporary solution) since ctpinput=0 is invalid.
Final solution has to be discussed with Roman


benedikt-voelkel:
I guess that makes sense for now.

benedikt-voelkel (approved):

sawenzel (approved):

PR #1370, opened 2023-12-07
Update ARGS_ALL_CONFIG default and example workflow for standalone detector workflows

martenole:
Remove obsolete option from detector workflows.

Some FT0, FV0 and TPC workflows don't yet use the common `add_W` functionality and they should be adapted in case the workflows are still used. Could you please take a look @rmunzer and @afurs ? Let me know in case you need any help.


davidrohr (approved):
thx!

PR #1371, opened 2023-12-07
dpl-workflow: tune 4gpu pbpb multiplicities

davidrohr:


davidrohr:
@chiarazampolli : with the additional commits:
- I changed the defaults for pbpb 4gpu to shmsize 100g and rate limit 30, as proposed in the mail.
- I fixed some white space / tabs
- I added a new option OPTIMIZED_PARALLEL_ASYNC_AUTO_SHM_LIMIT to use the proposed shmsize and rate limit from workflow-multiplicities.sh, so we don't need to keep them in sync manually.
- I added the 50% larger shmsize and tf rate limit for MI100 nodes. This we have for sync reco, but for async it was not working, so we were not using the additional compute power of the MI100s :(.

chiarazampolli:
Thanks @davidrohr !
I also think that the low eff on MI100 was due to too few TFs in flight. 
Do you want to wait for the CI?
Chiara

PR #1375, opened 2023-12-08
[FOCAL-24] Add trigger for particle in FOCAL acceptance

mfasDa:


benedikt-voelkel:
@gconesab Is there a reason why this was merged before the CI went through?
Is that an urgent PR?

gconesab:
No, not really, I clicked too fast

benedikt-voelkel:
Let's keep an eye on it, to my understanding, it might still be running at some point :)

PR #1378, opened 2023-12-08
Use corrected MFT geometry and apply non-ideal alignment

mcoquet642:
Also removing deprecated option for MFT tracking parameters

### DIFF
--- a/MC/run/ANCHOR/2021/OCT/pass4/anchorMC.sh
+++ b/MC/run/ANCHOR/2021/OCT/pass4/anchorMC.sh
@@ -104,7 +104,7 @@ baseargs="-tf ${NTIMEFRAMES} --split-id ${ALIEN_JDL_SPLITID:-1} --prod-split ${A
 # THIS NEEDS TO COME FROM OUTSIDE
 remainingargs="-gen pythia8 -proc cdiff -ns ${NSIGEVENTS}                                                                                                                 \
                -interactionRate ${INTERACTIONRATE}                                                                                                                        \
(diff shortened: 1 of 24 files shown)


MC/run/ANCHOR/2021/OCT/pass4/anchorMC.sh
```diff
@@ -104,8 +104,8 @@ baseargs="-tf ${NTIMEFRAMES} --split-id ${ALIEN_JDL_SPLITID:-1} --prod-split ${A
 # THIS NEEDS TO COME FROM OUTSIDE
 remainingargs="-gen pythia8 -proc cdiff -ns ${NSIGEVENTS}                                                                                                                 \
                -interactionRate ${INTERACTIONRATE}                                                                                                                        \
-               -confKey \"Diamond.width[2]=6.0;Diamond.width[0]=0.01;Diamond.width[1]=0.01;Diamond.position[0]=0.0;Diamond.position[1]=-0.035;Diamond.position[2]=0.41\"  \
-              --include-local-qc --include-analysis --mft-reco-full"
+               -confKey \"Diamond.width[2]=6.0;Diamond.width[0]=0.01;Diamond.width[1]=0.01;Diamond.position[0]=0.0;Diamond.position[1]=-0.035;Diamond.position[2]=0.41;MFTBase.buildAlignment=true\"  \
```
sawenzel, line 107: I am not sure this is the best place. It might make more sense to switch this option to "on" directly in the O2 default code. (In any case one could always fetch ideal alignment CCDB ... whenever we don't want to have alignment). Otherwise, the better place is in `o2dpg_sim_workflow.py`

sawenzel:
Part of this is now already in place. You are still welcome to reopen a PR concering simplification of reco options.

PR #1387, opened 2023-12-14
Env var starting with SPLIT are ignored by JDL parser

chiarazampolli:
@costing @catalinristea


chiarazampolli:
Trivial, merged.

PR #1398, opened 2023-12-21
Fix for HF MC configurations + example of script for anchored production

fcatalan92:


gconesab (approved):

fgrosa:
Thanks @fcatalan92! All good from HF side, it can be merged for us.

PR #1406, opened 2024-01-12
Make TPC reco sensitive to TPCCorrMap param

sawenzel:


sawenzel:
@shahor02 I guess this is one step to fix the TPC distortion map problem.

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -1163,7 +1165,7 @@ def getDigiTaskName(det):
 
    PVFINDERtask = createTask(name='pvfinder_'+str(tf), needs=pvfinderneeds, tf=tf, cwd=timeframeworkdir, lab=["RECO"], cpu=NWORKERS, mem='4000')
    PVFINDERtask['cmd'] = '${O2_ROOT}/bin/o2-primary-vertexing-workflow ' \
-                         + getDPL_global_options() + putConfigValuesNew(['ITSAlpideParam','MFTAlpideParam', 'pvertexer', 'TPCGasParam'], {"NameConf.mDirMatLUT" : ".."})
+                         + getDPL_global_options() + putConfigValuesNew(['ITSAlpideParam','MFTAlpideParam', 'pvertexer', 'TPCGasParam', 'TPCCorrMap'], {"NameConf.mDirMatLUT" : ".."})
```
shahor02, line 1168: Note that also SVertexer reads this object from the CCDB though at the moment does not use it. Could you change https://github.com/AliceO2Group/O2DPG/blob/181535bdaa01177b3585d222e0d4341f7625b523/MC/bin/o2dpg_sim_workflow.py#L1324C4-L1324C29 to 
```
   SVFINDERtask['cmd'] += getDPL_global_options(bigshm=True) + svfinder_threads + putConfigValuesNew(['svertexer', 'TPCCorrMap'], {"NameConf.mDirMatLUT" : ".."})
```
?

shahor02 (commented):
Thanks, see comment below.
Note that the `--corrmap-lumi-inst` and `--corrmap-lumi-mean` options checked in the 
https://github.com/AliceO2Group/O2DPG/blob/5ce0df9e8b52975d21be82d8b0c858ca86a67535/UTILS/parse-async-WorkflowConfig.py#L190-L195 are gone, while on on top of the `--lumi-type` checked at https://github.com/AliceO2Group/O2DPG/blob/5ce0df9e8b52975d21be82d8b0c858ca86a67535/UTILS/parse-async-WorkflowConfig.py#L198C29-L198C40 we got also an option `--corrmap-lumi-mode <int>` which are applied to the same workflows as the `TPCCorrMap` configurable you added in this PR.

chiarazampolli (approved):

sawenzel:
Should be good to go. Tested in for run number 544510 LHC23k6c.

PR #1408, opened 2024-01-15
track cut tuning for TOF match QC

noferini:
@ercolessi @njacazio 

this is for the next Pb-Pb apass2


chiarazampolli:
Hello @noferini , 

This will impact all QC, not just PbPb. Is this correct? I am asking explicitly due to your comment. There is a way to make parameters beam type dependent, if you need. 

Chiara

noferini:
It is fine for all collision systems. I explicitly mentioned Pb-Pb just to stress that it is needed for the today tag.

PR #1409, opened 2024-01-16
Update mid.json

ValerieRamillien:
proposed configuration changes

### DIFF
--- a/DATA/production/qc-async/mid.json
+++ b/DATA/production/qc-async/mid.json
@@ -3,23 +3,18 @@
     "config": {
       "database": {
         "implementation": "CCDB",
-        "host": "ali-qcdb.cern.ch:8083",
-        "username": "not_applicable",
-        "password": "not_applicable",
-        "name": "not_applicable"
+            "host": "ccdb-test.cern.ch:8080"
       },
       "Activity": {
-        "number": "42",
-        "type": "2"
       },
       "monitoring": {
-        "url": "influxdb-unix:///tmp/telegraf.sock"
+        "url": "infologger:///debug?qc"
       },
       "consul": {
-        "url": "null:8500"
+        "url": ""
       },
       "conditionDB": {
-        "url": "null:8083"
+        "url": "ccdb-test.cern.ch:8080"
       }
     },
     "tasks": {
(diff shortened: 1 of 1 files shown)


chiarazampolli (commented):
Hello @ValerieRamillien ,

What are the changes aimed at? Maybe i am overlooking, but the only relevant ones seems the MaxCycleDuration. The others are re-written by a common setup. Am i wrong?

@JianLIUhep , FYI.

JianLIUhep:
Hello @chiarazampolli, I think you are right. To me, nothing will be changed with this PR for async. 
It seems to be a cleanup rather than a change. But let's see what @ValerieRamillien says.

ValerieRamillien:
Yes it is only a cleanup,  proposed by QC expert.

DATA/production/qc-async/mid.json
```diff
@@ -3,23 +3,18 @@
     "config": {
       "database": {
         "implementation": "CCDB",
-        "host": "ali-qcdb.cern.ch:8083",
-        "username": "not_applicable",
-        "password": "not_applicable",
-        "name": "not_applicable"
+            "host": "ccdb-test.cern.ch:8080"
```
benedikt-voelkel, line 6: I cannot immediately see what happened here, but it doesn't seem to be a normal comma at the end of this line!? Not sure, maybe with your editor some special character was inserted here. The same is true for some other lines below.

ValerieRamillien: yes I suppose that it comes from my editor it is not normal.
How can I remove this?

benedikt-voelkel: What editor are yo using? Are you using any special encoding?

DATA/production/qc-async/mid.json
```diff
@@ -3,23 +3,18 @@
     "config": {
       "database": {
         "implementation": "CCDB",
-        "host": "ali-qcdb.cern.ch:8083",
-        "username": "not_applicable",
-        "password": "not_applicable",
-        "name": "not_applicable"
+            "host": "ccdb-test.cern.ch:8080"
```
benedikt-voelkel, line 6: Why use the test instance here (and also below in line 17)?
Shouldn't it be the production instance `alice-ccdb`?

knopers8: These were recommendations advised by me and Barth during QC module reviews we did recently. We suggested that test values are used in there, so that when testing, the results are not stored in the production database. When running on grid, these values are anyway overwritten to prod.

knopers8: also, prod url for QCDB is not `alice-ccdb`, we store QC objects in a separate CCDB instance.

benedikt-voelkel:
Since I am not a QC expert, I would like to ask @chiarazampolli or @JianLIUhep to approve this PR.
Note that it needs initial approval to steer the CI.

chiarazampolli (approved):

chiarazampolli:
This is not tested in the FST, but let's wait till tomorrow anyway.

PR #1410, opened 2024-01-17
Forbid AOD merging to fail

chiarazampolli:
@jgrosseo : I kept the same logic as before, but in principle, failing merging should make the process exit now.

### DIFF
--- a/DATA/production/configurations/asyncReco/async_pass.sh
+++ b/DATA/production/configurations/asyncReco/async_pass.sh
@@ -12,7 +12,7 @@
 
 # function to run AOD merging
 run_AOD_merging() {
-  o2-aod-merger --input list_$1.list --output $1/AO2D_merged.root > $1/merging.log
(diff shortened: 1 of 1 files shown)


jgrosseo:
Sounds good. And you keep the merging.log as well now, right?

chiarazampolli:
Hello @jgrosseo ,
Yes, but this is in the jdl, so I will put it in place once I have a tag with this merged, to not have too large jobs. 
Is the level 0 for verbosity ok? I did not check the executable...
Chiara

jgrosseo:
I checked and I think you can use level 1, it should not print too much.

chiarazampolli:
Hi @jgrosseo , 

Thanks for checking.
Done! Let me know if now I can merge.

Chiara

jgrosseo:
For me it looks good! Thanks!

chiarazampolli:
Great, thanks to you!

PR #1414, opened 2024-01-18
load pythia6 library

sawenzel:
Generator configs accessing TPythia6Decayer now need to load the pythia6 library explicitely, following https://github.com/AliceO2Group/AliceO2/pull/12514

Somehow the EGPythia6.so lib (where TPythia6Decayer is coming from) from ROOT is missing an automatic resolution link to these symbols.


sawenzel:
merging - currently broken - nightly tests will check it

PR #1416, opened 2024-01-22
[SimWF] Always set a process for Pythia8

benedikt-voelkel:
Always set a process for Pythia8.
if not set, the workflow creation goes through, pretending everything is fine.
However, if `-proc` was not passed, there was no Pythia8 configuration and the generation fails with 
```
[13:48:00][INFO] Initialising primary generator
 PYTHIA Error in ProcessLevel::init: no process switched on  
 PYTHIA Abort from Pythia::init: processLevel initialization failed  
[13:48:00][FATAL] Failed to init 'Pythia8': init returned with error
```


MC/config/common/pythia8/utils/mkpy8cfg.py
```diff
@@ -52,6 +66,9 @@
 
 args = parser.parse_args()
 
+# derive the process
+args.process = derive_process(args.process, args.idA, args.idB)
```
sawenzel, line 70: is there also the same for background processes?

benedikt-voelkel: This script is called for the signal as well for the background. So there is no difference and it would insert the process for signal as well as background. See for instance https://github.com/AliceO2Group/O2DPG/pull/1416/files#diff-d3d7f9e5d4c04202b033167540929ff40471636b8a1bb1ffdbecd1066c051eeaR414

sawenzel (approved):

benedikt-voelkel:
CI issue unrelated (comes up in AnalysisQC, @njacazio is going to fix it)
```
[ERROR] Exception caught: Couldn't get TTree "DF_1/O2pvmult" from "/sw/BUILD/85d704bd39bbd11095720208013cb6c51412b6f2/O2DPG-sim-tests/o2dpg-sim_tests/o2dpg_tests/workflows_bin/70_wf_test_pp.sh_dir/AO2D.root"
```

Merging since simulation runs just fine.

PR #1422, opened 2024-01-26
remove use-fit option in tof matcher (to be added in O2 default)

noferini:
This is to remove --use-fit in tof calibration from async, it would be added in O2 both for sync and async


chiarazampolli:
Ciao @noferini , 
Should I merge it? Are you sure that it is not something that at some point you will need for old data, for example? Do you want to configure it depending on the data taking year?
Cheers,
Chiara

noferini:
Hi @chiarazampolli,
for all data it is not needed since it affects only reconstruction with old tags and we don't foresee new reco with old tags to extract new TOF calibrations.
We can merge

chiarazampolli:
Ok!
Note that only DATA/production/configurations/asyncReco/setenv_extra.sh is used in production for any data, now, but it does not harm that you changed also the other scripts.
Chiara

PR #1423, opened 2024-01-26
Add FT0 QA to analysis QC

andreasmolander:


benedikt-voelkel:
As far as I can see. there seems to be a general issue with the TrackPropagation task in `O2Physics`.
Also the nightlies on alibi show that.
So it looks like the failed CI here might be unrelated. However, I would like to wait for a fix for the TrackPropagation issue and then rerun the CI here.

benedikt-voelkel:
Overall it looks fine!
Now, we have other problems with some analyses which are about to be solved.
However, I would propose to merge this PR; as far as I can see, everything looks good.
If there are no objections, I will do it shortly.

chiarazampolli:
Good for me too!

benedikt-voelkel (approved):

PR #1424, opened 2024-01-26
Include QC on the cluster information per track

lauraser:


chiarazampolli:
Merging, since we are preparing the tag.

PR #1425, opened 2024-01-26
TOF: add good match in anQC

njacazio:


njacazio:
Hi Chiara, this is due to an update in the o2, an error was introduced and that caused some issues at runtime. I think this might affect running ana QC with the most recent o2

benedikt-voelkel:
I asked a few days ago. The fox needs to be done in O2Physics. @mfasDa  can you fix it?

Concerning this PR and basically all others, it's unrelated.

chiarazampolli:
Ok, I would then merge, unless @benedikt-voelkel you prefer to wait.
Chiara
PS: @benedikt-voelkel , where is the problem tracked?

benedikt-voelkel:
Apologies, I didn't track the problem because for the experts it should have been super easy to solve.
But I will open a ticket now and link the people.

But the current conclusion is: It can be merged

benedikt-voelkel (approved):

PR #1432, opened 2024-01-31
Possibility to take external config for Pythia8

sawenzel:
In case of generator pythia8, we have so far always constructed a Pythia8 config file from the parameters given to o2dpg_sim_workflow.py

However, some expert users may want to use an external configuration for Pythia8.

This commit provides the possibility to do so via sensitivity to the `GeneratorPythia8.config` ConfigurableParam (so far ignored).

An example is:

```
${O2DPG_ROOT}/MC/bin/o2dpg_sim_workflow.py -eCM 14000 -col pp -gen pythia8 -proc cdiff -tf 2      \
                                           -ns 20 -e ${SIMENGINE}                                 \
                                           -j ${NWORKERS} -interactionRate 500000                 \
                                           -run 302000 -seed 624                                  \
                                           -confKey "GeneratorPythia8.config=/SOMEPATH/pythia8_powheg.cfg"
```

The new feature allows expert studies with specially setup Pythia8 configs. The development was motivated from https://its.cern.ch/jira/browse/O2-4549

However, note that options `-proc` `-eCM` etc. might have no effect or are ignored in such cases.

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -21,7 +21,7 @@ import sys
 import importlib.util
 import argparse
 from os import environ, mkdir, getcwd
-from os.path import join, dirname, isdir
+from os.path import join, dirname, isdir, isabs
 import random
 import json
 import itertools
@@ -505,6 +505,9 @@ TPC_SPACECHARGE_DOWNLOADER_TASK['cmd'] = '[ "${O2DPG_ENABLE_TPC_DISTORTIONS}" ] 
 workflow['stages'].append(TPC_SPACECHARGE_DOWNLOADER_TASK)
 
 
+# query initial configKey args for signal transport; mainly used to setup generators
+simInitialConfigKeys = create_geant_config(args, args.confKey)
+
 # loop over timeframes
 for tf in range(1, NTIMEFRAMES + 1):
    TFSEED = SIMSEED + tf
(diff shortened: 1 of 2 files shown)


MC/bin/o2dpg_sim_workflow.py
```diff
@@ -627,38 +630,46 @@ def getDPL_global_options(bigshm=False, ccdbbackend=True):
...
-                                --ptHatMin='+str(PTHATMIN)+'                             \
-                                --ptHatMax='+str(PTHATMAX)
-      if WEIGHTPOW   > 0:
-         SGN_CONFIG_task['cmd'] = SGN_CONFIG_task['cmd'] + ' --weightPow=' + str(WEIGHTPOW)
+      # see if config is given externally
+      externalPythia8Config = simInitialConfigKeys.get("GeneratorPythia8", {}).get("config", None)
+      if externalPythia8Config != None:
+         print ("Using external Pythia8 configuration " + str(externalPythia8Config))
+         # in this case, we copy the external config to the local dir (maybe not even necessary)
+         SGN_CONFIG_task['cmd'] = 'cp ' + externalPythia8Config + ' pythia8.cfg'
```
benedikt-voelkel, line 638: This will be executed inside the TF directory. So if one passes a relative path to the sim WF creation, the original file might not be found from inside the TF directory, right?

sawenzel: yes ... we should always use absolute/canonical paths for everything in O2DPG. Maybe need to add a special check function for this purpose.

benedikt-voelkel: one could do
```python
externalPythia8Config = abspath(externalPythia8Config)
if not exists(externalPythia8Config):
    # abort...
# now it can be used anywhere because we have an absolute path now
```

PR #1435, opened 2024-02-01
load libpythia6 for PWGEM GeneratorEMCocktailV2

DanielSamitz:


sawenzel:
@DanielSamitz , @benedikt-voelkel : There is a failing unit test for PWGEM. Could you take a look?

benedikt-voelkel:
I will test that locally now and report back

benedikt-voelkel:
Hello @DanielSamitz 
this generator does PbPb, right?
Because if so, I need to change something because otherwise it runs 100 PbPb events which is too much for a test like this.

DanielSamitz:
@benedikt-voelkel for this generator it does not really matter if pp or PbPb, with the current ini file it just produces 2100 mesons per event and decays them into electrons. But we run this generator always with the `--noGeant` option, that's why it is fast enough. I could reduce the number of mesons a bit, like a factor of 3 or so, by changing the ini file, but I don't know if this is enough to solve the problem.

benedikt-voelkel:
Seems that the test times out.
First round of digging (thanks @DanielSamitz ) has shown that the timeout of the CI/simulation is not related to this PR.
Hence merging.

At this point it seems that the macro stopped working correctly with https://github.com/alisw/alidist/pull/5278

Will open a JIRA ticket to follow up next week.

benedikt-voelkel (approved):

PR #1439, opened 2024-02-02
Bug fix for sampling

chiarazampolli:
@matthias-kleiner


miranov25:
trying to subscribe ...

miranov25:
Could you clarify which datasets were impacted by the bug? Was it confined solely to the triggered data productions?

Best regards,

Marian

chiarazampolli:
Ciao @miranov25 , 
No dataset since we have no production done with this change. If you see any problem with this change, let me know, otherwise I will merge it.
Chiara

chiarazampolli:
Hello Marian, 
The `ARGS_EXTRA_PROCESS_o2_tpc_time_series_workflow` is used to add extra options to workflow. What matters in this PR is the change in the env var used for the fraction: without this `ALIEN_JDL_SAMPLINGFACTORTIMESERIES` will be taken, which is a bug.

chiarazampolli:
Merged.

PR #1444, opened 2024-02-05
moving MC QC from test ccdb to qcdb

JianLIUhep:
Hi @chiarazampolli, @benedikt-voelkel, 

With this PR, we can upload MC QC objects to the newly created qcdb by @Barthelemy: https://its.cern.ch/jira/browse/QC-1094 
I tested it locally, and all looked good. If you agree, we can merge it and see if everything goes well with nightly tests. 

Thank you.

### DIFF
--- a/MC/bin/o2dpg_qc_finalization_workflow.py
+++ b/MC/bin/o2dpg_qc_finalization_workflow.py
@@ -33,7 +33,7 @@ def QC_finalize_name(name):
   return name + "_finalize"
 
 qcdir = "QC"
-def include_all_QC_finalization(ntimeframes, standalone, run, productionTag, conditionDB):
+def include_all_QC_finalization(ntimeframes, standalone, run, productionTag, conditionDB, qcdbHost):
 
   stages = []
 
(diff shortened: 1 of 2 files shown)


chiarazampolli (approved):

chiarazampolli:
Wonderful, thanks! 
I am not sure whether the fst runs also some MC tests, @benedikt-voelkel , maybe you know? If not, I would merge it.

benedikt-voelkel:
It does ruin MC tests.
Probably, EMCAL failed also here. I can check soon.
But if you can check adjust and it is again EMCAL, it can be merged.

chiarazampolli:
The error here is:
```
10-o2dpg_sim_workflow: Error! Interaction rate not >0 !!!
11:[FATAL]: O2DPG_TEST Workflow creation failed
```
and also:
```
1982:[INFO]    - First error: [02:34:10][FATAL] Unhandled o2::framework::runtime_error reached the top of main of o2-analysis-je-emcal-correction-task, device shutting down. Reason: Trying to dereference index with a wrong type in _as<>. Note that if you have several compatible index targets in your process() signature, the last one will be the one actually bound to the getter.
```

I think we should check the first, right?

chiarazampolli:
...but in both cases, it does not look like coming from this PR...

benedikt-voelkel:
Actually, you can ignore the first one. These are warnings, however, still mislabelled as errors.
What these tests are: Run all `run` shell scripts from PWGs. Note however, that in these cases the workflow would not be executed; it is just about creating it. But there, we don't always have everything under control, so we don't treat it as a failure.
We have one workflow that we maintain ourselves where the workflow is then also **executed**. And the is the one where EMCAL fails.

From that point of view, it is good to be merged.

I will make the CI better to read!

PR #1445, opened 2024-02-06
[SimWF] Change way to derive readout detectors

benedikt-voelkel:


benedikt-voelkel:
not needed anymore due to https://github.com/AliceO2Group/O2DPG/pull/1452

PR #1449, opened 2024-02-07
Adjust syst errors and ITS/TPC matching pars, use IDCCCDB as default scaling src

shahor02:
1) removed TPC cluster additive errors
2) reduces Z and tgL track syst errors
3) decreased matching chi2 cut and allowed max abs difference between TPC and ITS q/pt 
4) set the matching reference X = 60
5) default TPCSCALINGSOURCE changed from CTP to IDCCCDB

@miranov25 this sets the errors and matching chi2 cuts which should be reasonable with the tested corrections. If you have better ones, let me know.

### DIFF
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -293,8 +293,8 @@ elif [[ $ALIGNLEVEL == 1 ]]; then
   ERROB="100e-8"
   [[ -z $TPCITSTIMEERR ]] && TPCITSTIMEERR="0.2"
   [[ -z $ITS_CONFIG || "$ITS_CONFIG" != *"--tracking-mode"* ]] && export ITS_CONFIG+=" --tracking-mode async"
-  CUT_MATCH_CHI2=160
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Merging for the first tests.

miranov25:
Hello @shahor02 and @chiarazampolli 

Initially, I was under the impression that the error settings could be modified. However, upon closer inspection, I've realized that the values are hardcoded. My aim was to adjust the error settings based on the outcomes of our tests.

At present, I concur with the values you've shared. Nonetheless, I seek clarification on whether it will be feasible to adjust these values in the future, particularly in light of results from closure tests.

Given that the changes have already been merged, I presume this configuration is solely for testing purposes at this stage.
Could you please provide further clarification?

Marian

shahor02:
@miranov25 to change the errors a new stable-async tag will need to be generated with updated setev_extra in the O2DPG.

miranov25:
Thanks. Can we update O2DPG setev_extra   later?

shahor02:
@miranov25 : if you have no better values at hand then @chiarazampolli will likely start tomorrow a test pp run with current settings, then we can adjust/retag for the production.

chiarazampolli:
Hello, 
Just to confirm what @shahor02 said: for now we'll start with the values in this PR, if needed they can be changed. 
Chiara

PR #1452, opened 2024-02-09
Anchoring: Ability to get detector list from GRPECS

sawenzel:


benedikt-voelkel:
@TimoWilken maybe the CI failure is just intermittent, but maybe worth to check

benedikt-voelkel (approved):

PR #1453, opened 2024-02-10
Fix typo

benedikt-voelkel:


benedikt-voelkel:
@martenole @chiarazampolli there was a type on the TRD JSON file

martenole (approved):
Ah sorry about that. Thanks!

martenole:
I am wondering if we should add a basic check to the CI to make sure that all json files have the correct syntax to catch these kind of typos. Would it be feasible to run something like this for every PR @TimoWilken ?
```
for inpFile in `find . -not -path "./DATA/testing/private/*" -type f -name "*.json"`; do python -mjson.tool "$inpFile" > /dev/null; if [[ $? != 0 ]]; then echo ERROR found in $inpFile; fi; done
```

PR #1460, opened 2024-02-14
Debug mode for topology generation

martenole:
This would allow to run the `epn-topo-merger` with the option `--force-exact-node-numbers --nodes-mi50 1 --nmin-mi50 0 --nodes-mi100 1 --nmin-mi100 0` in case the user specifies `DEBUG_TOPOLOGY_GENERATION=1`. In that case also the temporary xml file would be kept for inspection.
Ping @davidrohr to see if its fine with you?

### DIFF
--- a/DATA/tools/epn/gen_topo.sh
+++ b/DATA/tools/epn/gen_topo.sh
@@ -18,7 +18,11 @@ if [[ -z "$EPN2EOS_METAFILES_DIR" ]] && [[ "0$WORKFLOWMODE" != "0print" ]]; then
   export EPN2EOS_METAFILES_DIR=/data/epn2eos_tool/epn2eos # Directory for epn2eos meta data files
 fi
 if [[ $USER == "epn" ]]; then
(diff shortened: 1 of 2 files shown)


martenole:
Hi @davidrohr could you check again if it makes sense for you like this? I have also added a separate directory for the topology cache on staging

davidrohr:
> Hi @davidrohr could you check again if it makes sense for you like this? I have also added a separate directory for the topology cache on staging

Only that there is this long-pending JIRA ticket https://its.cern.ch/jira/browse/EPN-436, to move the gen_topo temp folder of the scratch NFS, as requested by EPN. But up to them to do this, so long I am fine with your change. Afterwards one needs to think again.

davidrohr:
Also, in principle it is no problem if they use the same temp-folder, since I use a file-lock to serialize the access. It just means the topology cache uses the same folder, but since the cache is verified against all set env variables, this cannot be confused. Anyhow, a bit cleaner to separate it, so just go ahead.

martenole:
Thanks! Then pinging @lkrcal here as well for the temp folder on the infra nodes for the cache. And for now I merge this as is

PR #1462, opened 2024-02-14
Update anchorMC.sh - added proc arg

catalinristea:
Proc argument was missing.


benedikt-voelkel:
Just fyi, the default would be inel.

catalinristea:
Hi Benedikt,

Yes, I'll change it - it was requested by one PWG while ago to use cdiff - just adding very few diffractive events.
Thanks.

Yours, Catalin.

benedikt-voelkel:
@catalinristea 
Let's stick to `inel` as default. I am now working on making this configurable and will open PR with an hour.

benedikt-voelkel (approved):

PR #1463, opened 2024-02-15
[AnchorMC] Allow for any additional option for sim WF

benedikt-voelkel:
Use as ALIEN_JDL_ANCHOR_SIM_OPTIONS="-ini <ini_file> -confKey
"key=value;otherKey=otherValue" -trigger <trigger_file>
-productionTag myTag"

Build the final remainingargs by prepedning it.

The last argument wins, hence in this case -productionTag would not be
overwritten with this user choice

### DIFF
--- a/MC/bin/o2dpg_sim_workflow_anchored.py
+++ b/MC/bin/o2dpg_sim_workflow_anchored.py
@@ -409,7 +409,7 @@ def main():
            effTrigger = 28.0 # this is ZDC
          else:
            effTrigger = 0.759
-             
+
        # time needs to be converted to seconds ==> timestamp / 1000
        rate = retrieve_MinBias_CTPScaler_Rate(ctp_scalers, timestamp/1000., effTrigger, grplhcif.getBunchFilling().getNBunches(), ColSystem)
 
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
@chiarazampolli @catalinristea @noferini

benedikt-voelkel:
As an example, I tested with
```bash
ALIEN_JDL_SIM_OPTIONS="-gen external -proc cdiff -ini ${O2DPG_ROOT}/MC/config/PWGLF/ini/GeneratorLFStrangenessTriggered.ini"
```
and it runs through just fine.

benedikt-voelkel:
@chiarazampolli 
What would be your opinion on this? If you agree, I would merge it.
In addition, I will finally take care to add this and some other PRs to be ported back for 2022.

sawenzel:
Note that the name was changed to `ALIEN_JDL_ANCHOR_SIM_OPTIONS`. Better to adapt the comments in this PR.

benedikt-voelkel:
Done. When we merge, we can change the commit message accordingly

chiarazampolli:
Hello, 

I think we cannot have "_" in the JDL var, so `ALIEN_JDL_ANCHOR_SIM_OPTIONS` should be `ALIEN_JDL_ANCHORSIMOPTIONS`. @costing can you confirm? Some time ago it was like this.

Chiara

costing:
Ciao Chiara,

With the [current code](https://gitlab.cern.ch/jalien/jalien/-/blob/master/src/main/java/alien/site/JobWrapper.java#L1170) it is no longer a problem.

Tested with [this job](https://alimonitor.cern.ch/jobs/jdl.jsp?pid=3022041683) that produced [this output](https://alimonitor.cern.ch/users/download.jsp?view=true&path=/alice/cern.ch/user/g/grigoras/alien-job-3022041683/stdout). So underscores are no longer an issue.

Cheers,

.costin

chiarazampolli:
Good to know, thanks!

Chiara

PR #1464, opened 2024-02-15
[Anchor] Add test for anchored MC

benedikt-voelkel:
Runs a very simple anchored production, 2TFs, 2PbPb each

### DIFF
--- a/MC/run/ANCHOR/anchorMC.sh
+++ b/MC/run/ANCHOR/anchorMC.sh
@@ -241,6 +241,8 @@ if [[ -z "${DISABLE_QC}" && "${MCRC}" = "0" && "${remainingargs}" == *"--include
   # do QC tasks
   echo "Doing QC"
   ${O2DPG_ROOT}/MC/bin/o2_dpg_workflow_runner.py -f workflow.json --target-labels QC --cpu-limit ${ALIEN_JDL_CPULIMIT:-8} -k
+  # NOTE that with the -k|--keep-going option, the runner will try to keep on executing even if some tasks fail.
+  # That means, even if there is a failing QC task, the return code will be 0
   MCRC=$?
 fi
 
--- a/MC/run/ANCHOR/tests/test_anchor_2023_apass2_PbPb.sh
+++ b/MC/run/ANCHOR/tests/test_anchor_2023_apass2_PbPb.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+#
+# An example steering script for anchored MC simulations, PbPb
+#
+
+# example anchoring
+# taken from https://its.cern.ch/jira/browse/O2-4586
+export ALIEN_JDL_LPMANCHORPASSNAME=apass2
+export ALIEN_JDL_MCANCHOR=apass2
+export ALIEN_JDL_COLLISIONSYSTEM=Pb-Pb
+export ALIEN_JDL_CPULIMIT=8
+export ALIEN_JDL_LPMPASSNAME=apass2
+export ALIEN_JDL_LPMRUNNUMBER=544121
+export ALIEN_JDL_LPMPRODUCTIONTYPE=MC
+export ALIEN_JDL_LPMINTERACTIONTYPE=PbPb
+export ALIEN_JDL_LPMPRODUCTIONTAG=LHC24a1
+export ALIEN_JDL_LPMANCHORRUN=544121
+export ALIEN_JDL_LPMANCHORPRODUCTION=LHC23zzh
+export ALIEN_JDL_LPMANCHORYEAR=2023
+
+export NTIMEFRAMES=2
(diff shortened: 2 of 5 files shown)


benedikt-voelkel:
I would propose a CI for our anchored simulations.
In the current draft there is a PbPb, however, to be more conservative on the resource side, I think it should be a very small pp production instead (like 2 TFs, 10 events/TF) to take ~5min.
We can then also think about taking out the general WF tests in case an anchored test is run to save recourses.

It would be run if any of the following changes:
* `MC/run/ANCHOR/anchorMC.sh`
* `MC/bin|UTILS/parse-async-WorkflowConfig.py`
* `MC/bin/*`

What do you think @chiarazampolli @noferini ?

MC/run/ANCHOR/tests/test_anchor.sh
```diff
@@ -0,0 +1,39 @@
...
+export ALIEN_JDL_MCANCHOR=apass2
+export ALIEN_JDL_COLLISIONSYSTEM=Pb-Pb
+export ALIEN_JDL_CPULIMIT=8
+export ALIEN_JDL_LPMPASSNAME=apass2
+export ALIEN_JDL_LPMRUNNUMBER=544121
+export ALIEN_JDL_LPMPRODUCTIONTYPE=MC
+export ALIEN_JDL_LPMINTERACTIONTYPE=PbPb
+export ALIEN_JDL_LPMPRODUCTIONTAG=LHC24a1
+export ALIEN_JDL_LPMANCHORRUN=544121
+export ALIEN_JDL_LPMANCHORPRODUCTION=LHC22o
```
noferini, line 22: Period -> LHC23zzh

noferini:
Hi @benedikt-voelkel ,
I checked the anchorMC.sh script used and I saw
remainingargs="-gen pythia8 -proc inel -seed ${SEED} -ns ${NSIGEVENTS} --include-local-qc --pregenCollContext"


It seems to me that pythia8 is used (hardcoded).
Shall we parse

ALIEN_JDL_LPMINTERACTIONTYPE

to decide the generator (pp->pythia8 vs PbPb->pythia8hi)?

benedikt-voelkel:
Thanks for you feedback @noferini 

The latter decision is not needed. The PbPb or pp setting is derived automatically.
`pythia8` as well as the collision system will be forwarded from the anchored python script and that will then decide, which setting to apply.

For this PR, I will consider your pp setup and test it, thanks!

Following this PR, we can then consider: https://github.com/AliceO2Group/O2DPG/pull/1463

benedikt-voelkel:
Hej @TimoWilken 
Can you somehow see, if the test times out? Or anything else?
On my old desktop machine, it goes through in like 5 min.
Let me know if you need more information.

TimoWilken:
Hm, I've been getting a bunch of alerts that the O2DPG sim tests runs over its allocated 120GB of memory; specifically right now the command `o2-sim-primary-server-device-runner --control static --id primary-server --mq-config o2simtopology_4008.json -g external --noGeant -n 100 -j 4 --configFile /O2DPG/MC/config/PWGEM/ini/GeneratorEMCocktail.ini --configKeyValues GeneratorPythia8.includePartonEvent=true` is using 63% of a machine's 251G of memory (and slowly increasing).

It may be that the CI eventually gets killed by the OOM killer.

While it's currently testing this PR, I don't know if the large memory use is specifically due to this PR.

TimoWilken:
Adding to my previous comment, that process has also used 1h6m of CPU time already. Could there be something wrong there, or would it make sense to dial down the number of events it simulates?

benedikt-voelkel:
That is weird. 
```bash
o2-sim-primary-server-device-runner --control static --id primary-server --mq-config o2simtopology_4008.json -g external --noGeant -n 100 -j 4 --configFile /O2DPG/MC/config/PWGEM/ini/GeneratorEMCocktail.ini --configKeyValues GeneratorPythia8.includePartonEvent=true
```
should not come at all from this test.

Ok, so that thing has to be taken out: https://its.cern.ch/jira/browse/O2-4652

I will take care now

TimoWilken:
Thank you @benedikt-voelkel!

benedikt-voelkel:
@TimoWilken before I would merge it, I would like to ask if this time the CI was fine, cpu, mem and time-wise.

TimoWilken:
Thanks for checking @benedikt-voelkel, everything was fine this time!

noferini (approved):

noferini:
Fine with me.
@shahor02, @sawenzel , can you merge it?

chiarazampolli:
Label async-2023-pp-apass4 added to make the needed PR (https://github.com/AliceO2Group/O2DPG/pull/1686) apply

chiarazampolli:

PR #1465, opened 2024-02-15
Revert "Temporary workaround, to be reverted once O2 PR 12412 is in all async tags"

davidrohr:
This reverts commit fd619fe62c5c4f622492992a367da6945f1d5186.

@chiarazampolli : This is an old workaround I put in place for correct SHMSIZE and TIMEFRAME_RATE_LIMIT, since before O2 PR 12412 this was not used correctly.

I assume we are no longer running such old O2 tags, do we?
If not, I'd merge this to clean up the constants, since otherwise they are copy and pasted, and can get out of sync.
All is defined and used from workflow-multiplicities.sh now.


chiarazampolli:
This needs https://github.com/AliceO2Group/AliceO2/commit/f10cb506a9f3f14c0c0e0734ea9b9dcc1e296227, which was ported to the branch stable-async on 8.12, so it is fine to merge it, since we don't process any data with older tags. 

See:

https://github.com/AliceO2Group/AliceO2/commits/stable-async?after=194ecdddf57639c8544bc40b2c0f3edf8c6ec945+69
https://github.com/AliceO2Group/AliceO2/commits/async-20231208.1/

PR #1466, opened 2024-02-16
Revert "Update anchorMC.sh - added proc arg"

benedikt-voelkel:
Reverts AliceO2Group/O2DPG#1462

Needs to be reverted, doesn't work for PbPb. In that case, `heavy_ion` should be taken which would be the default if no `-proc` is set explicitly by the user.


benedikt-voelkel:
Just to let you know @catalinristea 
What I mean before by "`inel` being the default" is that it is set as the default in case it is a pp run and no `-proc` has been set by the user. In case of PbPb, that would be set to `heavy_ion`

A question: Did you do the initial PR #1462 because something was broken? Because in that case I understood that you would like to have something in the defaults in anchorMC.sh, for clarity.

chiarazampolli:
Hello @benedikt-voelkel , 
Yes, currently the tag of apass6 of 22o does not work if we don't pass anything.
Chiara

benedikt-voelkel:
Ciao @chiarazampolli ,
ok, but that is something for pp 22 tags, right?
Probably, we have to port 1 or 2 things from master to the 22 O2DPG tag. I will find the things to port first things in the morning and add them to the list.
Does that make sense?

PR #1467, opened 2024-02-16
TRD;) tracking task  in mc workflow

nasirmehdimalik:


MC/config/QC/json/trd-digits-task.json
```diff
@@ -43,7 +43,18 @@
...
+      },
+    "Tracklets": {
+        "active": "true",
+        "className": "o2::quality_control_modules::trd::TrackletsTask",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "dataSource": {
+          "type": "direct",
+           "query": "digits:TRD/DIGITS;tracklets:TRD/TRACKLETS;triggers:TRD/TRKTRGRD;noiseMap:TRD/NOISEMAP/0?lifetime=condition&ccdb-path=TRD/Calib/NoiseMapMCM;chamberStatus:TRD/CHSTATUS/0?lifetime=condition&ccdb-path=TRD/Calib/HalfChamberStatusQC;fedChamberStatus:TRD/FCHSTATUS/0?lifetime=condition&ccdb-path=TRD/Calib/DCSDPsFedChamberStatus"
```
martenole, line 55: I don't think you need to query the digits for the tracklets task

MC/config/QC/json/trd-tracking-task.json
```diff
@@ -0,0 +1,48 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "ccdb-test.cern.ch:8080",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable"
```
martenole, line 9: all these `not_applicable` entries can be removed

MC/config/QC/json/trd-tracking-task.json
```diff
@@ -0,0 +1,48 @@
...
+      }
+    },
+    "tasks": {
+      "Tracking": {
+        "active": "true",
+        "className": "o2::quality_control_modules::trd::TrackingTask",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "maxNumberCycles": "-1",
```
martenole, line 35: this is the default in any case, so not needed

martenole (commented):
Thanks @nasirmehdimalik . I have only some minor comments

MC/config/QC/json/trd-tracking-task.json
nasirmehdimalik: without this lines  gives error
`
std::exception::what: No such node (moduleName)

terminate called after throwing an instance of 'boost::wrapexcept<boost::property_tree::ptree_bad_path>'
  what():  No such node (moduleName)`
`

MC/config/QC/json/trd-tracking-task.json
nasirmehdimalik: I'm keeping them because all other JSON files have the same structure, and I'm unsure if they might be needed later

MC/config/QC/json/trd-digits-task.json
nasirmehdimalik: done

martenole (approved):
Approving only to start the CI

MC/config/QC/json/trd-tracking-task.json
knopers8: I'm quite sure Ole meant `maxNumberCycles`, not `moduleName`

martenole: Yes indeed. @nasirmehdimalik can you please remove `maxNumberCycles`?

MC/config/QC/json/trd-tracking-task.json
martenole: But we should try to keep the json files as compact as possible. It makes maintenance easier. So please remove them. If there are other TRD QC json files where you find this you could even clean them up there as well :)

nasirmehdimalik: done

martenole (commented):
Thanks @nasirmehdimalik ! Now that I look at it again, could you please also change the name `trd-digits-task.json` to something like `trd-standalone-tasks.json`? This would reflect that also the tracklets task is there and in the future all QC tasks not requiring any other detector could go there.

MC/config/QC/json/trd-tracking-task.json
```diff
@@ -0,0 +1,44 @@
...
+    "tasks": {
+      "Tracking": {
+        "active": "true",
+        "className": "o2::quality_control_modules::trd::TrackingTask",
+        "moduleName": "QcTRD",
+        "detectorName": "TRD",
+        "cycleDurationSeconds": "60",
+        "dataSource": {
+          "type": "direct",
+          "query": "trackITSTPCTRD:TRD/MATCH_ITSTPC;trigITSTPCTRD:TRD/TRGREC_ITSTPC"
```
martenole, line 34: Can you please add also TPC-TRD matches to the query here?

MC/config/QC/json/trd-tracking-task.json
```diff
@@ -0,0 +1,44 @@
...
+        },
+        "taskParameters": {
+          "detailedQC": "false",
+          "trackSources": "ITS-TPC-TRD"
```
martenole, line 38:
```suggestion
          "trackSources": "ITS-TPC-TRD,TPC-TRD"
```
and here as well

martenole (commented):
Looks good to me now. Could you just also add the TPC-TRD matches please? Online we disable the TPC-TRD matching, but in the asynch. processing it is there, so we should also check the QC for it. Sorry didn't notice this earlier in you PR. But this is a tiny change and then we should be ready for merging it

MC/config/QC/json/trd-tracking-task.json
martenole, line 34:
```suggestion
          "query": "trackITSTPCTRD:TRD/MATCH_ITSTPC;trigITSTPCTRD:TRD/TRGREC_ITSTPC;trackTPCTRD:TRD/MATCH_TPC;trigTPCTRD:TRD/TRGREC_TPC"
```

martenole (approved):
Should be fine now, let's wait for the CI to finish

nasirmehdimalik:
@martenole since the checks fail  and I pasted here the error "[ERROR] invalid workflow in o2-qc: No matching output found for <matcher query: (and origin:TRD (and description:MATCH_TPC (just startTime:$0 )))> as requested by data processor "qc-task-TRD-Tracking"". I think it is due to last commit.

martenole:
@sawenzel it seems in the simulation check the matching of TRD to TPC tracks is disabled. Can you explain to me how the `trd_track_sources` are obtained in https://github.com/AliceO2Group/O2DPG/blob/826526fdecf78c66359eba4f3fa3ef4c1cdc6973/MC/bin/o2dpg_sim_workflow.py#L1013 ?
Which configuration is used for the CI test here?

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -1208,7 +1208,12 @@ def addQCPerTF(taskName, needs, readerCommand, configFilePath, objectsFile=''):
      addQCPerTF(taskName='trdDigitsQC',
                 needs=[TRDDigitask['name']],
                 readerCommand='o2-trd-trap-sim',
-                configFilePath='json://${O2DPG_ROOT}/MC/config/QC/json/trd-digits-task.json')
+                configFilePath='json://${O2DPG_ROOT}/MC/config/QC/json/trd-standalone-task.json')
+
+     addQCPerTF(taskName='trdTrackingQC',
+                needs=[TRDTRACKINGtask2['name']],
+                readerCommand='o2-global-track-cluster-reader --track-types "ITS-TPC-TRD" --cluster-types none',
```
benedikt-voelkel, line 1215: I would assume that the reader needs to be configured correctly to take into account the TPC-TRD matches.

benedikt-voelkel:
From my side we can go ahead with this, @martenole @nasirmehdimalik

PR #1483, opened 2024-02-20
Cleanup of predefined ECS configurations

martenole:
Hi @shahor02 I have set `TPC_CORR_SCALING` as discussed in https://its.cern.ch/jira/browse/O2-4666. Shall we for synthetic runs disable the correction as we do for cosmics? The current central synthetic configurations don't have `TPC_CORR_SCALING` set, so they are using the default settings at the moment. If the behaviour should remain the same I will update this PR.


shahor02:
@martenole sorry for late response. Disabling corrections would lead to very bad reco quality for high rate runs. I would use the same settings as for real run. If at home stage we get conflict with calibration object loaded for async and requiring different workflow  options, we will think how to circumvent it. E.g. we can add a ccdb option `--not-later-than` pointing to a timestamp of relevant SOR + a few days offset.

martenole:
Thanks @shahor02 I will try to summarize if I got it correctly:
- **PHYSICS** runs shall use `TPC_CORR_SCALING="--lumi-type 1 TPCCorrMap.lumiInstFactor=2.414"`
- **SYNTHETIC** runs shall use the same option as **PHYSICS**, since otherwise `REPLAY` runs won't be reconstructed correctly. But for simulated raw data we don't yet have distortions implemented, or? So should in this case the corrections rather be disabled?
- **COSMICS** we can safely disable the corrections
- **ASYNC** I expect the correct settings will be set manually by the async scripts overwriting `TPC_CORR_SCALING` as it is done today

shahor02:
Correct. For the SYNTH simulated runs we can use anything since the correction maps for "non-anchored" period are empty.

martenole:
Thanks, then I think this PR is already correct. For synthetic runs the beamtype is set to either pp or PbPb, so we will use the same settings as for physics there.

PR #1487, opened 2024-02-21
Change eta cut for TPC tracks in GLO QC.

mfaggin:


mfaggin:
Hi @noferini @chiarazampolli @benedikt-voelkel this is the modification discussed this morning at the DPG corodination meeting.

However, we cannot put any lable on it because we do not have the permissions (the `Label` field in the right sidebar is not clickable for us)

noferini:
Done
For the next time you can follow @benedikt-voelkel suggestion in slide 4
https://indico.cern.ch/event/1385445/contributions/5828895/attachments/2805533/4895407/20240221_slides.pdf

noferini (approved):
trivial PR

mfaggin:
> Hi @mfaggin , I removed the label for pp 2022 apass6. I would leave only the one for Pb-Pb since pp apass6 date reco was already done and I would keep the same config for DATA and MC. Do you agree?

ciao Francesco, since the apass6 is already done then yes, nothing against it

chiarazampolli (approved):

PR #1489, opened 2024-02-23
Add asyn-label workflows for PRs

benedikt-voelkel:
Following what was introduced in https://github.com/alisw/ali-bot/pull/1293

### DIFF
--- a/.github/workflows/async-auto-label.yml
+++ b/.github/workflows/async-auto-label.yml
@@ -11,7 +11,7 @@ permissions:
   pull-requests: write   # to update labels
 
 jobs:
-  clean:
+  apply_async_labels:
     name: Apply requested async label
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
@TimoWilken what do you think?
I hope I did everything correctly

TimoWilken:
I think you may need to add the `pull-requests: write` permission to the top-level permissions block as well, not just the `jobs` block!

TimoWilken:
Hm, I'm not sure, maybe you need `issues: write` as well? But then I don't know why it worked before...

benedikt-voelkel:
Is it because it is run from my PR and not from upstream?

benedikt-voelkel:
Merge to see if things work once they are in the upstream repo

PR #1493, opened 2024-02-23
Fix converter logic for o2-analysis-v0converter in test workflow

jezwilkinson:
Since o2-analysis-v0converter produces the v0_002 table, it should be checking for that and not O2v0_001 to decide whether to run the converter

### DIFF
--- a/MC/analysis_testing/o2dpg_analysis_test_workflow.py
+++ b/MC/analysis_testing/o2dpg_analysis_test_workflow.py
@@ -190,7 +190,7 @@ def get_additional_workflows(input_aod):
         o2_analysis_converters = {"O2collision_001": "o2-analysis-collision-converter --doNotSwap",
                                   "O2zdc_001": "o2-analysis-zdc-converter",
                                   "O2bc_001": "o2-analysis-bc-converter",
-                                  "O2v0_001": "o2-analysis-v0converter",
+                                  "O2v0_002": "o2-analysis-v0converter",
                                   "O2trackextra_001": "o2-analysis-tracks-extra-converter"}
         for i in froot.GetListOfKeys():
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
fyi: Don't worry about the failed "Collect and print async labels" CI for now.

benedikt-voelkel (approved):

benedikt-voelkel:
@jezwilkinson should this be contained in any of the production tags for async reconstruction/anchored MC? It would depend on whether the corresponding O2 developments for the `v0_002` table are contained in those tags. Do you know?

Otherwise, would the worst-case scenario be that the `v0_002` table would be simply overwritten the converter?

jezwilkinson:
Hi @benedikt-voelkel ,
I am not sure if the script is actively used with a configuration that would be affected in the production pipeline so can't really speak to that. But the worst-case is a crash of the analysis testing workflow (reason for spotting this issue was some tests ongoing in the TPC reconstruction group that were failing - if the v0 converter tries to run with a missing v0s_001, as would be triggered before the change, then it will crash with a "missing tree" error)

benedikt-voelkel:
To get back to this @jezwilkinson 
This is not yet included in the corresponding `async-2023-pbpb-apass3` software. However, I am not aware of any issues due to the fact that it is not included. Could you comment?
If for any reason strictly necessary, let's include it. Otherwise, could we also think about removing the label and not include it?
Thanks a lot!

jezwilkinson:
as this was to fix a bug that @wiechula and @aschmah were encountering I leave it for them to comment whether it is still required

wiechula:
It might be we only use it for the local testing. So I cannot really comment. If there we no issues reported up to now, I would guess not.

PR #1499, opened 2024-02-24
Enable strangeness tracking in strange particle simulations

fmazzasc:


fmazzasc:
Hi @sawenzel , @benedikt-voelkel , could we merge this?

benedikt-voelkel:
Sure @fmazzasc 
Quick question before I merge though: These are only steering shell scripts so I would assume they are not important for any official physics pass simulation software, right?

fmazzasc:
Hi @benedikt-voelkel , yes exactly!

benedikt-voelkel (approved):

PR #1500, opened 2024-02-26
Do not ask for geometry in GLO ITS-TPC and TOF matching QC

chiarazampolli:
@shahor02 , @noferini , @ercolessi , @njacazio 
@ercolessi , @njacazio , @noferini : for TOF, the same should be done online

### DIFF
--- a/DATA/production/qc-async/itstpc.json
+++ b/DATA/production/qc-async/itstpc.json
@@ -58,7 +58,7 @@
           "etaCut": "1e10f"
         },
 	"grpGeomRequest" : {
-          "geomRequest": "Aligned",
+          "geomRequest": "None",
           "askGRPECS": "false",
           "askGRPLHCIF": "false",
           "askGRPMagField": "true",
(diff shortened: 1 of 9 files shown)


noferini:
For my understanding, why is geometry no longer required?

shahor02 (approved):
Thanks! Labels :) ?

chiarazampolli:
Yes, I am still not used to them :-) 

Chiara

chiarazampolli:
@benedikt-voelkel , what is "async-data"? what is "async-mc"? 
I also expect no MC for apass1 of pp 2023 anymore: do we want to patch it too? @noferini

chiarazampolli:
> For my understanding, why is geometry no longer required?

To my knowledge, the full geometry is needed only when we use material corrections, but we don't. We use propagateParamToDCA where Bz is used.

@shahor02  can comment more. 
See also O2-4687.

noferini:
Since I think that this is critical only for data, I would mark it only for async-2023-pbpb-apass,
apass4/apass6 2022 data reco is over and we will not reconstruct anything with those tags but MC (-> no need to cheery-pick this change there).
Do you agree?
In general I would not touch QC in a tag when data reco is over (better to have the same setting in data and MC)

noferini:
Hi @shahor02 , 
also @ercolessi ran a local test with and without geo enabled in the json and she got the same result (and no crash).
Then , I confirm we don't need geo for TOF there

PR #1501, opened 2024-02-26
Create ITS/MFT light geom snapshots together with geometry-aligned

shahor02:
Goes together with https://github.com/AliceO2Group/AliceO2/pull/12761 : since MC creates its own snapshot of GeometryAligned, it should also create the lightweight ITS and MFT geometries. This is because now many workflows don't use GeometryAligned directly but fetch ITS(MFT)/Config/Geometry objects created from GeometryAligned.


benedikt-voelkel:
Since there is nothing explicitly in the logs and I see
```
##########################
# ERROR for anchored MCs #
##########################
```
it points to this issue: https://its.cern.ch/jira/browse/O2-4706

shahor02:
Thanks @benedikt-voelkel !

benedikt-voelkel:
https://github.com/AliceO2Group/O2DPG/pull/1506 looks good @shahor02 
I guess once you merge it, you could maybe steer the CI here again. Then we would have our first test of the pp anchored already here in the CI

shahor02:
Merging after rebasing, CIs were green.

PR #1505, opened 2024-02-28
Default for RECO_NUM_NODES_WORKFLOW_CMP

chiarazampolli:
@benedikt-voelkel , for https://its.cern.ch/jira/browse/O2-4706
@davidrohr , @martenole : 0 should be invalid as default, but I prefer that it crashes, than to put a meaningless value. Let me know.

PR #1506, opened 2024-02-28
start with numa scaler 1, will be set to real value on EPN

shahor02:


davidrohr (changes requested):
See my comment in #1505 1505

shahor02:
@davidrohr I don't see where the `RECO_NUM_NODES_WORKFLOW` and `NUMAGPUIDS` are defined, isn't parser running for epns only?

davidrohr:
`RECO_NUM_NODES_WORKFLOW` is defined in dpl-workflow.sh, and `NUMAGPUIDS` is defined in setenv.sh, so both should be available.
And the default should not be 1, it should be 230. If you set it to one, it will assume it has to do the load of all EPN nodes, so will multiply multiplicities by 230 with some cuts.

Anyhow, I would just move up the `RECO_NUM_NODES_WORKFLOW_CMP` definition, as I remarked in Chiara's PR.

shahor02:
ok, thanks

davidrohr (approved):

chiarazampolli:
Hello @shahor02 , 

Why did you add the labels? The problem was introduced by https://github.com/AliceO2Group/O2DPG/pull/1483, which is not in stable-async. 

Chiara

shahor02:
I thought it should be ported. Removed the labels.

PR #1507, opened 2024-02-28
TPC: Add possibility to simulate distortions

wiechula:
Requires adding `--tpc-distortion-type 2` in the anchorMC script to `remainingargs` not done in this PR.


wiechula:
@chiarazampolli , @sawenzel , shall we go ahead with this?

chiarazampolli:
Hello @wiechula , 
I would let @sawenzel and @benedikt-voelkel review in detail. For now, I have a question: you checked thoroughly that the use of `rate.second` is what you need, then scaled by 2.414, right?

wiechula:
@chiarazampolli , I didn't check the numbers in detail. But `rate.second` are the scalers (ZDC for PbPb) if I understand correctly. And to scale to `pp` in `PbPb` we should multiply by `2.414`. Unless I overlook something.

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -988,13 +1010,36 @@ def getDigiTaskName(det):
...
+   tpcLocalCFreco=dict()
+
+   # handle distortion corrections and scaling using MC maps
+   # this assumes the lumi inside the maps is stored in FT0 (pp) scalers
+   # in case of PbPb the conversion factor ZDC ->FT0 (pp) must be set
+   tpc_corr_options_mc=''
+
+   if tpcDistortionType == 0: # disable distortion corrections
+      tpc_corr_options_mc=' --corrmap-lumi-mode 0 '
+      tpcLocalCFreco['TPCCorrMap.lumiInst'] = '-1';
```
shahor02, line 1023: Disabling corrections now goes via `TPCCorrMap.lumiMean=-1`

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -988,13 +1010,36 @@ def getDigiTaskName(det):
...
+   elif tpcDistortionType == 1: # disable scaling
+      tpc_corr_options_mc=' --corrmap-lumi-mode 2 '
+      tpcLocalCFreco['TPCCorrMap.lumiInst'] = str(CTPSCALER)
+      tpcLocalCFreco['TPCCorrMap.lumiMean'] = str(CTPSCALER)
```
shahor02, line 1027: Is the comment `disable scaling` typo here? Note that to disable scaling one should pass `--lumi-type 0`.

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -988,13 +1010,36 @@ def getDigiTaskName(det):
...
 
+   # ===| TPC reco |===
+   tpcLocalCFreco=dict()
+
+   # handle distortion corrections and scaling using MC maps
+   # this assumes the lumi inside the maps is stored in FT0 (pp) scalers
+   # in case of PbPb the conversion factor ZDC ->FT0 (pp) must be set
+   tpc_corr_options_mc=''
+
+   if tpcDistortionType == 0: # disable distortion corrections
```
benedikt-voelkel, line 1021: In anchored simulations I think these configurations might clash with some we would derive from the anchoring and pass from the outside. I think to be discussed with @sawenzel as well.

MC/bin/o2dpg_sim_workflow.py
wiechula: Thanks for spotting!

MC/bin/o2dpg_sim_workflow.py
wiechula: Here, the scaling is simply disabled by setting `lumiInst` and `lumiMean` to the same value.

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -232,19 +232,22 @@ def retrieve_MinBias_CTPScaler_Rate(ctpscaler, finaltime, trig_eff, NBunches, Co
...
     if rate.first >= 0:
       # calculate true rate (input from Chiara Zampolli) using number of bunches
       coll_bunches = NBunches
       mu = - math.log(1. - rate.second / 11245 / coll_bunches) / trig_eff
       finalRate = coll_bunches * mu * 11245
-      return finalRate
+      return finalRate, scaler
 
     print (f"[ERROR]: Could not determine interaction rate; Some (external) default used")
     return None
```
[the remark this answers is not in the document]
wiechula: Indeed, thanks!

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -232,19 +232,22 @@ def retrieve_MinBias_CTPScaler_Rate(ctpscaler, finaltime, trig_eff, NBunches, Co
...
+    print("Fetching rate with time " + str(finaltime) + " class " + str(ctpclass) + " type " + str(ctptype))
     rate = ctpscaler.getRateGivenT(finaltime, ctpclass, ctptype)
     #if ColSystem == "PbPb":
     #  rate.first = rate.first / 28.
     #  rate.second = rate.second / 28.
 
     print("Global rate " + str(rate.first) + " local rate " + str(rate.second))
+    scaler = None
+    if rate.second >= 0:
+      scaler = rate.second
```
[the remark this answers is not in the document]
wiechula: I guess this is a matter of taste. But what `ctpscaler` returns are the scaler values or some kind of counts from its trigger inputs. So how about `ctp_rate`?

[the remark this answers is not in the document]
wiechula: But `finalRate` also uses `rate.second` which you refer to as `local_rate`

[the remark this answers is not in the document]
benedikt-voelkel: Yes, you are right, `finalRate` is *computed* from `rate.second`.
When I read this (https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/CTP/include/DataFormatsCTP/Scalers.h#L126-L127), what you return is the bare/raw local rate in some sense. I think, it could be called something like `local_rate_raw` or `ctp_local_rate_raw` or something.

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -232,19 +232,22 @@ def retrieve_MinBias_CTPScaler_Rate(ctpscaler, finaltime, trig_eff, NBunches, Co
...
     if rate.first >= 0:
```
benedikt-voelkel, line 245: @chiarazampolli do you know why we would only return the final/corrected local rate if the global rate is `rate.first >= 0`? I don't see it being used inside this branch.

chiarazampolli: Was it maybe added here: https://github.com/AliceO2Group/O2DPG/pull/1333?

sawenzel:
Sorry for the delay. I'll take a look asap.

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -976,13 +998,36 @@ def getDigiTaskName(det):
...
+      tpcLocalCFreco['TPCCorrMap.lumiMean'] = '-1';
+   elif tpcDistortionType == 1: # disable scaling
+      tpc_corr_options_mc=' --corrmap-lumi-mode 2 '
+      tpcLocalCFreco['TPCCorrMap.lumiInst'] = str(CTPSCALER)
+      tpcLocalCFreco['TPCCorrMap.lumiMean'] = str(CTPSCALER)
+   elif tpcDistortionType == 2: # full scaling with CTP values
+      if COLTYPE == 'PbPb':
+         tpcLocalCFreco['TPCCorrMap.lumiInstFactor'] = str(lumiInstFactor)
+      tpc_corr_options_mc=' --corrmap-lumi-mode 2 '
+      tpcLocalCFreco['TPCCorrMap.lumiInst'] = str(CTPSCALER)
```
benedikt-voelkel, line 1020: This does not correspond to what was done for digitization.
Here, `TPCCorrMap.lumiInst` and `TPCCorrMap.lumiInstFactor` are set separately whereas for digitization `TPCCorrMap.lumiInst` is set directly to the factor.

Is it because the configKey values are digested differently in reconstruction and digitization?

wiechula: Indeed, the digizer does not use the lumiInstFactor https://github.com/AliceO2Group/AliceO2/blob/0129938292bf379849ec58652be5c1cfa88e60f1/Detectors/TPC/simulation/src/Digitizer.cxx#L227
While to correction map in reco does: https://github.com/AliceO2Group/AliceO2/blob/0129938292bf379849ec58652be5c1cfa88e60f1/Detectors/TPC/calibration/src/CorrectionMapsLoader.cxx#L201
For reconstruction I wanted to mirror what is done in data. But I assume one could also directly multiply here, right @shahor02 ?

sawenzel:
I have tested this PR and it works for run 544013 (in the range 544013 - 544126).

There was slight glitch with `defaults-o2-epn` (segfault in ROOT access to CCDB objects) on the EPNs but with `default-o2` it is ok. The PR also doesn't modify default behaviour since distortions in digitization need to be enabled with `--tpc-distortion-type` in the arguments to the anchoring script.

From my point of view this can be merged.

benedikt-voelkel (approved):

chiarazampolli:
Hello @wiechula , 

Following what we said just now at the WP12+13 meeting, just to be sure: @sawenzel  did the test with run 544013, did this run have the CCDB for the test? Or do we have it only for runs in:

544116
544121
544122
544123
544124
?
Chiara

wiechula:
@chiarazampolli , the runs covered are 544013 - 544126, see O2-4531.

chiarazampolli:
Good, thanks!

PR #1509, opened 2024-02-28
To allow to set the trackQC sampling from the JDL, or have a subsampl…

chiarazampolli:
…e with full sampling

@miranov25 @noferini @catalinristea 

I used cksum to hash the first CTF file, let me know if you agree.


miranov25:
Hello Chiara,

Thank you for the implementation. I am not entirely certain about the fraction of data that will be selected at a 100% fraction, whether this can be adjusted, or if the selection process is deterministic, allowing for straightforward identification  for the future usage

Best regards,
Marian

chiarazampolli:
Hello @miranov25 , 
The fraction will be 10% (100 permil), unless specified differently in the JDL (via ALIEN_JDL_PERMILFULLTRACKQC). This is what you asked. 
The use of cksum converts the name of the first CTF that will be processed in an integer, and this will be used to decide. In principle, it is deterministic - I tried on linux and mac and got the same number - but I am not entirely sure. 
Feel free to change, for now we will start with this.
Chiara

PR #1511, opened 2024-02-29
[Anchor] Add some more help messages

benedikt-voelkel:
* help messages
* set ALIEN_JDL_ADDTIMESERIESINMC to a value at the beginning

### DIFF
--- a/MC/run/ANCHOR/anchorMC.sh
+++ b/MC/run/ANCHOR/anchorMC.sh
@@ -41,10 +41,13 @@ print_help()
   echo "PRODSPLIT."
   echo
   echo "Optional are:"
-  echo "NWORKERS,"
-  echo "ALIEN_JDL_CPULIMIT or CPULIMIT,"
-  echo "ALIEN_JDL_SIMENGINE or SIMENGINE."
-  echo "DISABLE_QC (set this to disable QC, e.g. DISABLE_QC=1)"
+  echo "ALIEN_JDL_CPULIMIT or CPULIMIT, set the CPU limit of the workflow runner, default: 8,"
+  echo "NWORKERS, set the number of workers during detector transport, default: 8,"
+  echo "ALIEN_JDL_SIMENGINE or SIMENGINE, choose the transport engine, default: TGeant4,"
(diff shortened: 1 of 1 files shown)


catalinristea:
Ciao @benedikt-voelkel, @chiarazampolli 

Just a short comment, regarding the use of ALIEN_JDL_ADDTIMESERIESINMC: if we do this in anchorMC.sh, then it will concern only the anchored MC, while we want it for any MC (common, PWG?) right?
Isn't it:
- Anchored MC: anchorMC --> o2dpg_sim_workflow_anchored --> o2dpg_sim_workflow.py
- Unanchored: o2dpg_sim_workflow.py
Couldn't this be done in the base script o2dpg_sim_workflow.py?

Thanks, Catalin.

chiarazampolli:
Hello @benedikt-voelkel , 
Why do you need a default? I mean, even without it, it will be ON if you don't specify anything different. 
Is this just more elegant? Or is there something I overlook?

Chiara

benedikt-voelkel:
It's just for consistency. Of course it would not create problems but this relates somewhat to the "unbound issue" we had although - again - in this case, it would be fine.
But most importantly, I wanted to align things a little bit more and have everything that is somehow used at one place at the top of the file if that makes sense.

chiarazampolli:
Sure, fine with me, I wanted to make sure that I did not miss anything, since I tagged already without this change.

miranov25:
> Hello,
> 
> I don't think that this is needed in unanchored MCs. I would even think that unanchored is not needed at all anymore...
> 
> @miranov25, can you comment?
> 
> Chiara

I am not sure I understand the comment. For unanchored MC, the time series are indeed useless, but representatively sampled data are useful (I was checking it)
If the unanchored production is still used in the physical analysis for corrections, I would like to keep the representative sampling. I will check what happened in the jet production which I was running at GSI.

benedikt-voelkel:
In any case, that is a different topic: For unanchored we would need to change different things.
Merging.

chiarazampolli:
Unanchored should *not* be used for corrections, which is why i only applied the change to the anchoring script.

if needed, let’s take it in another pr.


Chiara

chiarazampolli:
Label async-2023-pp-apass4 added to make the needed PR (https://github.com/AliceO2Group/O2DPG/pull/1686) apply

chiarazampolli:

PR #1513, opened 2024-03-01
[Anchor] Remove -k from QC run, return error code

benedikt-voelkel:
Additionally, some minor clarifications and transparent changes.

### DIFF
--- a/MC/run/ANCHOR/anchorMC.sh
+++ b/MC/run/ANCHOR/anchorMC.sh
@@ -46,8 +46,8 @@ print_help()
   echo "ALIEN_JDL_SIMENGINE or SIMENGINE, choose the transport engine, default: TGeant4,"
   echo "ALIEN_JDL_WORKFLOWDETECTORS, set detectors to be taken into account, default: ITS,TPC,TOF,FV0,FT0,FDD,MID,MFT,MCH,TRD,EMC,PHS,CPV,HMP,CTP,"
   echo "ALIEN_JDL_ANCHOR_SIM_OPTIONS, additional options that are passed to the workflow creation, default: -gen pythia8,"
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
@chiarazampolli @catalinristea 
This allows to let `anchorMC.sh` return with the QC error code in case QC failed.
By default, the runner has the `-k` option which makes it keep going as far as possible. But that also means that the return code will be 0 by default.

See changed tests in this PR for examples.

If you agree, I would merge it.

chiarazampolli:
Ciao @benedikt-voelkel , 

I think we should stop if QC fails, yes. Is this the case in production? Because I would have it there too.

Chiara

chiarazampolli:
I personally think that failing QC should make the process fail. QC should just work. @JianLIUhep , Elena (I don't find her github name), @noferini , what do you think?

If 3 is possible, then it is good, so that we find all that would fail :-)

benedikt-voelkel:
If we want to go for something like option 3, I would like @sawenzel to comment because I don't want to simply change the return code logic of the workflow runner itself.
But that decision can be made independent of this PR.

Until then, I would remove the `-k` option as default.
That would mean that for now, QC would abort on the first error and the achnorMC. return code would be different from 0.

For that I would need to make slight adjustments to this PR

JianLIUhep:
Hi all, I am also in favor of having the option 3, and agree with above the comment by Benedikt.  

I cannot find Elena, either. Shared this PR with her.

benedikt-voelkel:
Done: `-k` removed, QC run will abort as soon as there is an issue.
Discussion on option 3 to be followed up.

benedikt-voelkel:
In case of no objections, I would merge this @chiarazampolli

PR #1514, opened 2024-03-01
[AnalysisQC] Make common args better adjustable

benedikt-voelkel:


benedikt-voelkel:
Every now and then I get 
```
shmem: could not create a message of size <some-size>, alignment: 64, free memory: <some-available-size>
```
for EMCAL.
Increasing the SHM by 25% gets rid of that consistently.

PR #1526, opened 2024-03-06
Fixing cases in which the number starts with 0 and is considered base 8

chiarazampolli:
@shahor02 
I am still checking why I had `wc -l`...

### DIFF
--- a/DATA/production/common/setVarsFromALIEN_PROC_ID.sh
+++ b/DATA/production/common/setVarsFromALIEN_PROC_ID.sh
@@ -3,18 +3,25 @@
 # Script to set a couple of variables that depend on ALIEN_PROC_ID.
 # Since several scripts might need this, it is in a separate script.
 
-# let's get the last 10 (for INT) and 5 (for int16) digits of ALIEN_PROC_ID, to be passed to define NUMAID and shm-segment-id via O2JOBID, which are int and int16 respectively. Then we make them an int or int16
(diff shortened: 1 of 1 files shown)


chiarazampolli:
In this case, `-l` gives a negative value --> so I take the substring from the end:
e.g.:
3027145464:-9 --> 027145464
`-c` gives a positive value, and it is used as offset for the substring:
e.g.:
3027145464:1 --> 027145464
So the net result is the same. 
I think that `-c` is better, because `-l` will always be 1, no matter the length of the proc_id.

chiarazampolli:
Note that with this script, we don't get the last 10 or 9 digits, as in the comment, but the last 9 (base 32) or 8 (base 16). 
To get the last 10 (or 9), we could just use:
```
ALIEN_PROC_ID_OFFSET_INT32=$((10#${ALIEN_PROC_ID: -10}))
ALIEN_PROC_ID_OFFSET_INT16=$((10#${ALIEN_PROC_ID: -9}))
```
I will try with more time tomorrow.

chiarazampolli:
Hello @shahor02 , @davidrohr , 

This seems much better to me. Can you please check?

Chiara

chiarazampolli:
Hello @davidrohr , @shahor02 , 

Any comment about this PR?

Chiara

davidrohr:
I didn't test it, but fine with me

shahor02:
Fine for me

PR #1527, opened 2024-03-08
Adjusting the logic for the rate limiting in split wf

chiarazampolli:
@shahor02 
Now by default the rate limiting is the one of the full wf, also if we split. 
This means that it will depend on whether we run on CPU or GPU, and if we are in pp or PbPb.
Do you prefer that I give a fixed value instead?

### DIFF
--- a/DATA/production/configurations/asyncReco/async_pass.sh
+++ b/DATA/production/configurations/asyncReco/async_pass.sh
@@ -586,10 +586,15 @@ else
     echo "WORKFLOW_PARAMETERS=$WORKFLOW_PARAMETERS"
     echo "Step 3) matching, calib, AOD, potentially QC"
     echo -e "\nStep 3) matching, calib, AOD, potentially QC" >> workflowconfig.log
-    export TIMEFRAME_RATE_LIMIT=${ALIEN_JDL_TIMEFRAMERATELIMITSSPLITWF:-0}
+    # This uses the same time frame rate limiting as in full wf, unless differently specified in the JDL
+    export TIMEFRAME_RATE_LIMIT=${ALIEN_JDL_TIMEFRAMERATELIMITSSPLITWF:-${TIMEFRAME_RATE_LIMIT}}
     echo "Removing detectors $DETECTORS_EXCLUDE"
-    READER_DELAY=${ALIEN_JDL_READERDELAY:-30}
-    export ARGS_EXTRA_PROCESS_o2_global_track_cluster_reader+=" --reader-delay $READER_DELAY "
+    if [[ $ALIEN_JDL_USEREADERDELAY == 1 ]]; then
+      # if we add a delay, the rate limiting should be disabled
+      TIMEFRAME_RATE_LIMIT=0
+      READER_DELAY=${ALIEN_JDL_READERDELAY:-30}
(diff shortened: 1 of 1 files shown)


DATA/production/configurations/asyncReco/async_pass.sh
```diff
@@ -586,10 +586,15 @@ else
     echo "WORKFLOW_PARAMETERS=$WORKFLOW_PARAMETERS"
     echo "Step 3) matching, calib, AOD, potentially QC"
     echo -e "\nStep 3) matching, calib, AOD, potentially QC" >> workflowconfig.log
-    export TIMEFRAME_RATE_LIMIT=${ALIEN_JDL_TIMEFRAMERATELIMITSSPLITWF:-0}
+    # This uses the same time frame rate limiting as in full wf, unless differently specified in the JDL
+    export TIMEFRAME_RATE_LIMIT=${ALIEN_JDL_TIMEFRAMERATELIMITSSPLITWF:-${TIMEFRAME_RATE_LIMIT}}
```
shahor02, line 590: is TIMEFRAME_RATE_LIMIT guaranteed to be defined upstream?

DATA/production/configurations/asyncReco/async_pass.sh
```diff
@@ -620,8 +625,8 @@ else
       echo "WORKFLOW_PARAMETERS=$WORKFLOW_PARAMETERS"
       echo "Step 4) QC"
       echo -e "\nStep 4) QC" >> workflowconfig.log
-      export TIMEFRAME_RATE_LIMIT=${ALIEN_JDL_TIMEFRAMERATELIMITSSPLITWF:-0}
       echo "Removing detectors $DETECTORS_EXCLUDE"
+      echo "The rate limiting will be the same as in step 3: TIMEFRAME_RATE_LIMIT = ${TIMEFRAME_RATE_LIMIT}"
```
shahor02, line 629: Perhaps one should do 
```
      export TIMEFRAME_RATE_LIMIT=${ALIEN_JDL_TIMEFRAMERATELIMITSSPLITWF:-${TIMEFRAME_RATE_LIMIT}}
```
so that by default the same limit is used but we have a possibility to override it? Most probably, the same limit will work, just in case.

DATA/production/configurations/asyncReco/async_pass.sh
chiarazampolli: In principle, it should always be there. If not, then it will just be zero, the jobs will crash, and we will fix it.

DATA/production/configurations/asyncReco/async_pass.sh
chiarazampolli: This is the 4th step for QC, where again we use the reader. So I think it should be configured as the step above. If we add what you suggest, it will be like the previous step. If you want to differentiate it, we need another JDL variable. 
Since all this is not yet finalized, maybe we could for not leave it assuming step 3 and step 4 are to be configured in the same way?

shahor02: Fine for me, we can add special setting later, if needed.

chiarazampolli:
Thanks, merging, since it is not tested in the CI.

PR #1329, opened 2023-11-14
[WFRunner] Estimate resources dynamically

benedikt-voelkel:
* introduce functionality to update resource estimates based on values
  which are collected during monitoring
  (the same that go into pipeline_metric_<pid>.log)

* disentangle resource management from WorkflowExecutor

* manage resources via ResourceManager object
  * memory and cpu
  * semaphores
  * number of parallel tasks to be launched
  * niceness

* main interfaces of ResourceManager
  * ok_to_submit
  * book
  * unbook
  * add_monitored_resources

* group corresponding tasks together to be able to update resource
  estimates for upcoming tasks

* TaskResources as container of a task's resource estimate
  holding
  * resource estimates (initial and new estimates)
  * references to Semaphore object (optional
  * reference to list of related tasks (optional)

* make sure, final niceness assigned to a launched process is the same
  that the TaskResources hold for that process

* enable functionality with --dynamic-resources

* add more comments to classes and functions

* update --cgroup option to require entire path
  --> handle both cgroups v1 and cgroups v2

PR #1532, opened 2024-03-14
[WFRunner] Handle resource limits and CPU better

benedikt-voelkel:
* Account for relative CPU factor in case of sampling
  Studies have shown that being able to backfill tasks can have a
  difference for CPU efficiency, especially for transport.
  That is in particular important for high-efficiency jobs such as
  high-interaction-rate pp simulations.

* Abort by default if estimated resources exceed limits

* Run anyway, if --optimistic-resources is passed
  * Fix: Actually reset the overestimated resources to the limits as
    otherwise the runner would silently quit when nothing else can be
    done.

* In case of dynamically sampled resources and if a corresponding task
  has been run already:
  Reset the assigned resources to the limits if they exceed the
  boundaries.

### DIFF
--- a/MC/bin/o2_dpg_workflow_runner.py
+++ b/MC/bin/o2_dpg_workflow_runner.py
@@ -535,13 +535,16 @@ class TaskResources:
     """
     Container holding resources of a single task
     """
-    def __init__(self, tid, name, cpu, mem, resource_boundaries):
+    def __init__(self, tid, name, cpu, cpu_relative, mem, resource_boundaries):
         # the task ID belonging to these resources
         self.tid = tid
         self.name = name
         # original CPUs/MEM assigned (persistent)
         self.cpu_assigned_original = cpu
         self.mem_assigned_original = mem
+        # relative CPU, to be multiplied with sampled CPU; set by the user, e.g. to allow to backfill tasks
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
@sawenzel do you have ay comments here? If not, I would go ahead and merge.

sawenzel:
The commit message sounds reasonable. But is this necessary to test the dynamic system (just wondering how time-critical it is)?

benedikt-voelkel:
It's not necessary for testing right now.
It is mostly to get that fix in. Because before, one would `--optimistic-resources` (e.g. to run locally anyway although TPCDigi might have the initial estimate of 16GB but the local RAM is lower).
And to consolidate that a little bit more, I encapsulated the checking and limiting of resource boundaries a bit more.

Tested and works just fine.

benedikt-voelkel:
@chiarazampolli 
When ported, first, we also need to take
* https://github.com/AliceO2Group/O2DPG/pull/1329
* https://github.com/AliceO2Group/O2DPG/pull/1413

Anyway, first of all to be discussed with @sawenzel 

This is important to keep the CPU efficiency where it is in case it is already high by default. I will explain in an email.

PR #1533, opened 2024-03-14
qc-workflow: Do not try to write to InfoLogger at all if not in online mode

davidrohr:
To be merged once QC 0.138 becomes available


martenole (approved):
Thanks!

PR #1534, opened 2024-03-14
Update Lambda1520 pdg code

BongHwi:
From Pythia 8.303, The PDG code of Lambda(1520) has been updated to `102134` instead of `3124`

### DIFF
--- a/MC/config/PWGLF/pythia8/generator/resonancelist.gun
+++ b/MC/config/PWGLF/pythia8/generator/resonancelist.gun
@@ -12,8 +12,8 @@
 -3114       10     0.2     10      -1.2    1.2     genDecayed
 3224        10     0.2     10      -1.2    1.2     genDecayed
 -3224       10     0.2     10      -1.2    1.2     genDecayed
-3124        10     0.2     10      -1.2    1.2     genDecayed
--3124       10     0.2     10      -1.2    1.2     genDecayed
+102134      10     0.2     10      -1.2    1.2     genDecayed
+-102134     10     0.2     10      -1.2    1.2     genDecayed
 3324        10     0.2     10      -1.2    1.2     genDecayed
 -3324       10     0.2     10      -1.2    1.2     genDecayed
 10323       10     0.2     10      -1.2    1.2     genDecayed
--- a/MC/config/PWGLF/pythia8/generator/resonancelistgun.json
+++ b/MC/config/PWGLF/pythia8/generator/resonancelistgun.json
@@ -90,7 +90,7 @@
         "genDecayed": true
     },
     "Lambda(1520)0" : {
-        "pdg": 3124,
+        "pdg": 102134,
         "n": 1,
         "ptMin": 0.0,
         "ptMax": 20,
@@ -99,7 +99,7 @@
         "genDecayed": true
     },
     "anti-Lambda(1520)0" : {
-        "pdg": -3124,
+        "pdg": -102134,
         "n": 1,
         "ptMin": 0.0,
         "ptMax": 20,
--- a/MC/config/PWGLF/pythia8/generator/resonances.cfg
+++ b/MC/config/PWGLF/pythia8/generator/resonances.cfg
@@ -2,33 +2,33 @@
 ProcessLevel:all = off # will not look for the 'process'
 
 # id::all = name antiName spinType chargeType colType m0 mWidth mMin mMax tau0
-3124:all =  Lambda1520 Lambda1520bar 0 0 0 1.5195 0.01560 0.06240 0.00000E+00 0 1
+102134:all =  Lambda1520 Lambda1520bar 0 0 0 1.5195 0.01560 0.06240 0.00000E+00 0 1
 
 ### add Resonance decays absent in PYTHIA8 decay table and set BRs from PDG for other
-3124:oneChannel = 1 0.223547 0 2212 -321
-3124:addChannel = 1 0.223547 0 2112 -311
-3124:addChannel = 1 0.139096 0 3222 -211
-3124:addChannel = 1 0.139096 0 3212 111
-3124:addChannel = 1 0.139096 0 3112 211
-3124:addChannel = 1 0.028780 0 3224 -211
-3124:addChannel = 1 0.028780 0 3214 111
-3124:addChannel = 1 0.028780 0 3114 211
-3124:addChannel = 1 0.019373 0 3212 22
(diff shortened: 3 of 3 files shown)


benedikt-voelkel:
Hej @BongHwi 
good call! Thanks for spotting. In fact we use Pythia8@8.304 since months or even years now when building SW with `--defaults o2` as far as I can see.

benedikt-voelkel:
@BongHwi once merged, should this be ported to the code we use for PbPb MC productions?

BongHwi:
@benedikt-voelkel Yes, we'd like to apply this to O2-4365.

benedikt-voelkel:
The ticket you linked seems to be a different one or I don't see the connection to that ticket 🤔

BongHwi:
Oops.. yes. it was O2-4306

benedikt-voelkel:
ok, so it is for pp 2022 apass4. I added the label. In addition, I put PbPb 2023 because since it is generator functionality, it wouldn't hurt to have it there as well. If your disagree, please remove the label for that.

benedikt-voelkel (approved):

BongHwi:
Thanks a lot! We'll prepare the Pb-Pb production as well once we finish the pp production, so I guess it's an appropriate tag.

BongHwi:
Sorry @benedikt-voelkel , we found an issue with running the Pythia with this new PDG code.
Therefore, we'd like to revert this commit until we solve the issue.

In detail, the Pythia tells this error when we try to produce the event:
```
{"etaMax":1.2,"etaMin":-1.2,"genDecayed":true,"n":1,"pdg":102134,"ptMax":20,"ptMin":0.0}
===> Unknown particle requested with PDG 102134, mass set to 0
[01:03:19][FATAL] Could not find mass for mPdg 102134
libc++abi: terminating due to uncaught exception of type fair::FatalException
```

So it couldn't find the 102134 even though it was an 8.3 version of Pythia.
The definition of the particle is on the Pythia particle list: https://pythia.org/manuals/pythia8308/ParticleData.html

benedikt-voelkel:
@BongHwi The error does not seem to come from Pythia but ultimately from the fact that **we** do not have that PDG in our table.

It fails after running /MC/config/PWGLF/pythia8/generator_pythia8_longlived.C where it looks up the particle in `TDatabasePDG`.

So, I think, you should be able to fix it by adding that to 
* https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/simulation/include/SimulationDataFormat/O2DatabasePDG.h
* https://github.com/AliceO2Group/AliceO2/blob/dev/Steer/src/O2MCApplication.cxx#L234

BongHwi:
Great, thanks for the suggestion, I'll try to update these files too.

BongHwi:
Here I tried to update the O2 database.

https://github.com/BongHwi/AliceO2/commit/f6aa24b7daa0679cf40f41b0696960bcd24e847e

But it still shows an another issue as below:
```
[15:42:22][INFO] importParticles 0
[15:42:22][INFO] Printing particles that are appended
[15:42:22][INFO] Particle 0 is a 2224 with status -1610094078 and px = -14.2135 py = -8.90426 pz = 19.1223
[15:42:22][INFO] Particle 1 is a 2212 with status -1610566143 and px = -12.8402 py = -7.8437 pz = 17.0538
[15:42:22][INFO] Particle 2 is a 211 with status -1610566143 and px = -1.37331 py = -1.06056 pz = 2.0685
[15:42:22][INFO] Event generation took 0s and produced 3 primaries 
[15:42:22][INFO] GENEVENT CHANGING STATE TO SERVING
[15:42:22][INFO] ASSIGNED PIPE HANDLE 27
[15:42:22][INFO] INITTASK CHANGING STATE TO SERVING
[15:42:22][STATE] INITIALIZING TASK ---> READY
[15:42:22][STATE] READY ---> RUNNING
[15:42:22][INFO] fair::mq::Device running...
[15:42:23][INFO] INFO REQUEST RECEIVED
[15:42:23][INFO] Received config request
[15:42:23][INFO] config reply send 
[15:42:23][INFO] INFO REQUEST RECEIVED
[15:42:23][INFO] Received config request
[15:42:23][INFO] config reply send 
[15:42:30][INFO] Received device shutdown request (signal 15).
[15:42:30][INFO] Waiting for graceful device shutdown. Hit Ctrl-C again to abort immediately.
[15:42:30][ERROR] Some error/interrupt occurred on socket during receive
[15:42:30][INFO] CONDRUN CHANGING STATE TO STOPPED
[15:42:30][INFO] Waiting info thread
[15:42:30][STATE] RUNNING ---> READY
[15:42:30][STATE] READY ---> RESETTING TASK
[15:42:30][STATE] RESETTING TASK ---> DEVICE READY
[15:42:30][STATE] DEVICE READY ---> RESETTING DEVICE
[15:42:30][STATE] RESETTING DEVICE ---> IDLE
[15:42:30][STATE] IDLE ---> EXITING
[15:42:30][STATE] Exiting FairMQ state machine

```

Especially,
```
[15:42:30][ERROR] Some error/interrupt occurred on socket during receive
```

This error is coming from this line: https://github.com/AliceO2Group/AliceO2/blob/dev/run/O2PrimaryServerDevice.h#L438

If this issue is not trivial to solve, we would like to go back and produce the mc based on the old configuration to boost up the speed.

benedikt-voelkel:
Mh, is that reproducible?
It looks a bit odd that now the message received seems to be somewhat corrupt. Are you including any other updates in your setup besides this PR and the changes in O2 for the `TDatabasePDG`?

Otherwise, for now we should probably revert and investigate in more detail

BongHwi:
Yes, I tested twice and the results were the same.
Now I'm trying to perform the same test on another system since I don't trust my MacOS env.

I'll update the result as soon as I get it.

BongHwi:
I tested it on another system (Linux ubuntu 18.04) and.. it looks like it's working well...
So I have to say that it was an issue with the MacOS system.

I'll open the PR for the O2 update.

benedikt-voelkel:
ok, thanks for checking. Good and somewhat bad news at the same time.
But let's focus on the production system, which is Linux.
It might be worth to follow up on that, I will put something on my list. Honestly, I haven't run simulation on a Mac in a long time so I would need to see.

Once I get to it and if there is an issue, we will follow up in a ticket.

BongHwi:
+async-label async-2022-pp-apass6-2023-PbPb-apass2

PR #1535, opened 2024-03-15
PHS: modernize Ped and LED calibration workflows

sevdokim:
Hello! This PR is for modernization of PHS standalone calibration workflows, namely Pedestal and LED calibration.

### DIFF
--- a/DATA/production/calib/phs-led.sh
+++ b/DATA/production/calib/phs-led.sh
@@ -4,36 +4,67 @@ source common/setenv.sh
 
 source common/getCommonArgs.sh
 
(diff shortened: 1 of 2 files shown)


sevdokim:
hi @davidrohr 
when can we expect to have this PR at P2? BTW, this is not urgent.

davidrohr:
With the next SW update, or whenever O2DPG is bumped at P2 for another reason.
Meanwhile, you can already use it by passing in a custom O2DPG hash in the EPN Workflow Expert panel instead of the "default" keyword. (But please do not save such a configuration with a non-default setting!)

PR #1537, opened 2024-03-15
Get all labels and change table organisation

benedikt-voelkel:
Logic is now the following:

1. Fetch all labels (optionally given some regex).
2. Get closed PRs that have at least one of those labels.
3. Split into merged and not merged.
4. Populate a table per label and list merged PRs. (A PR can therefore appear in more than one table.)
5. Write everything to a txt file.


benedikt-voelkel:
No tests will be run, changing an unrelated utility. Merging

PR #1539, opened 2024-03-18
Prototype decouple event gen from transport

sawenzel:
(Optional) decoupling of event gen from transport
Development allowing event generation to run in a separate task
from detector simulation. In this mode, event generation will write
kinematics to disc, which is later picked up by o2-sim.

Development done in order to increase CPU efficiency in cases
where event generation was slow and not being able to serve the workers
of o2-sim with enought work. The development fixes such cases, because
multiple event generators (for different timeframes) can now run in parallel.

The new mode is the default, but we can return to the intregrated mode
by using option `--event-gen-mode integrated`.


sawenzel:
Needs https://github.com/AliceO2Group/AliceO2/pull/12899 and https://github.com/AliceO2Group/AliceO2/pull/12891 to work properly.

sawenzel:
+async-label async-2022-pp-apass6

sawenzel:
Also needs https://github.com/AliceO2Group/AliceO2/pull/12920

PR #1542, opened 2024-03-21
Adding missing input

chiarazampolli:
@aferrero2707 : at present, master + dev are broken in async reco, I guess due to https://github.com/AliceO2Group/QualityControl/pull/2150

### DIFF
--- a/DATA/production/qc-async/mch-tracks.json
+++ b/DATA/production/qc-async/mch-tracks.json
@@ -33,7 +33,7 @@
                 "maxNumberCycles": "-1",
                 "dataSource": {
                     "type": "direct",
-                    "query": "trackMCH:MCH/TRACKS;trackMCHROF:MCH/TRACKROFS;trackMCHTRACKCLUSTERS:MCH/TRACKCLUSTERS"
+                    "query": "trackMCH:MCH/TRACKS;trackMCHROF:MCH/TRACKROFS;trackMCHTRACKCLUSTERS:MCH/TRACKCLUSTERS;mchtrackdigits:MCH/TRACKDIGITS"
                 },
                 "taskParameters": {
                     "maxTracksPerTF": "600",


DATA/production/qc-async/mch-tracks.json
```diff
@@ -33,7 +33,7 @@
                 "maxNumberCycles": "-1",
                 "dataSource": {
                     "type": "direct",
-                    "query": "trackMCH:MCH/TRACKS;trackMCHROF:MCH/TRACKROFS;trackMCHTRACKCLUSTERS:MCH/TRACKCLUSTERS"
+                    "query": "trackMCH:MCH/TRACKS;trackMCHROF:MCH/TRACKROFS;trackMCHTRACKCLUSTERS:MCH/TRACKCLUSTERS;mchtrackdigits:MCH/CLUSTERDIGITS"
```
aferrero2707, line 36: It should be `;mchtrackdigits:MCH/TRACKDIGITS` instead of `;mchtrackdigits:MCH/CLUSTERDIGITS`

aferrero2707:
@chiarazampolli thanks a lot for the fix! In future I will be more careful in checking the ASYNC configuration when making changes to the QC.
I just left a comment with a correction for the missing DPL input.

DATA/production/qc-async/mch-tracks.json
chiarazampolli: Thanks!! Can you check now, @aferrero2707 ?

chiarazampolli: Thanks, I think I took this from the sync-qc, maybe the wrong place, though (but the label was the correct one). Merging now.

chiarazampolli:
As commented in the resolution of Andrea's feedback, I took the input from `https://github.com/AliceO2Group/O2DPG/blob/31853f570c1178aa5040e9d0f946ded0eaeeb7ba/DATA/production/qc-sync/glo-mchmid-mtch-qcmn-epn.json#L36`. @aferrero2707 , maybe for the sync it is correct, but can you check?

Anyway, this is now fixed, merging.

chiarazampolli:
Hello @aferrero2707 , 
Unfortunately, master is broken, I wonder if your suggestion was the right one... We get:
```
[ERROR] invalid workflow in o2-qc: No matching output found for <matcher query: (and origin:MCH (and description:TRACKDIGITS (just startTime:$0 )))> as requested by data processor "qc-task-MCH-Tracks". Candidates:
```
With my initial change, it was working. Can you please check? I am also recompiling dev/master and will try again, but at the moment, master does not work.
Chiara

chiarazampolli:
Re-fixed in https://github.com/AliceO2Group/O2DPG/pull/1564

PR #1543, opened 2024-03-21
[SimWF] Recompute number of workers used in TFs

benedikt-voelkel:


benedikt-voelkel:
@sawenzel note that at the moment, the argument that is passed with `-j <workers>` has no effect at this point.
Of course, what we can do is to simply take it if provided by the user and do not re-compute the number of workers in such cases.
However, that brings some complication that at the moment we set the number of workers usually in `anchorMC.sh`.
What we could do is to remove it there, so that always the default value of 8 can be taken for anchored (GRID) MCs.

benedikt-voelkel:
Actually I just adapted the logic according to my precious comment: The re-computation of the number of workers is only done if `--pregenCollContext` is set. Otherwise, up to the user to choose

PR #1546, opened 2024-03-21
[SimWF,Anchor] Allow to force collision system and CM energy

benedikt-voelkel:


benedikt-voelkel:
Should all be doable via the method explained here: https://aliceo2group.github.io/simulation/docs/o2dpgworkflow/anchored.html#run-pythia-with-a-different-collision-system

PR #1556, opened 2024-03-26
[SimWF] Set MCANCHOR to value of LPMANCHORPASSNAME

benedikt-voelkel:


benedikt-voelkel:
x-checked with @catalinristea

PR #1573, opened 2024-04-03
Remove EventTrackQA configurations, keeping default ones.

mfaggin:


mfaggin:
@benedikt-voelkel not sure how to label it, I understand that you want it asap regardless the reco pass, period and year right? Then shall I tag it with

`+async-label async-data, async-mc`

benedikt-voelkel:
Dear @mfaggin 
we recently changed the logic of the AnalysisQC quite a bit. Could you please have a look? https://github.com/AliceO2Group/O2DPG/pull/1590
The configs are split into the single once corresponding to their executable.
Let me know if there are specific requirements or changes. Just contact me e.g. via mattermost in that case.
Thanks a lot!

benedikt-voelkel:
The configuration files that are removed here do anyway not exist anymore.
Closing.

PR #1574, opened 2024-04-03
Fixes for isolated QC step

shahor02:
Goes with https://github.com/AliceO2Group/AliceO2/pull/12982
Still some problem with GLO-MuonTracks QC, don't merge


shahor02:
tested locally.

PR #1581, opened 2024-04-07
PWGHF: add possibility to use different config for MB events in gap trigger

fgrosa:
This PR adds the possibility to use a different config for MB events in the gap trigger HF generator. This is useful as alternative to the embedding with replacement when for example using `HardQCD` for the signal events. The implementation is similar to the DQ gap trigger generator (see [generator_pythia8_NonPromptSignals_gaptriggered_dq.C](https://github.com/AliceO2Group/O2DPG/blob/23d985d220f4020bc055499e3c2eb01dd0f502b3/MC/config/PWGDQ/external/generator/generator_pythia8_NonPromptSignals_gaptriggered_dq.C))
I tested it (currently for unanchored MCs) but @fcatalan92 could you also please cross check that it is fine? Thanks!

One note: I noticed that passing the second config for the PYTHIA instance for MB events was resetting the decay table also of the main PYTHIA instance (the one for signal events). It might be that the decay table is global and shared among PYTHIA instances, hence I set the same decay table for signal and background events (as done in the usual HF gap trigger configs).

### DIFF
--- a/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
+++ b/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
@@ -64,8 +64,8 @@ public:
     addSubGenerator(4, "Charm injected");
     addSubGenerator(5, "Beauty injected");
     bool initAltOk{true};
-    bool initMainOk = o2::eventgen::GeneratorPythia8::Init();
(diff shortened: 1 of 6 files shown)


fgrosa:
I am closing this PR, which is not needed anymore thanks to the fix in this other PR https://github.com/AliceO2Group/O2DPG/pull/1587

PR #1584, opened 2024-04-08
Step 3 should keep the root output in split wf mode

chiarazampolli:
@shahor02 
Locally you don't see it since by default it keeps all root output.


chiarazampolli:
Tested locally, merging.

PR #1587, opened 2024-04-09
Fix options for collision context creation in (old) embedding path

sawenzel:
did not take into account first-orbit, which is used in anchoring MC.


sawenzel:
+async-label async-2022-pp-apass4, async-2023-pbpb-apass, async-2023-pp-apass1, async-data async-mc, async-2022-pp-apass6-2023-PbPb-apass2

benedikt-voelkel (approved):

PR #1590, opened 2024-04-10
[AnalysisQC] Change configuration logic

benedikt-voelkel:
* disable EMCAL analysis
  (to enable, need to revise resource needs, crashes or hangs
   sporadically)

* split config JSONs into executable-specific single JSONs Most of those are seen as configurations for "service wagons". Each analyses is forced to use those comon configurations and it is not possible to have analysis-specific configurations for service wagons.

* possible to use special configuration strings in configs; this is done to avoid duplications of configs
  * "!ANALYSIS_QC_is_mc!" ("!ANALYSIS_QC_is_data!") will be replaced with "true"("false") or "false"("true") depending on whether the analyses are run on MC(data)

* new default is to pipe all analyses together; to split into single analyses, run o2dpg_analysis_testing_workflow.py [...] --split-analyses


benedikt-voelkel:
@chiarazampolli @noferini @catalinristea 
The default is now to merge all analyses together, see also the commit message.

At this point, there is no difference in configurations for pp and PbPb, it's all the same.
Tested on
* pp MC
* PbPb MC

Test on data ongoing
(`/alice/data/2023/LHC23zs/539580/apass4_test_dev_0329/0340/o2_ctf_run00539580_orbit0275138592_tf0001956972_epn108/003/AO2D.root`)

PR #1593, opened 2024-04-12
[EMCAL-1118] Make QC workflow mode settable

mfasDa:
Allow adjustable QC workflow as stand-alone or
multi-node workflow with separate configurations.
In addition minor fixes in the input proxy configuration.


mfasDa (commented):
Tested on staging

PR #1595, opened 2024-04-12
Add MC/config for ALICE3 otf simulations

jesgum:
This pull request adds the `MC/config/ALICE3` directory to be used for on-the-fly simulations on hyperloop.
Contains config and ini files for different beams. Is currently being tested and will be marked ready very shortly. Input is already welcome now, thanks!


ddobrigk:
@benedikt-voelkel @sawenzel this is to be used with the OTF simulations on hyperloop: we'd need the `ini` files to be with an O2DPG tag. Can you please take a look and if ok carry on? Many many thanks! tagging also: @mconcas @jgrosseo

sawenzel:
Fine by me. Maybe we could just add a `README.md` in the `MC/config/ALICE3` saying that these configs are used in Hyperloop studies, etc?

ddobrigk:
Yes that's very reasonable. @jesgum can you please take care of that? Thank you!

jesgum:
It has now been added to the directory.

sawenzel:
Good. I think now you need remove the draft status.

sawenzel (approved):

benedikt-voelkel:
For this use case it would be fine to have dummy tests for the specific generators for now.
In any case, one basic test is anyway ran over the kinematics.

What is your opinion on that @sawenzel ?

sawenzel:
A CI test would be a "nice to have". But since this is ALICE3 R&D, it is probably not obligatory for now.

benedikt-voelkel (changes requested):

benedikt-voelkel (approved):

PR #1597, opened 2024-04-15
[Anchor] Use STF and ETF

benedikt-voelkel:
Use BasicCCDBManager's getRunDuration to determine time range to be covered by simulation.

This will first try to get STF and ETF. If not available, the fallback chain is
1. SOX and EOX
1. SOR and EOR

### DIFF
--- a/MC/bin/o2dpg_sim_workflow_anchored.py
+++ b/MC/bin/o2dpg_sim_workflow_anchored.py
@@ -67,28 +67,15 @@ class CCDBAccessor:
 
         return timestamp, obj
 
+    def get_run_duration(self, run_number):
+       return o2.ccdb.BasicCCDBManager.instance().getRunDuration(run_number)
+
     def fetch_header(self, path, timestamp=None):
         meta_info = std.map["std::string", "std::string"]()
         if timestamp is None:
             timestamp = -1
         header = self.api.retrieveHeaders(path, meta_info, timestamp)
         return header
-
-
-
-def retrieve_sor_eor(ccdbreader, run_number):
-    """
-    retrieves start of run (sor) and end of run (eor) given a run number
-    from the RCT/Info/RunInformation table
-    """
-
-    path_run_info = "RCT/Info/RunInformation"
-    header = ccdbreader.fetch_header(path_run_info, run_number)
-    if not header:
-       print(f"WARNING: Cannot find run information for run number {run_number}")
-       return None
-    # return this a dictionary
-    return {"SOR": int(header["SOR"]), "EOR": int(header["EOR"])}
 
 
 def retrieve_CCDBObject_asJSON(ccdbreader, path, timestamp, objtype_external = None):
@@ -111,7 +98,7 @@ def retrieve_CCDBObject_asJSON(ccdbreader, path, timestamp, objtype_external = N
     jsonTString = TBufferJSON.ConvertToJSON(obj, TClass.GetClass(objtype))
     return json.loads(jsonTString.Data())
 
-def retrieve_params_fromGRPECS(ccdbreader, run_number, rct = None):
+def retrieve_params_fromGRPECS(ccdbreader, run_number, run_start=None, run_end=None):
     """
     Retrieves start of run (sor), end of run (eor) and other global parameters from the GRPECS object,
     given a run number. We first need to find the right object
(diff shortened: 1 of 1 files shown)


MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -150,22 +137,18 @@ def retrieve_params_fromGRPECS(ccdbreader, run_number, rct = None):
...
-       # verify that the variaous sor_eor information are the same
-       if SOR != rct["SOR"]:
-         print ("WARNING: Inconsistent SOR information on CCDB (divergence between GRPECS and RCT) ... will take RCT one")
-         SOR=rct["SOR"]
-
-       if EOR != rct["EOR"]:
-         print ("WARNING: Inconsistent EOR information on CCDB (divergence between GRPECS and RCT) ... will take RCT one")
-         EOR=rct["EOR"]
+    if run_start is not None:
+       print (f"INFO: GRPECS SOR ({SOR}) will be superseded by externally provided run start ({run_start})")
```
chiarazampolli, line 147: This is not anymore from GRPECS, right?

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -150,22 +137,18 @@ def retrieve_params_fromGRPECS(ccdbreader, run_number, rct = None):
...
+       SOR = run_start
+    if run_end is not None:
+       print (f"INFO: GRPECS EOR ({EOR}) will be superseded by externally provided run end ({run_end})")
```
chiarazampolli, line 150: As above.

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -329,34 +311,21 @@ def main():
     # make a CCDB accessor object
     ccdbreader = CCDBAccessor(args.ccdb_url)
     # fetch the EOR/SOR
-    rct_sor_eor = retrieve_sor_eor(ccdbreader, args.run_number) # <-- from RCT/Info
-    GLOparams = retrieve_params_fromGRPECS(ccdbreader, args.run_number, rct=rct_sor_eor)
+    run_duration = ccdbreader.get_run_duration(args.run_number)
+    run_start = run_duration.first
+    run_end = run_duration.second
+
+    GLOparams = retrieve_params_fromGRPECS(ccdbreader, args.run_number, run_start=run_start, run_end=run_end)
```
chiarazampolli, line 318: Is it needed to pass run_start and run_end here?

MC/bin/o2dpg_sim_workflow_anchored.py
benedikt-voelkel: Yes, as seen by this function, the value from `GRPECS` will be overridden.
We can simply do that silently. Or make the output more clear that all the messages at this stage come from processing the `GRPECS`

benedikt-voelkel:
I see you comments @chiarazampolli and mostly, they seem to go in the direction of "what's printed might be confusing". Maybe there could be some `START processing GRPECS` and `END processing GRPECS`.
Or we simply make those steps silently and do not print anything about it to stdout.

MC/bin/o2dpg_sim_workflow_anchored.py
chiarazampolli: I understand, but isn't run_start and run_end from RCT? Why GRPECS is used at all to retrieve the SOR?

MC/bin/o2dpg_sim_workflow_anchored.py
chiarazampolli: Can we simply not avoid to retrieve it from GRPECS now?

chiarazampolli:
Hello @benedikt-voelkel , 
What can be confusing is accessing GRPECS to retrieve start and stop, since now only RCT should be used, right? (apart from overriding, of course).
Chiara

MC/bin/o2dpg_sim_workflow_anchored.py
benedikt-voelkel: We get the orbit info from that. I asked @shahor02 specifically via email if we should stick to that method and as far as I understand, that should be the way. So we need to process `GRPECS` for that reason.

chiarazampolli: Yes, but you can avoid retrieving SOR and EOR, this is what I mean. They will not be consistent with RCT.

chiarazampolli: Thanks! This is what I meant.

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -141,43 +129,25 @@ def retrieve_params_fromGRPECS(ccdbreader, run_number, rct = None):
...
-    if rct != None:
-       # verify that the variaous sor_eor information are the same
-       if SOR != rct["SOR"]:
-         print ("WARNING: Inconsistent SOR information on CCDB (divergence between GRPECS and RCT) ... will take RCT one")
-         SOR=rct["SOR"]
-
-       if EOR != rct["EOR"]:
-         print ("WARNING: Inconsistent EOR information on CCDB (divergence between GRPECS and RCT) ... will take RCT one")
-         EOR=rct["EOR"]
-
```
benedikt-voelkel, line 169: I took all these things out. Now, simply the given `run_start` and `run_end` are used where needed.

benedikt-voelkel:
@chiarazampolli What do you think? Can we move on?
I left `run_start` and `run_end` for the now called `def retrieve_params_fromGRPECS_and_OrbitReset` function as they are needed for the orbit calculation in there.

sawenzel (approved):

PR #1603, opened 2024-04-23
[ALICE3] Add empty external generator and adjust ini files to adhere to hyperloop workflow

jesgum:
This purpose of this pull request is to allow the use of the default pythia8 generator without changing the command specifying which generator to use from "external" to desired generator on hyperloop.


ddobrigk:
Hi @benedikt-voelkel @sawenzel - this is a requirement to keep the hyperloop calls fully identical also in future use cases. Can you please take a look and if ok for you go ahead and merge? Many thanks!

PR #1606, opened 2024-04-24
gen_topo.sh: Use /var/tmp/gen_topo as gen_topo temp folder

davidrohr:
@martenole : Could we deploy this with the next software update?

It moves the gen_topo temp folder to a local folder on the infra node.
Not sure if you'll need to adapt any of your fetch/check scripts.


martenole (approved):
Looks good, I will include a new GenTopo for the next sw update. The log fetcher only queries `/var/log/topology/gen-topo.log`, so I don't have to change anything there

davidrohr:
@martenole : I think this was not yet deployed at P2? Will you check, or shall I follow up with Christian?

PR #1607, opened 2024-04-24
Remove option ctp-readout-create

mfasDa:
- ctp readout create does not properly work in
  2023 as only 1 trigger class per trigger is
  expeected
- By design it is impossible to recover the trigger
  classes from the trigger records as they are not
  encoded in the trigger record, so the min. bias trigger
  is assumed. This does not work for triggers that
  provide triggers themselves.

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -1433,9 +1433,6 @@ for tf in range(1, NTIMEFRAMES + 1):
    if args.no_strangeness_tracking:
       AODtask['cmd'] += ' --disable-strangeness-tracker'
 
-   # Enable CTP readout replay for triggered detectors (EMCAL, HMPID, PHOS/CPV, TRD)
-   # Needed untill triggers are supported in CTP simulation
-   AODtask['cmd'] += ' --ctpreadout-create 1'
    workflow['stages'].append(AODtask)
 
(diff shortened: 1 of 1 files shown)


mfasDa:
+async-label async-2023-pbpb-apass3, async-2023-pbpb-apass4, async-2022-pp-apass6-2023-PbPb-apass2, async-2022-pp-apass4-accepted, async-2023-pbpb-apass3-accepted, async-2023-pbpb-apass4-accepted

benedikt-voelkel:
Thanks for the labels @mfasDa . I just removed the `accepted` ones.
They will be added once that is actually in. Adding those will actually not be possible in the very near future ;)

Could you just check once more if your labels are correct now?

chiarazampolli:
Hello @mfasDa , 
I thought we discussed to first ask TRD, HMP, PHS, CPV, did I misunderstand?
Chiara

mfasDa:
@chiarazampolli I have opened the PR for review from the other triggered detectors. I added a WIP label to prevent merging.

chiarazampolli:
Merging. 
I guess we should add also 2024 labels, right, @mfasDa ?

chiarazampolli:
Wait, before merging, I will let the CI run.

mfasDa:
Right, we need it also for the MCs anchored to 2024.

PR #1608, opened 2024-04-25
PWGLF: Update the number of injected particle for PbPb

BongHwi:
So far, we intend to inject only one particle per event.

But in PbPb collision, we could increase this to save the computing resources.

We updated the number of injected particles in the PbPb injector case.


BongHwi:
+async-label async-2022-pp-apass6-2023-PbPb-apass2 async-2023-pbpb-apass3 async-2023-pbpb-apass4

njacazio (approved):

PR #1616, opened 2024-05-03
ALICE3: Allow external generator to accept seed during initialization

jesgum:
Since we specify the use of an external generator on hyperloop the o2-sim-dpl-eventgen task does not accept a seed given using the "--seed" command line option and we instead have to give the seed manually when we initialize the generator.


ddobrigk:
@benedikt-voelkel @sawenzel this is the last piece (I hope) required for proper operation of the on-the-fly MC trains in hyperloop. When you get a moment, can you please take a look? Thank you!

sawenzel (approved):

sawenzel:
Fine for me. (Ideally we should try to make this more frameworked by doing in the core O2 Pythia8 class. Will try to think about how this can be done).

ddobrigk:
> Fine for me. (Ideally we should try to make this more frameworked by doing in the core O2 Pythia8 class. Will try to think about how this can be done).

Yeah indeed, that would be good - in fact, I wonder if we are actually using OK seeds in all generators deriving from GeneratorPythia8 without this? Maybe the default is `0` and set-per-clock but ideally we should also change that... Thanks a lot!

PR #1618, opened 2024-05-06
PHS: add DCS CCDB populator to pedestal calibration workflow

sevdokim:
Hi! This commit adds DCS CCDB populator to PHS pedestal calibration workflow.


sawenzel:
@davidrohr @sevdokim : Is this still relevant? Otherwise let's close it.

PR #1619, opened 2024-05-07
Add simple coalescence generator for He3

mpuccio:
This generator can be used to study the signal loss for nuclei analyses in pp collisions.
For sake of simplicity, currently only He3 is supported but I will then extend it to other nuclei as well.


njacazio (approved):
Thanks a lot @mpuccio !

PR #1620, opened 2024-05-08
Add CTF size QC and pass active detector list

martenole:
PR for QC follows later. This passes the list of active detectors to the QC task so that it can publish only the histograms for the active detectors

### DIFF
--- a/DATA/production/qc-workflow.sh
+++ b/DATA/production/qc-workflow.sh
@@ -117,6 +117,9 @@ elif [[ -z ${QC_JSON_FROM_OUTSIDE:-} ]]; then
     elif has_detectors_reco MCH MID && has_matching_qc MCHMID; then
         [[ -z "${QC_JSON_GLO_MCHMID:-}" ]] && QC_JSON_GLO_MCHMID=consul://o2/components/qc/ANY/any/glo-mchmid-mtch-qcmn-epn
     fi
+    if has_processing_step ENTROPY_ENCODER && [[ ! -z "$WORKFLOW_DETECTORS_CTF" ]] && [[ $WORKFLOW_DETECTORS_CTF != "NONE" ]]; then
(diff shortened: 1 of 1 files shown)


DATA/production/qc-workflow.sh
```diff
@@ -259,6 +262,12 @@ elif [[ -z ${QC_JSON_FROM_OUTSIDE:-} ]]; then
     fi
   done
 
+  # CTF QC
+  if [[ ! -z "${QC_JSON_CTF_SIZE:-}" ]]; then
+    add_QC_JSON GLO_CTF ${QC_JSON_CTF_SIZE}
+    add_pipe_separated QC_DETECTOR_CONFIG_OVERRIDE '.qc.tasks.CTFSize.taskParameters.detectors=${WORKFLOW_DETECTORS}'
```
davidrohr, line 268: probably doesn't matter, but shouldn't this be WORKFLOW_DETECTORS_CTF?

martenole: all I want to achieve is that plots for one detector are not published by different runs, so WORKFLOW_DETECTORS is ok. In case for one detector CTF storage is disabled then the plots would show zero size which is fine I think.

martenole:
This should be tested in production first and could then be merged. On staging I did not manage to activate the GLO QC task on the QC merger node. The QC tasks on the EPNs ran fine. A good opportunity would be synthetics run, where this could be used from a local EPN checkout and if there are no problems this is good to go

chiarazampolli (approved):

rmunzer:
> This should be tested in production first and could then be merged. On staging I did not manage to activate the GLO QC task on the QC merger node. The QC tasks on the EPNs ran fine. A good opportunity would be synthetics run, where this could be used from a local EPN checkout and if there are no problems this is good to go

I did a test with replay in production. Seems to run fine.

chiarazampolli:
Thanks, since @rmunzer confirmed it is fine, I am merging.

Chiara

PR #1623, opened 2024-05-13
MC/PWGEM: HF Cocktail

DanielSamitz:
Generator and configs for LMee HF cocktail. The generator is just Pythia8+DecayerPythia8 but propagates the particle weights < 1 that arise from forcing certain decay channels (which Pythia itself does not do).


sawenzel (approved):

benedikt-voelkel:
@DanielSamitz 
I would assume this change need to be propagated to production software, right? If so, could you add corresponding labels to this PR?
Thanks a lot!

DanielSamitz:
Hi @benedikt-voelkel,
this generator is actually meant to be used for on-the-fly MC analysis. Of course in principle it is more general, because it  does nothing else than pure Pythia8 + DecayerPythia8 for user set hadrons that will be forced to decay. But that's the only thing I have in mind with it for now.

benedikt-voelkel:
Ok, thanks!
In that case, it would not need to go into any special tag.

PR #1624, opened 2024-05-13
TPC: Update stochastic cluster and systematic track errors

wiechula:


chiarazampolli:
@shahor02 , looks ok to me, could you also take a look?

chiarazampolli:
Ok, then I will wait for green light before merging.

wiechula:
In my local test everything is propagated properly to the workflow. The final validation if not all parameters set as default are as we expect is still missing. But this is not related to this PR.

chiarazampolli:
Hello @wiechula , @shahor02 , 

Did you run the final tests? Can I merge?

Chiara

shahor02:
Yes, checking results now, will finish soon

shahor02:
Hi,
here are the results of reco with the errors of this PR. The tracks are 


Track lengths in low and high rate runs (normalized per interaction):
NTPCcl

The tracks are significantly shorter than in my old reconstruction with `GPU_rec_tpc.clusterError2AdditionalY=0.1;GPU_rec_tpc.clusterError2AdditionalZ=0.15` (and no special error on seeding):
highIR_nclOldNew
I guess it is expected with the stricter errors.

We also have ~4% less matches, though it should be mostly due to the effectively stricter (due to the stricter cluster errors) matching chi2 cut (matches in the same 39kHz run, normalized per interaction):
highIR_mtcOldNew

Still, looking at the plots below, I would not change the matching chi2 (currently set to 80), since from the flattening of the matching chi2 distribution it is clear that large values are totally dominated by fakes (or badly reconstructed correct track with no use)

ITS-TPC matching (no afterburner) chi2 distribution in low and high rate runs (normalized per interaction):
matchChi2

In any case, the matches with chi2>60 contain practically no signal:
Mpipi (K0) signal (ITS-strandalone V0s removed) vs worst matching chi2 of 2 prongs in low and high rate runs (normalized per interaction):
k0vschisel

chiarazampolli:
Hello @shahor02 , 

Thanks! So if I understand your comments, all is as expected. @wiechula  is there something else we should check? You wanted too to run some tests, right?

Chiara

wiechula:
@chiarazampolli , the tests were run, I'm still waiting for feedback from @miranov25

chiarazampolli:
Hello,

I then propose to merge and start. Let me know if you have anything against it.

Chiara

shahor02:
I would wait for response from @miranov25

miranov25:
Hello all.

In the scan at GSI we got 5-6 % smaller cluster track association as the error estimator is smaller - and chi2 is closer to the chi2.

In my plot below I used only tracks with ITS, but the factor in the number of clusters is similar, in my parameterization and Ruben plots


image

chiarazampolli:
Hello @miranov25 , 

Please comment the plots above: I assume top is with the new cl error param, e bottom without, right?

I am now merging to move to test mode. 

Chiara

chiarazampolli (approved):

miranov25:
Hello.
Uper - is new error paramterization
middle - reference (old) parameterization
bottom - is the ratio

miranov25:
Hello @shahor02,

Regarding your statement about the K0s and chi2 cut, I understand your points and agree with them.

However, as we discussed over the past weeks, and especially last Wednesday, the TPC-ITS matching should be modified. I will attempt to implement a better TPC-ITS matching strategy utilizing max likelihood, taking background into account, and using checking causality for shorter TPC track (allowing them) ....
**There are many aspects that can be improved, but even the initial step—adjusting the cut between the last padrow and the first allowed padrow—will significantly enhance the situation.**

Marian

shahor02:
I think we should try to understand why the tracks fail to use lower pad-rows and try to recover them. David provided me with the settings which can affect tracking at lower pad-rows, but it will take time to test this.

PR #1627, opened 2024-05-14
MC skimmed data reducing sampling from 0.1 to 0.01

miranov25:
See discussion in https://its.cern.ch/jira/browse/O2-4829?focusedId=6379994&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6379994


chiarazampolli (approved):

chiarazampolli:
Hello @miranov25 , 

I guess you would like this for all upcoming tags, right? How urgent is this? We would need to create new tags, so we should know for which it is crucial to have it.

Chiara

miranov25:
Hello [Chiara Zampolli](https://github.com/jira/secure/ViewProfile.jspa?name=zampolli) 

I would like to have it as soon as possible, otherwise the output size is too big to process and store.

Marian

miranov25:
OHello @chiarazampolli 
I will try.

PR #1629, opened 2024-05-15
Fixes for workflow-multiplicities and setVarsFromALIEN_PROC_ID scripts

shahor02:

PR #1630, opened 2024-05-15
Define DCSCCDBSERVER_PERS env.var for persistent DCS CCDB server URL

shahor02:


benedikt-voelkel:
@shahor02 could you label this in case it should be ported? @mpuccio just said in an email thread this is a fix that seems to be needed in production.

mpuccio:
Hi @benedikt-voelkel, I was actually referring to #1629, that is necessary for the skimming, for which I am specifying manually the O2DPG tag on top of the O2PDPSuite one (as I need also another modification that is not present in the tag used for apass7). I am not sure if the modification in #1629 is necessary in production.

benedikt-voelkel:
Right, thanks for the clarification!

shahor02:
Hi, no, this is for online calibrations only.

PR #1642, opened 2024-05-21
ALICE3: Correcting printout

jesgum:
Correcting the printout for the seed used by the generator.


ddobrigk:
Hi @benedikt-voelkel @sawenzel can we please still fix this very simple printout mishap ? I think it's currently the last remaining bit before we go ahead in a more systematic way :-D thanks!

PR #1648, opened 2024-05-24
[ALICE3] Add configuration files for pythia8 pp

jesgum:
Hi! With this pull request we want to add some different pp configurations to be used for OTF simulations on hyperloop


jesgum:
Hi @benedikt-voelkel @sawenzel ! We would like to add some pp configurations for the otf simulations on hyperloop. Could you please have a look? Thanks a lot! Also tagging @ddobrigk

PR #1650, opened 2024-05-24
[SimWF,WIP] Enable MaterialManager input from CCDB

benedikt-voelkel:
* first demonstrator


benedikt-voelkel:
@sawenzel this works fine for me.

Next should be uploading a material parameter file to the production CCDB and then change this demonstrator accordingly to use the correct paths.

Once done, I would say this can be merged.

PR #1652, opened 2024-05-28
Tune workflow multiplicities for 4GPU Pb-Pb and pp async workflows

davidrohr:


davidrohr:
@chiarazampolli : This should improve the throughput of async Pb-Pb processing on the EPNs.
Now that ITS tracking needs less memory and sec vtx needs less CPU, we have to run more ITS trackers in parallel to fully load all CPU cores.

PR #1655, opened 2024-05-31
Prepare QC json fetching from apricot

martenole:
I still need to do some tests, thus marked as draft.


Barthelemy:
seems good to me (although I did not test it)

Barthelemy (approved):

rmunzer:
> seems good to me (although I did not test it)

Hi Ole,
I did serveral test with tpc standlone workflows and also with global run in staging. The qc files could be loaded by apricot.

PR #1657, opened 2024-06-03
PWGEM: Lmee add generator flags

hscheid:
Add flags to identify different external generators.

@feisenhu @rbailhac @dsekihat could you check that is implemented as we discussed?


feisenhu:
Thank you very much for the implementation.
The changes look good to me.

ivorobye (approved):

PR #1666, opened 2024-06-12
PWGHF: Pb-Pb generator to embed N HF events in a single underlying event.

mfaggin:


mfaggin:
@fcolamar @fcatalan92 @fgrosa @DelloStritto @stefanopolitano here what I was discussing with you before

MC/config/PWGHF/external/generator/generator_pythia8_embed_hf.C
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  return true;
+}
+
+private:
+
+    Generator* mGeneratorEvHF; // to generate HF signal events
+    
+    int mNumSigEvs; // number of HF signal events to be merged in one Pythia event
```
fgrosa, line 140: I would put a default value different from `0`, since I am not sure that the test works properly otherwise.
```suggestion
    int mNumSigEvs{1}; // number of HF signal events to be merged in one Pythia event
```

MC/config/PWGHF/ini/tests/GeneratorHF_D2H_ccbar_and_bbbar_PbPb.C
```diff
@@ -0,0 +1,129 @@
...
+    //std::cout << Form("# events injected with %d quark pair: ", checkPdgQuarkTwo) << nEventsInjTwo << "\n";
+    std::cout << Form("# %d (anti)quarks: ", checkPdgQuarkOne) << nQuarksOne << "\n";
+    std::cout << Form("# %d (anti)quarks: ", checkPdgQuarkTwo) << nQuarksTwo << "\n";
+    std::cout <<"# signal hadrons: " << nSignals << "\n";
+    std::cout <<"# signal hadrons decaying in the correct channel: " << nSignalGoodDecay << "\n";
+
+    if (nEventsMB < nEvents * (1 - ratioTrigger) * 0.95 || nEventsMB > nEvents * (1 - ratioTrigger) * 1.05) { // we put some tolerance since the number of generated events is small
+        std::cerr << "Number of generated MB events different than expected\n";
+        return 1;
+    }
```
fgrosa, line 103: I would remove this check, since `nEventsMB` is always `0`, so it is not really meaningful
```suggestion
//    if (nEventsMB < nEvents * (1 - ratioTrigger) * 0.95 || nEventsMB > nEvents * (1 - ratioTrigger) * 1.05) { // we put some tolerance since the number of generated events is small
//        std::cerr << "Number of generated MB events different than expected\n";
//        return 1;
//    }
```

fgrosa (changes requested):
Ciao @mfaggin, many thanks for the work! I just have a two minor comments (mostly for the test, not really the generator itself)

Then I would like to ask you to remove this other generator https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/PWGHF/external/generator/GeneratorHF.C, which is an outdated version of yours (to avoid confusion in the future). Thanks!

fgrosa (approved):

fgrosa:
label async-pp-2023-apass4 was removed by mistake, readding

PR #1667, opened 2024-06-17
Change from consul to apricot (tpc standalone + global sync)

rmunzer:
Change from consol to apricot usage for retrieving qc configuration.
Changes were tested in REPLAY runs in production with all detectors.


shahor02:
@ktf the O2fst regularly fails on O2DPG due to the missing alien token, while in O2 tests it seems to be quite stable.

PR #1672, opened 2024-06-21
Using the detector list from GRP for the wf and other minor adjustments

chiarazampolli:
@shahor02 : I tested imposing that the WORKFLOW_DETECTORS does not contain MCH, MID, MFT, ZDC (as we discussed, you had the doubt that this could create issues in AODs) and all works fine.

@davidrohr , here I use always "+=" for the `ARGS_EXTRA_PROCESS` and `CONFIG_EXTRA_PROCESS`. For the first, everything was behaving like a `+=` already, while the `CONFIG_EXTRA_PROCESS` needed a couple of fixes, @andreasmolander , @matthias-kleiner , FYI, FT0 and TPC are concerned.


chiarazampolli:
After a few fixes, since there were no objections, I am merging this one.

alcaliva:
Adding labels "async-2024-pp-apass1" and "async-2024-pp-cpass0" in order to apply the PR https://github.com/AliceO2Group/O2DPG/pull/1698

PR #1673, opened 2024-06-21
PWGHF: fix PDG code of L(1520) in Lc decays.

mfaggin:
@fcatalan92 @fgrosa @fcolamar @stefanopolitano @DelloStritto triggered by a couple of unexpected things in analysis, I started debugging a bit our MC productions and I found that apparently the PDG code of the Lambda(1520) is wrong in our PYTHIA configurations.

Apparently the correct PDG code is 102134 as listed in the manual https://pythia.org/latest-manual/Welcome.html `>> Particles and Decays >> Particle Data` , and not 3124 as written when adding and forcing the decay `Lc --> Lambda(1520)pi`. For some reasons, the correct PDG code was there already for the forcing of the `Lambda(1520) --> pK`.

Just as an example, I have tested it with the Pb-Pb generator `MC/config/PWGHF/external/generator/generator_pythia8_embed_hf.C` using the bash script here https://cernbox.cern.ch/s/BdlYlDpPXFRhZmy (10 generated events, w/o anchoring and w/o embedding), forcing all Lc hadrons to decay into the channel `Lc --> Lambda(1520)pi` (see thic config https://cernbox.cern.ch/s/WxM6pgRGcc2BSYT, which is https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/PWGHF/pythia8/generator/pythia8_charmhadronic_with_decays_Mode2_hardQCD_5TeV.cfg just commenting out all other decays) and then I checked the content with this tester https://cernbox.cern.ch/s/82E7SFp01j3nU0m (which is exactly this one `MC/config/PWGHF/ini/tests/GeneratorHF_D2H_ccbar_and_bbbar_PbPb.C` a bit modified just to monitor this specific aspect).
Below a screenshot of one of my tests:

photo_2024-06-21_13-56-08


mfaggin:
thanks @fcatalan92 , I've fixe it also in the additional file that you found.
Concerning your second point, I would agree with you. What about writing a mail when this PR will be merged?

fcatalan92 (approved):

fgrosa:
Hi @mfaggin @fcatalan92 thanks! Regarding `force_hadronic_D.cfg`, I think we should remove it because it is not used and might create confusion in the future. We can do it in another iteration though.

fcatalan92:
@AliceO2Group/reviewers-pwg-je, @AliceO2Group/reviewers-pwg-dq, @AliceO2Group/reviewers-pwg-em, @AliceO2Group/reviewers-pwg-lf
Please note this fix for the PDG code of the Lambda(1520), the same problem is present in other configuration files for your PWGs

fgrosa:
label async-pp-2023-apass4 was removed by mistake, readding

PR #1675, opened 2024-06-25
ALICE 3: proc_id is actually a long

ddobrigk:
...does not yet guarantee PYTHIA will accept it!

### DIFF
--- a/MC/config/ALICE3/pythia8/generator_pythia8_ALICE3.C
+++ b/MC/config/ALICE3/pythia8/generator_pythia8_ALICE3.C
@@ -18,20 +18,20 @@ public:
   GeneratorPythia8ALICE3() {
 
     char* alien_proc_id = getenv("ALIEN_PROC_ID");
-    uint32_t seedFull;
-    uint32_t seed = 0;
+    uint64_t seedFull;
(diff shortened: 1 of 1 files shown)


ddobrigk:
Hi again, sorry for the spam - maybe the vars set in [setVarsFromALIEN_PROC_ID.sh](https://github.com/AliceO2Group/O2DPG/blob/743647eb6946d0ea7ef8d76eb83bb0cc3c8388dc/DATA/production/common/setVarsFromALIEN_PROC_ID.sh) are the better ones for the seed set? Perhaps the int16 version, for instance, of the ALIEN_PROC_ID? ...

ddobrigk:
@jesgum I changed such that we only use bits 0-28 from the proc id. Can you please give this a try with your local setup? Many thanks! I'm reverting to draft until checked.

ddobrigk:
@jesgum has tested this, and everything works flawlessly - and we should have perfect handling of seeds, as we'll be giving seeds to PYTHIA that go from 0 to 536870911 (which is allowed by PYTHIA). @sawenzel @benedikt-voelkel can we please merge if that's alright with you? Thank you very much!

PR #1677, opened 2024-06-26
CTF Size QC task - Only for Global run

rmunzer:
To avoid conflict between parallel runs, the QC task on the EPNs is only started if CTP readout is enabled. 

In this case the task is only executed for one partion in parallel.

### DIFF
--- a/DATA/production/qc-workflow.sh
+++ b/DATA/production/qc-workflow.sh
@@ -272,7 +272,7 @@ elif [[ -z ${QC_JSON_FROM_OUTSIDE:-} ]]; then
   # CTF QC
   if [[ ! -z "${QC_JSON_CTF_SIZE:-}" ]]; then
     add_QC_JSON GLO_CTF ${QC_JSON_CTF_SIZE}
-    add_pipe_separated QC_DETECTOR_CONFIG_OVERRIDE '.qc.tasks.CTFSize.taskParameters.detectors=\"${WORKFLOW_DETECTORS}\"'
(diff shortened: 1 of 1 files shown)


DATA/production/qc-workflow.sh
```diff
@@ -124,7 +124,7 @@ elif [[ -z ${QC_JSON_FROM_OUTSIDE:-} ]]; then
     elif has_detectors_reco MCH MID && has_matching_qc MCHMID; then
         [[ -z "${QC_JSON_GLO_MCHMID:-}" ]] && QC_JSON_GLO_MCHMID=consul://o2/components/qc/ANY/any/glo-mchmid-mtch-qcmn-epn
     fi
-    if has_processing_step ENTROPY_ENCODER && [[ ! -z "$WORKFLOW_DETECTORS_CTF" ]] && [[ $WORKFLOW_DETECTORS_CTF != "NONE" ]]; then
+    if has_processing_step ENTROPY_ENCODER && [[ ! -z "$WORKFLOW_DETECTORS_CTF" ]] && [[ $WORKFLOW_DETECTORS_CTF != "NONE" ]] && has_detector CTP; then
       [[ -z "${QC_JSON_CTF_SIZE:-}" ]] && QC_JSON_CTF_SIZE=consul://o2/components/qc/ANY/any/glo-qc-data-size
```
shahor02, line 128: Good, the trending now is anyway triggered by the new CTP object filled. 
Could you also comment out the line https://github.com/AliceO2Group/O2DPG/blob/c3e723b03913b775de82af1e5937e70fa00154ab/DATA/production/qc-workflow.sh#L275, to suppress the messages on missing objects for detectors which are not in the readout?

rmunzer: Thanks @shahor02,

done

PR #1680, opened 2024-06-27
Skimming needs to know NHBPERTF, extract from ALIEN_JDL_LPMRUNNUMBER

shahor02:
@chiarazampolli  @mpuccio the TFs with no data in the skimmed CTFs corresponded to TFs for which nothing was selected but the following 3TFs had something selected. The reason is that we did not set from the command line the NHBFPERTF (via HBFUTILS), as a result, the workflow was taking default 128 HBF. So, the `ctf-reader` was seeing something selected for the `[1st_orbit : 1st_orbit+128]` but detector encoders did not find any selected BC overlapping with the TF being processed and were writing empty CTF. Effectively, the `--skip-skimmed-out-tf` was not working for most of TFs, but this should be fully transparent for further processing.
The corruption reports is something else, still debugging.

### DIFF
--- a/DATA/production/configurations/CTFSkimming/ctf-skim-workflow.sh
+++ b/DATA/production/configurations/CTFSkimming/ctf-skim-workflow.sh
@@ -3,6 +3,22 @@
 # Non-zero exit code already if one command in a pipe fails
 set -o pipefail
 
+if [[ -z "$NHBPERTF" ]]; then  # try to extract from JDL variables
+  NHBPERTF=32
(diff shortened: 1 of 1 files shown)


chiarazampolli:
Thanks @shahor02 , do we need some further review, or should I merge?
To be understood how 2023 worked, since it is also with 128 orbits per TF.

shahor02:
@chiarazampolli all 2023 has the same problem: many TFs rejected by the trigger are stored with 0 data, but this does not pose any problem (except ~1% increase of the new CTF size).

PR #1681, opened 2024-06-28
Script to run STARlight anchored productions

mbroz84:
For STARlight production would be very handy if we can pass arguments to the job script in grid_submit.sh, so I did this small adjustment.


GRID/utils/grid_submit.sh
```diff
@@ -575,7 +576,7 @@ export PATH=$PATH:$PWD
 # ----------- EXECUTE ACTUAL JOB  ------------------------------------ 
 # source the actual job script from the work dir
 chmod +x ./alien_jobscript.sh
-./alien_jobscript.sh
+./alien_jobscript.sh ${SCRIPTARGS}
```
sawenzel, line 579: This is unfortunately not as easy as this. One would, in addition, need to pass around the arguments in the JDL specification. Honestly, I would avoid doing this change for now. You can easily write a script which creates jobfiles containing the right arguments before submitting.

mbroz84: Ok, I reverted the changes in the grid_submit script.

mbroz84:

chiarazampolli:

PR #1683, opened 2024-07-01
Fix px/py bug in generator_pythia8_gun.C

ddobrigk:
This fixes a very old bug in which a random uniform number is chosen for py and then px is calculated accordingly, which is not a valid manner of randomly emitting particles uniformly in azimuth and will lead to a modulation versus azimuthal angle. Not sure if this is relevant for anyone at all but it's best fixed for posterity. 

Note that the relevant generation is still done in p (total momentum) and not pT (transverse momentum): this remains unchanged.


fcatalan92 (approved):
Hi @ddobrigk, thanks a lot for the fix. I agree that this generator is very old but could be useful for some simple studies.

PR #1686, opened 2024-07-02
Add --run to alignment geometry creation

sawenzel:
We need to put `--run` to DPL workflows which are prefixed with echo "echo ... pipe".

This should fix a problem where (under so far unknown circumstances), the geometry alignment workflow did not execute properly.


sawenzel:
+async-label async-2023-pbpb-apass3,async-2023-pbpb-apass4,async-2023-pp-apass4,async-2024-pp-apass1,async-2022-pp-apass7,async-2024-pp-cpass0

chiarazampolli:

PR #1688, opened 2024-07-03
ALICE 3: Add xicc injector for pp and PbPb

jesgum:
Adding generators that inject xicc for both pp and PbPb. There is also ini files that make use of these generators so I also added test scripts for respective ini file. However, I'm not sure if the tests are done properly yet so I'm currently marking as a draft for now.

### DIFF
--- a/MC/config/ALICE3/pythia8/generator/pythia8_hi.cmnd
+++ b/MC/config/ALICE3/pythia8/generator/pythia8_hi.cmnd
@@ -1,0 +1,21 @@
+### beams
+Beams:idA 1000822080		# Pb
+Beams:idB 1000822080		# Pb
+Beams:eCM 5520. 		# GeV
+
+### heavy-ion settings (valid for Pb-Pb 5520 only)
+HeavyIon:SigFitNGen = 0
+HeavyIon:SigFitDefPar = 13.88,1.84,0.22,0.0,0.0,0.0,0.0,0.0
+HeavyIon:bWidth = 14.48
+
+### processes (apparently not to be defined)
+
+### decays
+ParticleDecays:limitTau0 on	
+ParticleDecays:tau0Max 0.001
+
+! 2) Seed settings
+! Seed is set inside the generator
+! If run on the grid, seed is set to job id
+! If run locally, seed is set to 0
+Random:setSeed = on                           ! Random seed on
--- a/MC/config/ALICE3/ini/tests/xicc_PbPb.C
+++ b/MC/config/ALICE3/ini/tests/xicc_PbPb.C
@@ -1,0 +1,31 @@
+int External() 
+{
+  std::string path{"o2sim_Kine.root"};
+
+  TFile file(path.c_str(), "READ");
+  if (file.IsZombie()) {
+    std::cerr << "Cannot open ROOT file " << path << "\n";
+    return 1;
+  }
+
+  auto tree = (TTree *)file.Get("o2sim");
+  if (!tree) {
+    std::cerr << "Cannot find tree o2sim in file " << path << "\n";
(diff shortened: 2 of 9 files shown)


jesgum:
Hi @sawenzel @benedikt-voelkel ! It seems to be some issues with the tests for the .ini files. At at first the tests for both External passed but failed for Pythia, but now it looks all of them fails, and I am not sure as to why. Could you perhaps have a look and provide some advice? Many thanks!

sawenzel:
There is an errror saying `error: redefinition of 'Pythia8' as different kind of symbol
6-int Pythia8()`. I would try to avoid calling your function the same as an existing C++ class.

jesgum:
Is this concerning the files in the test directory? From the previous ini files that were added, the associated test files also had a Pythia8() function, which as I understand, is fine according to the readme in (https://github.com/AliceO2Group/O2DPG/blob/master/test/README.md) or am I missing something?

sawenzel:
I believe the README is misleading and wrong in this regard. All I can say is that there was a compile error because of duplication of symbols.

MC/config/ALICE3/pythia8/generator/pythia8_hi.cmnd
```diff
@@ -0,0 +1,21 @@
...
+
+### decays
+ParticleDecays:limitTau0 on	
+ParticleDecays:tau0Max 0.001
+
+! 2) Seed settings
+! Seed is set inside the generator
+! If run on the grid, seed is set to job id
+! If run locally, seed is set to 0
+Random:setSeed = on                           ! Random seed on
```
sawenzel, line 21: We may no longer need the seeding part. All o2::eventgen::GeneratorPythia8 will be seeded automatically from now on.

nzardosh:
> I believe the README is misleading and wrong in this regard. All I can say is that there was a compile error because of duplication of symbols.
> 
> I am less familiar with the testing framework but will try to take a look into this. You/ALICE3 seems to be the only place where tests with `int Pythia8` are present (although the tests don't actually do anything). Potentially this needs to be changed.
> 
> Update: I studied the code a bit. The testing seems to be ok with the `int Pythia8` function. No changes necessary. However, your test macros are not really testing anything which defies the purpose. The intention was to verify that the output of the generator is good (for instance non-empty ...). Will need to rethink if we should still force this upon the developers or make it optional.

Hi @sawenzel , we have this problem now with another PR that is stuck (https://github.com/AliceO2Group/O2DPG/pull/1703) due to the test failing. Is there a way to resolve this

ddobrigk:
> > I believe the README is misleading and wrong in this regard. All I can say is that there was a compile error because of duplication of symbols.
> > I am less familiar with the testing framework but will try to take a look into this. You/ALICE3 seems to be the only place where tests with `int Pythia8` are present (although the tests don't actually do anything). Potentially this needs to be changed.
> > Update: I studied the code a bit. The testing seems to be ok with the `int Pythia8` function. No changes necessary. However, your test macros are not really testing anything which defies the purpose. The intention was to verify that the output of the generator is good (for instance non-empty ...). Will need to rethink if we should still force this upon the developers or make it optional.
> 
> Hi @sawenzel , we have this problem now with another PR that is stuck (#1703) due to the test failing. Is there a way to resolve this

Hi @nzardosh , in the end Jesper and I solved the issue by deliberately typecasting to FairGenerator - this seems to continue working just fine and it solves the CI problem. I'd suggest you do the same :-D

ddobrigk:
@benedikt-voelkel @sawenzel can we please merge this? It's been on the system for a few weeks now and we really need this deployed for some ALICE 3 work. Thank you!

PR #1695, opened 2024-07-10
Pythia8 seeding improvements

sawenzel:
simplification of seeding code and harmonization thereof


MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
```diff
@@ -228,8 +228,7 @@ FairGenerator *GeneratorPythia8GapTriggeredCharm(int inputTriggerRatio, float yQ
   auto myGen = new GeneratorPythia8GapTriggeredHF(inputTriggerRatio, std::vector<int>{4}, hadronPdgList);
   auto seed = (gRandom->TRandom::GetSeed() % 900000000);
   myGen->setUsedSeed(seed);
-  myGen->readString("Random:setSeed on");
-  myGen->readString("Random:seed " + std::to_string(seed));
+  myGen->setInitialSeed(seed);
```
fcatalan92, line 231: Hi @sawenzel, if I understood correctly the implementation in https://github.com/AliceO2Group/AliceO2/pull/13281, this part is now redundant and can be removed, since it does exactly the steps performed when no specific seed is provided (taking it from ROOT)

sawenzel: yes, I was hesitating here because of the `myGen->setUsedSeed` in front.

sawenzel: Hi @fcatalan92 : Yes, I think we could simplify this further by taking away even the `auto seed = ; ... setInitialSeed()` steps. But I wasn't sure if this breaks something because of `setUsedSeed`.

fcatalan92: Indeed, after some recent developments for Pb-Pb HF productions, `setUsedSeed` is used to store the seed in our custom generator. This seed is later accessed by `generator_pythia8_embed_hf.C` to select a pT-hard bin. https://github.com/AliceO2Group/O2DPG/blob/7f94fb810db6d4958ea5b8eecdc5c5fbb396760b/MC/config/PWGHF/external/generator/generator_pythia8_embed_hf.C#L83

I wonder if a getter to the used seed could be added to `GeneratorPythia8`, this will allow us to simplify our custom generators. In this case, I think the seed should be updated when calling `GeneratorPythia8::seedGenerator()`.

At this point, all looks good to me for this PR. I can take care of updating the HF generators later on, if the development I proposed is done.

fcatalan92 (approved):

PR #1696, opened 2024-07-11
Set the analysisQC axis range depending on the coll system

chiarazampolli:
This should fix the issue seen in the analysis QC for the qa-event-track task for which the axis range is always the one of PbPb instead of changing in case of pp data. 
@mfaggin , @alcaliva


mfaggin:
thanks a lot Chiara, this is indeed what we want

chiarazampolli:
I took the risk and merged. Locally it seemed fine.

PR #1699, opened 2024-07-11
Adopt settings of pbpb23 apass4 test3 + --thin-tracks in AOD

shahor02:
@chiarazampolli Here I've added to the maser version of `setenv_extra.sh` the TPC errors and extra flags for shared treatment as they were in apass4_test3. Note that they are always applied, not only for pbpb.

Also enabled the AOD thinning (` --thin-tracks`) but only for pbpb23. Please confirm that this is what we need.


DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -671,6 +678,10 @@ if [[ $PERIOD == "LHC22c" ]] || [[ $PERIOD == "LHC22d" ]] || [[ $PERIOD == "LHC2
   export ARGS_EXTRA_PROCESS_o2_aod_producer_workflow+=" --ctpreadout-create 1"
 fi
 
+if [[ $ALIEN_JDL_LPMANCHORYEAR == "2023" ]] && [[ $BEAMTYPE == "PbPb" ]] ; then
```
chiarazampolli, line 681: Hello @shahor02 ,
Please remove this. Or even better change it to (so I don't open a PR myself):

```
if [[ $ALIEN_JDL_THISAODS == "1" ]] ; then
...
```
Thanks!

shahor02: OK, done.

chiarazampolli (approved):

chiarazampolli:
Thanks! For me it can be merged (@shahor02 , do you need to check further?)

PR #1711, opened 2024-07-23
MC/PWGEM: no forced b-decays for HF Cocktail

DanielSamitz:


DanielSamitz:
The failure of the test O2DPG/MC/config/PWGEM/ini/tests/Pythia8_Beauty_Cocktail.C is expected. The test was designed to check the forced semileptonic decays of the b-hadrons, but with this PR these decays are not forced anymore (only c-hadron are forced to decay to electrons)

sawenzel:
Ok. But shouldn't we then adjust the test as well?

DanielSamitz:
OK, then it would be the same test as O2DPG/MC/config/PWGEM/ini/tests/Pythia8_**Charm**_Cocktail.C (checking only the forced charm decays). I can do that.

sawenzel (approved):

PR #1714, opened 2024-07-30
ALICE3: Reduce the amount of injected xicc in pp

jesgum:


jesgum:
Hi @sawenzel and @benedikt-voelkel, with this pr I would just like to adjust the reduce injected xicc in the pp generator. There is no other significant changes so if possible could you go ahead and merge please? Thanks!

PR #1725, opened 2024-08-09
enable possibility to apply different set of cuts to Muon and Global Muon Tracks

lmassacr:
Change in the asynch QC config file for MFTMCHMID objects in order to apply different sets of cuts on MCH, MCH+MID tracks with respect to MCH+MFT or MCH+MID+MFT tracks. 
A new variable to cut on the Chi^2 between MFT and MCH was introduced. It is defined in the TrackTasks.cxx class in QualityControl/Modules/MUON/Common/src (PR 2382)

### DIFF
--- a/DATA/production/qc-async/mftmchmid-tracks.json
+++ b/DATA/production/qc-async/mftmchmid-tracks.json
@@ -14,9 +14,56 @@
           "query": "trackMCH:MCH/TRACKS;trackMCHROF:MCH/TRACKROFS;trackMCHTRACKCLUSTERS:MCH/TRACKCLUSTERS;mchtrackdigits:MCH/CLUSTERDIGITS;trackMFT:MFT/TRACKS;trackMFTROF:MFT/MFTTrackROF;trackMFTClIdx:MFT/TRACKCLSID;alpparMFT:MFT/ALPIDEPARAM/0?lifetime=condition&ccdb-path=MFT/Config/AlpideParam;fwdtracks:GLO/GLFWD;trackMID:MID/TRACKS;trackMIDROF:MID/TRACKROFS;trackMIDTRACKCLUSTERS:MID/TRACKCLUSTERS;trackClMIDROF:MID/TRCLUSROFS;matchMCHMID:GLO/MTC_MCHMID"
         },
         "taskParameters": {
-          "maxTracksPerTF": "600",
-          "GID" : "MFT,MCH,MID,MFT-MCH,MCH-MID,MFT-MCH-MID"
+                 "maxTracksPerTF": "600",
+                 "cutRAbsMin": "17.6",
+                 "cutRAbsMax": "89.5",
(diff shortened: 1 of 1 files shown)


DATA/production/qc-async/mftmchmid-tracks.json
```diff
@@ -14,9 +14,56 @@
...
+                 "cutRAbsMax": "89.5",
+                 "cutEtaMin": "-3.3",
+                 "cutEtaMax": "-2.5",
+                 "cutPtMin": "0.5",
+                 "nSigmaPDCA": "6",
+                 "cutChi2Max": "1000",
+                 "matchChi2MaxMFT": "45",
+                 "diMuonTimeCut": "100",
+                 "fullHistos": "0",
+                 "GID" : "MFT-MCH,MFT-MCH-MID"
```
aferrero2707, line 65: In order to have the possibility to evaluate the tracks matching fraction between MFT and MCH-MID, the GID parameter should be modified as
```
                 "GID" : "MFT-MCH,MCH-MID,MFT-MCH-MID"
```
This is to have MCH-MID tracks with the same kinematic cuts as the MFT-MCH-MID ones.

In case one needs to reduce the number of QC objects that are produced by the task, I would propose to eventually remove the `MFT-MCH` tracks from the selection:
```
                 "GID" : "MCH-MID,MFT-MCH-MID"
```

aferrero2707: @lmassacr actually I just realized that `MFT-MCH` in the GID list is useless without adding `MCH` as well, because you need the set of MCH-only plots with the same cuts as `MFT-MCH` for the comparisons.

If the number of plots becomes too big, you would need to restrict to the tracks with MID:
```
"GID" : "MCH-MID,MFT-MCH-MID"
```

lmassacr: Hi @aferrero2707,
For the moment I commited the config file adding the MCH standalone tracks with the same cuts as MCH-MFT tracks. 
I will see if there are concerns from DPG regarding the number of plots, I will modify according to your proposal if this is the case. Otherwise, we can go with the current configuration now.

chiarazampolli (approved):

chiarazampolli:
Hello @lmassacr , 

The change looks fine. How much will the output increase?

@JianLIUhep , @botta (I actually don't know her git account, Jian, can you add her?), FYI.

Chiara

chiarazampolli:
Hello @lmassacr , 

I guess this is 10% of the MFT output, right? I don't recall what is the MFT total size, maybe @JianLIUhep does, from the exercise to reduce the size? In any case, it seems ok to me. 
I will merge, while @JianLIUhep  takes the time to answer (he is away till Monday 23rd). If you see that this is overlooked, please, Laure, ping us.

Chiara

chiarazampolli (approved):

PR #1738, opened 2024-08-30
ITS calib: always parse trigger option

iravasen:
cc @shahor02 
to go with https://github.com/AliceO2Group/AliceO2/pull/13459/ (which is already merged)


iravasen:
@shahor02 can you please also merge this one when you have a second? Thanks!

shahor02 (approved):

PR #1745, opened 2024-09-13
PWGHF: add example to modify on the fly ini in run script

fgrosa:
@stefanopolitano this PR provides an example on how to modify the MC configurations in the `.ini` file to take them from any tag in CVMFS and avoid to port the configurations in all the tags, as discussed with @chiarazampolli.
To save the modified `.ini` file for bookkeeping, we can specify it as argument when running the `grid_submit.sh` (or at least I could not find a better solution).


MC/run/PWGHF/run_anchor_ini_from_cvmfs_example.sh
```diff
@@ -0,0 +1,42 @@
...
+CFGTOREPLACE="\${O2DPG_ROOT}/MC/config/PWGHF/pythia8/generator/pythia8_beautyhadronic_with_decays_Mode2.cfg" # original config file name to be modified
+CFGFROMCVMFS="/cvmfs/alice.cern.ch/el9-x86_64/Packages/O2DPG/daily-20240912-0200-1/MC/config/PWGHF/pythia8/generator/pythia8_beautyhadronic_with_decays_Mode2.cfg" # new config file name to use
+
+GENTOREPLACE="\${O2DPG_ROOT}/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C" # original external generator file name to be modified
+GENFROMCVMFS="/cvmfs/alice.cern.ch/el9-x86_64/Packages/O2DPG/daily-20240912-0200-1/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C" # new external generator file name to use
+
+if [ ! -f $MODIFIEDINI ]; then
+    sed -e "s|$CFGTOREPLACE|$CFGFROMCVMFS|g" -e "s|$GENTOREPLACE|$GENFROMCVMFS|g" $ORIGINALINI > $MODIFIEDINI
+fi
+
```
zhangbiao-phy, line 39:
```suggestion
MODIFIEDINI_ABSPATH=$(readlink -f $MODIFIEDINI)
```

MC/run/PWGHF/run_anchor_ini_from_cvmfs_example.sh
```diff
@@ -0,0 +1,42 @@
...
+export ALIEN_JDL_ANCHOR_SIM_OPTIONS="-gen external -ini $MODIFIEDINI"
```
zhangbiao-phy, line 40:
```suggestion
export ALIEN_JDL_ANCHOR_SIM_OPTIONS="-gen external -ini $MODIFIEDINI_ABSPATH"
```

MC/run/PWGHF/run_anchor_ini_from_cvmfs_example.sh
fgrosa: Should this be done also for the modified configuration path?

MC/run/PWGHF/run_anchor_ini_from_cvmfs_example.sh
```diff
@@ -0,0 +1,42 @@
...
+export SPLITID=100
+export PRODSPLIT=153
+export CYCLE=0
+export ALIEN_PROC_ID=2963436952
+
+# disable the QC
+export DISABLE_QC=1
+
+# modify ini file, to have external generator and/or config from a specific tag different from the one used for anchoring
+ORIGINALINI=${O2DPG_ROOT}/MC/config/PWGHF/ini/GeneratorHF_D2H_bbbar_Bforced_gap5_Mode2.ini # original .ini file to be modified
```
zhangbiao-phy, line 27: Additionally, some of the older tags don’t exist in this configuration. Perhaps we're using an old template?
```suggestion
ORIGINALINI=${O2DPG_ROOT}/MC/config/PWGHF/ini/GeneratorHF_D2H_bbbar_gap5.ini# original .ini file to be modified
```

MC/run/PWGHF/run_anchor_ini_from_cvmfs_example.sh
zhangbiao-phy: > @zhangbiao-phy should this be done also for the modified configuration path?

if you mean the CFG and GEN files, no, they are already the absolute path

MC/run/PWGHF/run_anchor_ini_from_cvmfs_example.sh
fgrosa: I would prefer to keep it like this (it is only an example). But of course it can be applied to any `.ini` file, in case the one we want to modify is not included (one just have to be careful that the other `ini` is as desired, except for the parts that are modified).

stefanopolitano:
Hi @fgrosa and @zhangbiao-phy! Thanks a lot for the PR and the work. I don't have comments on this. Approved.

stefanopolitano (approved):

PR #1748, opened 2024-09-16
LF: add multi-strange enriched generator

ddobrigk:


ddobrigk:
@njacazio @alcaliva for the cascade efforts in Pb-Pb, this should do the trick

njacazio:

PR #1756, opened 2024-09-27
Introduced new environment config variable

jackal1-66:
O2DPG_MC_CFG_ROOT is going to replace O2DPG_ROOT inside the config files. This will make the system more flexible by allowing the user to change this variable on-the-fly, and hence not be limited by the loaded version of O2DPG. At environment loading it will be by default equal to O2DPG_ROOT, as specified in PR [5649](https://github.com/alisw/alidist/pull/5649)

### DIFF
--- a/.github/workflows/syntax-checks.yml
+++ b/.github/workflows/syntax-checks.yml
@@ -71,6 +71,27 @@ jobs:
           jq -r '.comments[] | "Error found in \(.file) line \(.line):\n::error file=\(.file),line=\(.line),endLine=\(.endLine),col=\(.column),endColumn=\(.endColumn)::\(.message)"' errors.json
           exit "$error"
 
+  config-check:
+    name: Config-check
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Run configurations checker
+        run: |
+          error=0
+          readarray -d '' files < \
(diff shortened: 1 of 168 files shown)


jackal1-66:
O2DPG_MC_CFG_ROOT changed to O2DPG_MC_CONFIG_ROOT for clarity

sawenzel (approved):

jackal1-66:
build/O2DPG/sim/o2 will fail until https://github.com/alisw/alidist/pull/5649 is approved

sawenzel:
This should fix (or be the foundation for) https://its.cern.ch/jira/browse/O2-5366

@chiarazampolli @alcaliva : I think we could merge it if you are fine with it. (Not that backports after a merge would also need the corresponding commit in alidist/o2dpg https://github.com/alisw/alidist/pull/5649 ... maybe we should backport this in any case).

jackal1-66:
Issue seems related to gSystem->ExpandPathName not being able to expand ${O2DPG_MC_CONFIG_ROOT}. It could be that the variable is not set, even though the PR on alidist has been merged, since locally I get the same error message on ROOT when I unset manually the variable.

jackal1-66:
This merged PR https://github.com/AliceO2Group/O2DPG/pull/1757 manually sets the new configs variable to O2DPG_ROOT inside the test generators script. This should fix the failing CI

PR #1758, opened 2024-10-02
Creating the tar from root files only if there is a JDL env var that …

chiarazampolli:
…tells to do so

@catalinristea


chiarazampolli:
Hello @catalinristea , 

If this seems ok to you, I will merge it.

Chiara

catalinristea:
Ciao Chiara,
Yes, this was what I had in mind for it. Thanks a lot!
Yours, Catalin.

chiarazampolli:
Thanks @catalinristea , if we want it for older tags than the one that will be created end of November, we need to add the labels. What do you think?
Chiara

catalinristea:
Ciao Chiara! Yes, please flag it for all the active tags. It seems quite important feature for grid usage, but we can take it at the time of builds needed also for other things. Thanks!

Yours, Catalin.

chiarazampolli:
Done, thanks.

chiarazampolli:
This is not really needed for cpass0, but it was in the tag.

PR #1760, opened 2024-10-07
Basic implementation of event pooling in O2DPG workflow

jackal1-66:
This PR is related to the ticket https://its.cern.ch/jira/browse/O2-5216
It has been tested locally and the feature works as expected.
Check the ticket for more information. 
The approval of this PR is strictly connected to this other https://github.com/AliceO2Group/AliceO2/pull/13571


MC/bin/o2_dpg_workflow_runner.py
```diff
@@ -56,6 +56,7 @@
 parser.add_argument('--cpu-limit', help='Set CPU limit (core count)', default=8, type=float)
 parser.add_argument('--cgroup', help='Execute pipeline under a given cgroup (e.g., 8coregrid) emulating resource constraints. This m\
 ust exist and the tasks file must be writable to with the current user.')
+parser.add_argument('--kine-input', help='Use pre-existent event generation.', default="", type=str)
```
sawenzel, line 59: I would prefer not putting such treatment in the workflow runner because the runner shouldn't know about what it is running (just some topology).

Let's rather put such a flag in the actual MC workflow script generation part.
Or we provide a dedicated generator which will just do this and simply call this generator with existing flag `--gen`.

PR #1764, opened 2024-10-09
Mother particle transport off

mbroz84:
Bug fix for transport of mother particle.


mbroz84:

MC/config/PWGUD/external/generator/GeneratorStarlight.C
```diff
@@ -217,7 +217,7 @@ class GeneratorStarlight_class : public Generator
 				   0,0,0,0);
 	  //particle.Print();
 	  mParticles.push_back(particle);
-	  o2::mcutils::MCGenHelper::encodeParticleStatusAndTracking(mParticles.back(), 11);
+	  o2::mcutils::MCGenHelper::encodeParticleStatusAndTracking(mParticles.back(), 0);
```
sawenzel, line 220: For clarity it might actually be better to use "false/true" because the API specifies this as a boolean. (The value 11 from before is indeed misleading).

mbroz84: Ok, done.

sawenzel (approved):

PR #1765, opened 2024-10-09
DPMJET and other new processes in STARlight

mbroz84:
Main feature of this commit is to enable DPMJET process in STARlight.
STARlight can be compiled against DPMJET, in that case a separate library is created (libDpmJetLib) that uses routines from DPMJET to simulate photo-nuclear interactions. UPC group would like to use this mode. Few things are needed on the top of this PR:

Update DPMJET in the ALICE fork. Current version in the fork is 19.1.2, the photo-interactions in this version doesn't work (see: https://github.com/DPMJET/DPMJET/pull/20). We need latest version. I tried to do this exercise, there will be several conflicts, for my local test I always took "theirs" version and updated CMakeLists with new list of src files, then it compiles fine and works with STARlight.

Adjust the alidist: I attach here the git patch I used. The STARlight-DPMJET library is somehow not installed and remains in the build directory so I added explicit line to move it, maybe there is other solution...
[alidist.patch](https://github.com/user-attachments/files/17307166/alidist.patch)

Adjust STARlight: The additional library was build as static, I added option to build it as Shared. There is an issue with "FIND_LIBRARY" in the CMake module when the library path defined as a environment variable in the alidist/starlight.sh is preceded by ENV. Library is then not found when this is build via aliBuild. It definitely works in standalone installation. An expert opinion would be appreciated :) Again I attach here the git patch I used:
[STARlight.patch](https://github.com/user-attachments/files/17307169/STARlight.patch)

This PR then includes other developments (Tau processes) that are more minor and don't need any special treatment.


sawenzel:
A CI test is failing with:
```
[0;31m[FATAL]: O2DPG_TEST Script /O2DPG/MC/config/PWGUD/ini/tests/dpmjet_PbPb5360.C not defined for ini file /O2DPG/MC/config/PWGUD/ini/dpmjet_PbPb5360.ini[0m
```
which prevents merging. It seems that there is a file missing in the commit.

mbroz84:
That file is not a normal ini file for o2sim, but config for dpmjet, so I changed the suffix, so its not picked up for test

sawenzel:
> That file is not a normal ini file for o2sim, but config for dpmjet, so I changed the suffix, so its not picked up for test

Ok understood. Thanks.

sawenzel:
If you have patches for alidist, I would propose simply opening a PR. Then we can directly discuss on the code.

sawenzel:
Do you have a concrete script or workflow, which doesn't yet work, to evaluate success?

mbroz84:
I will prepare the script. What about the update of the DPMJET fork, how is it done?
Here is the PR for alidist: https://github.com/alisw/alidist/pull/5680

sawenzel:
> I will prepare the script. What about the update of the DPMJET fork, how is it done? 
Updates are done like for any other github repo: One has to open a PR with the proposed changes. Here, we probably need to bump master so that it contains the latest tag that you need. But I see that we have some ALICE specific commits for DPMJET, which might complicate matters a bit. Maybe @jackal1-66 or @pzhristov could take a look and help.

mbroz84:
Here is the run script:
[runSL_DpmjetSingle.txt](https://github.com/user-attachments/files/17634225/runSL_DpmjetSingle.txt)

mbroz84:
> > I will prepare the script. What about the update of the DPMJET fork, how is it done?
> 
> Updates are done like for any other github repo: One has to open a PR with the proposed changes. Here, we probably need to bump master so that it contains the latest tag that you need. But I see that we have some ALICE specific commits for DPMJET, which might complicate matters a bit. Maybe @jackal1-66 or @pzhristov could take a look and help.

On the top of the latest tag I don't need anything, so there is no PR from my side.

sawenzel:
Update of DMPJET is tracked here: https://its.cern.ch/jira/browse/O2-5508

PR #1767, opened 2024-10-10
Use new AggregatedRunInfo to fetch SOR/EOX/FirstOrbit

sawenzel:


sawenzel:
Should be merged only when the same transformations are applied in O2Physics.

sawenzel:
Needs https://github.com/AliceO2Group/AliceO2/pull/13584

PR #1769, opened 2024-10-13
avoid that PRODSPLIT is overwritten

pbuehler:


pbuehler:
This corrects a bug which affects simulations launched with the submit_grid.sh script. Currently PRODSPLIT is always set to 1 which causes troubles when SPLIT_ID > 1.

pbuehler:
Hi @sawenzel , @benedikt-voelkel , could you please have a look at this PR?

PR #1778, opened 2024-10-29
protection against corrupt O2DPG repository during topology generation

ehellbar:
I tested this locally both from scratch and with a corrupt repository from one of the topology generations today.

### DIFF
--- a/DATA/tools/epn/gen_topo_o2dpg.sh
+++ b/DATA/tools/epn/gen_topo_o2dpg.sh
@@ -82,15 +82,21 @@ while true; do
         break
       fi
     fi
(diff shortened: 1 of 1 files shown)


ehellbar:
@davidrohr I updated the commit as we discussed
1) reset to HEAD from the start
2) clean untracked files

davidrohr (approved):
Thanks, looks good!
Have you tested it on the corrupted repository that you have saved?

ehellbar:
Yes, it works as expected

davidrohr:
great, then let's get it merged, and let's ask RC to deploy with the next SW update, since we had 2 such failures already.

PR #1782, opened 2024-11-04
utility script to facilitate cherry-picking

sawenzel:
Script which helps to determine the complete list
of commits so that a given commit can be
cherry-picked without problems.


sawenzel:
@alcaliva @chiarazampolli : Could be of interest for the cherry-picking and tag creation

PR #1784, opened 2024-11-04
PWGLF: improve Xi/Om injected spectra

ddobrigk:
Necessary to fix https://its.cern.ch/jira/browse/O2-5402 (at 10% QA)


ddobrigk:
@njacazio

PR #1786, opened 2024-11-05
enable parallel world in anchored simulation

maciacco:
@mconcas @sawenzel I added the parallel world configurables in the python and bash scripts for the anchored workflow creation. I also added a variable in the pp test script to enable the parallel geometry navigation


MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -312,6 +312,7 @@ def main():
     parser.add_argument("-tf", type=int, help="number of timeframes per job", default=1)
     parser.add_argument("--ccdb-IRate", type=bool, help="whether to try fetching IRate from CCDB/CTP", default=True)
     parser.add_argument("--trig-eff", type=float, dest="trig_eff", help="Trigger eff needed for IR", default=-1.0)
+    parser.add_argument("-enable-parallel-world", type=int, dest="enable_parallel_world", help="Enable parallel geometry", default=0)
```
sawenzel, line 315: This should not be needed. We can already pass simulation config key values from the outside.

sawenzel (changes requested):
In principle, we can pass config key values from the outside using the `-confKey` of o2dpg_sim_workflow.py; 

Before doing bigger changes such as in this PR, could you please check if this works or what are the show-stoppers with such approach?

MC/bin/o2dpg_sim_workflow_anchored.py
maciacco: Thanks for your review, I removed this in the latest commit

maciacco:
@sawenzel thanks for your comment, I modified the implementation by forwarding the parallel world configuration to o2dpg_sim_workflow.py as you suggested. This gives the same results as before in the workflow creation

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -216,6 +216,7 @@ baseargs="-tf ${NTIMEFRAMES} --split-id ${SPLITID} --prod-split ${PRODSPLIT} --c
 remainingargs="-seed ${SEED} -ns ${NSIGEVENTS} --include-local-qc --pregenCollContext"
 remainingargs="${remainingargs} -e ${ALIEN_JDL_SIMENGINE} -j ${NWORKERS}"
 remainingargs="${remainingargs} -productionTag ${ALIEN_JDL_LPMPRODUCTIONTAG:-alibi_anchorTest_tmp}"
+remainingargs="${remainingargs} -confKey \"GeometryManagerParam.useParallelWorld=${ENABLE_PARALLEL_WORLD};GeometryManagerParam.usePwGeoBVH=${ENABLE_PARALLEL_WORLD};GeometryManagerParam.usePwCaching=${ENABLE_PARALLEL_WORLD}\""
```
sawenzel, line 219: this is not ideal. Someone might give `-confKey` from the outside ... and this will override the external choice of parameters.

alcaliva:
Hi all,
there was no progress over the last few weeks. What is the status of the review? We need this PR to reproduce the dips in the phi distribution in data. @sawenzel, could you have a look?

maciacco:
hi @alcaliva @sawenzel I got back on this in the past few days, I will update on the required checks in the coming days

maciacco:
@sawenzel @alcaliva I created a new PR on this as I made a mistake in the branch merging

PR #1798, opened 2024-11-12
PWGLF: fix seeding issues

ddobrigk:
This fixes a seeding issue in a few generators from LF: see plot from current simulations coming from one of them (the `extraStrangeness` one):

image

@sawenzel I think these all inherit from `GeneratorPythia8` so the seeding should have been taken care of automatically, should they not? I am not sure why this was not the case...


njacazio (approved):
Indeed

PR #1802, opened 2024-11-15
Script for hybrid JSON template generation

jackal1-66:
This PR is highly connected to https://github.com/AliceO2Group/AliceO2/pull/13699


jackal1-66:
I see there are multiple tests script that are failing, I will try to fix them directly in this commit (most of them are missing the interaction rate flag)

sawenzel:
> I see there are multiple tests script that are failing, I will try to fix them directly in this commit (most of them are missing the interaction rate flag)

I would prefer fixing unrelated issues in a separate PR.

MC/bin/o2_hybrid_gen.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
+
```
sawenzel, line 2: please add a short documentation what this tools does or achieves (including a short example usage)

jackal1-66:
I will open a new PR from a different branch. It's curious why this one caused every script in the /run/ folder to be retested, considering no pre-existing script was modified.

jackal1-66:
PR is now opened here https://github.com/AliceO2Group/O2DPG/pull/1808

PR #1812, opened 2024-11-21
GLOQC: Add K0s monitoring

f3sch:
Testing where CI fails.


chiarazampolli:
Hello @f3sch , 

I think it is ok like this. I would have not yet opened a PR here, I would have continued on our branch, so maybe we can move this to WIP for now?

Chiara

chiarazampolli:
Thanks, we can open here when we are sure everything is under control.

PR #1825, opened 2024-12-02
reduce ITS MV cycle

JianLIUhep:


chiarazampolli (approved):

JianLIUhep:
Hi @chiarazampolli, I added the label right after your approval. It seems you need to re-approve it. 
Thanks a lot.

chiarazampolli (approved):

chiarazampolli:
Error is unrelated, but it should be checked, @ktf , @singiamtel

Note that "@singiamtel" seems to not be recognized in this repository... @ktf , can you check with Sergio?

PR #1833, opened 2024-12-04
[PWGHF] Add configurations for non-prompt Omega MCs (gen only, no gap trigger)

fgrosa:
Hi @mpuccio @atriolo, this PR adds the configurations to produce a pool of generated events triggered on an OmegaC decaying to Omega + X. Could you please check that it's ok?

@creetz16 @pstahlhu @mfaggin @fcatalan92 @stefanopolitano we should do the same for OmegaC and Xic with forced decays (only `ini` files to be added, for the decay table we can use the usual config file).

### DIFF
--- a/MC/config/PWGHF/pythia8/generator/pythia8_charmhadronic_Omegac_to_Omega.cfg
+++ b/MC/config/PWGHF/pythia8/generator/pythia8_charmhadronic_Omegac_to_Omega.cfg
@@ -1,0 +1,45 @@
+### configuration for Omega from OmegaC
+
+### beams
+Beams:idA 2212			# proton
+Beams:idB 2212 			# proton
+Beams:eCM 13600. 		# GeV
+
+### processes
+SoftQCD:inelastic on		# all inelastic processes
+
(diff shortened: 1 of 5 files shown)


MC/config/PWGHF/pythia8/generator/pythia8_charmhadronic_Omegac_to_Omega.cfg
```diff
@@ -0,0 +1,45 @@
...
+5122:tau0 = 0.4390
+# Xic0
+4132:tau0 = 0.0455
+# OmegaC
+4332:tau0 = 0.0803
+
+### switch off OmegaC decays
+4332:onMode = off
+
+### Omega_c -> Omega pi
```
fcatalan92, line 44:
```suggestion
### Omega_c -> Omega + X
```
Hi @fgrosa, I have only this minor comment for clarity. 
BTW, not clear why the test is failing, maybe re-triggering it is enough.

sawenzel (approved):

PR #1834, opened 2024-12-04
Possibility to set the input specs for GLO QC dynamically, for K0s

chiarazampolli:
@shahor02 @f3sch

It was tested in the FST enabling the sec vtx:

```
export PRINT_WORKFLOW=1
export NHBPERTF=32
export SHMSIZE=8000000000
export NJOBS="$JOBS"
export DPL_REPORT_PROCESSING=1
export ARGS_EXTRA_PROCESS_o2_secondary_vertexing_workflow='--disable-cascade-finder --disable-3body-finder --disable-strangeness-tracker'

export WORKFLOW_EXTRA_PROCESSING_STEPS=TPC_DEDX,MFT_RECO,MID_RECO,MCH_RECO,MATCH_MFTMCH,MATCH_MCHMID,MUON_SYNC_RECO,ZDC_RECO,MATCH_SECVTX,MATCH_TPCTOF 
export FST_SYNC_EXTRA_WORKFLOW_PARAMETERS=QC,CALIB_LOCAL_AGGREGATOR,CALIB_LOCAL_INTEGRATED_AGGREGATOR 
export QC_REDIRECT_MERGER_TO_LOCALHOST=1 
export GEN_TOPO_WORKDIR=`pwd` 
export ALICE_O2SIM_DUMPLOG=1 
export NEvents=5 
export NEventsQED=100 
export O2SIMSEED=12345 
export QC_CONFIG_PARAM="--local-batch=QC.root"
$O2_ROOT/prodtests/full_system_test.sh
```

During the FST, there were issues with MFTdigits (when saving the QC output to file via the `--local-batch=QC.root` option) and also the TOF raw QC (which was then disabled).
For the FST, to enable sec vtx, I set:

`WORKFLOW_EXTRA_PROCESSING_STEPS=...MATCH_SECVTX,MATCH_TPCTOF `

where `MATCH_TPCTOF` can also be omitted. This same `WORKFLOW_EXTRA_PROCESSING_STEPS=...MATCH_SECVTX` would then be needed in online. 

Note that the reco_ASYNC in the FST also needs a change in the fst script, see https://github.com/AliceO2Group/AliceO2/pull/13770, since we need strangeness tracking in ASYNC for the AODs (disabled above by `export ARGS_EXTRA_PROCESS_o2_secondary_vertexing_workflow='--disable-cascade-finder --disable-3body-finder --disable-strangeness-tracker'`)

async reco was then tested on one run from zzo.

### DIFF
--- a/DATA/common/gen_topo_helper_functions.sh
+++ b/DATA/common/gen_topo_helper_functions.sh
@@ -49,6 +49,11 @@ has_detector_matching()
   [[ $WORKFLOW_DETECTORS_MATCHING =~ (^|,)"ALL"(,|$) ]] || [[ $WORKFLOW_DETECTORS_MATCHING =~ (^|,)"$1"(,|$) ]]
 }
 
+has_secvtx_source()
+{
+  [[ $SVERTEXING_SOURCES =~ (^|,)"ALL"(,|$) ]] || [[ $SVERTEXING_SOURCES =~ (^|,)"$1"(,|$) ]]
+}
+
 has_detector_qc()
 {
   has_detector $1 && [[ $WORKFLOW_DETECTORS_QC =~ (^|,)"$1"(,|$) ]]
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -712,7 +712,10 @@ fi
 if [[ $ALIEN_JDL_QCOFF != "1" ]]; then
   export WORKFLOW_PARAMETERS="QC,${WORKFLOW_PARAMETERS}"
 fi
-export QC_CONFIG_PARAM="--local-batch=QC.root --override-values \"qc.config.Activity.number=$RUNNUMBER;qc.config.Activity.passName=$PASS;qc.config.Activity.periodName=$PERIOD\""
+
+export QC_CONFIG_OVERRIDE+=";qc.config.Activity.number=$RUNNUMBER;qc.config.Activity.passName=$PASS;qc.config.Activity.periodName=$PERIOD;"
+
+export QC_CONFIG_PARAM+=" --local-batch=QC.root "
 export GEN_TOPO_WORKDIR="./"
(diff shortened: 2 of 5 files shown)


shahor02:
@chiarazampolli why not putting the 

```
export ARGS_EXTRA_PROCESS_o2_secondary_vertexing_workflow='--disable-cascade-finder --disable-3body-finder --disable-strangeness-tracker'
export WORKFLOW_EXTRA_PROCESSING_STEPS=TPC_DEDX,MFT_RECO,MID_RECO,MCH_RECO,MATCH_MFTMCH,MATCH_MCHMID,MUON_SYNC_RECO,ZDC_RECO,MATCH_SECVTX
```
to the `full_system_test.sh` (eventually, should go to dpl-workflow.sh). I would not add  `MATCH_TPCTOF ` since we don't want to run it online.

chiarazampolli:
Hello @shahor02 , 

I guess I have to, or it will not work (the CI fails). Another way is to switch off the K0s QC in the JSON for sync. 
Before that, I am actually trying to apply another modification: forcing the K0S QC to be off if there is no SECVTX. I need to test it. I will report here tomorrow.

For now, I am moving this PR to WIP.

Chiara

shahor02 (commented):
OK, assuming that the `jq` manipulations are correct, I am not familiar with this.
BTW, `EPNSYNCMODE` implies  `SYNCMODE`, it is enough to check the latter.

chiarazampolli:
Hello @shahor02 , 

Comparing lines 61 and 147 (where there are an "if"... "elif") it seems they are different. That is why we did like this. Should I remove ` SYNCMODE`?

Chiara

shahor02:
`SYNCMODE` does not imply `EPNSYNCMODE`, but  `EPNSYNCMODE` means `SYNCMODE` running on the EPN. 
That said, checking both with OR does not hurt.

chiarazampolli:
Checkes passed, merging.

PR #1836, opened 2024-12-05
[PWGHF] Add configurations for D2H XiC0, XiC+ and OmegaC  MCs (gen only, no gap trigger)

fcatalan92:
@fgrosa @fcolamar @mfaggin @creetz16 @pstahlhu @stefanopolitano, this PR adds the `ini` files for the production of generated MC pools for XiC0 + XiC+ and OmegaC0 (separately for charm and beauty)

P.s. I launched the tests for these configurations locally (but they are slow as we know)


fgrosa (approved):
Thanks @fcatalan92! Same issue with the tests here as in my https://github.com/AliceO2Group/O2DPG/pull/1833 (let's see if it is solved now after having retriggered the tests)

PR #1837, opened 2024-12-06
Possibility to thin only a percentage of the production

chiarazampolli:
@ddobrigk , @mfaggin , @catalinristea , @alcaliva 
@ddobrigk , @mfaggin , please take a careful look. I will merge upon your confirmation.


mfaggin:
thanks @chiarazampolli , the logic looks ok to me. Regarding metadata I've no experience, last word to @ddobrigk for it

chiarazampolli:
As discussed in the mail thread, we cannot use metadata in case AODs get merged (which is not the case at the moment in data reco, but one never knows). We'll need some further development. 
This means that for now the percentage option should *not* be used. I will add a protection, leaving the code there for when things will be mature.

Chiara

chiarazampolli:
The second commit avoids that we can use the option.

sawenzel:
Is this still relevant or should it be closed?

PR #1838, opened 2024-12-06
Add run type and beam type metadata for async and MC QC

knopers8:
QC-1188, QC-1002

Seeing no available run type in the surrounding scripts, I naively assume that absolutely all runs in async and MC are considered to be of type PHYSICS.


chiarazampolli:
Yes, we do not process calibration runs with this script. We might have cosmics, VdM etc, but these are also physics.

chiarazampolli (approved):

PR #1843, opened 2024-12-10
[PWGHF] Configuration for charm-baryon enriched MC anchored to pp ref run

klsmith15k:
Adding two new pythia8 MC files for pp ref run (only changes from original files are beam energy and pythia8 path)


fcatalan92 (approved):
Thanks @klsmith15k ! I just added a proper title for the PR

PR #1848, opened 2024-12-12
ITS::tracking: stop fatalising job, drop TF instead

mconcas:
Change the default behaviour, provided https://github.com/AliceO2Group/AliceO2/pull/13791 is included.
We stop breaking the processing if too much memory is required, we cleanup and move to the next TF.
@shahor02 @chiarazampolli @alcaliva

### DIFF
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -474,7 +474,7 @@ export ITSEXTRAERR="ITSCATrackerParam.sysErrY2[0]=$ERRIB;ITSCATrackerParam.sysEr
 # ad-hoc options for ITS reco workflow
 EXTRA_ITSRECO_CONFIG=
 if [[ $BEAMTYPE == "PbPb" ]]; then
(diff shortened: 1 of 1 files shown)


shahor02:
Thanks @mconcas.

For @chiarazampolli and @alcaliva : Matteo sees that the reconstruction with ITS per-ROF mode (instead of per-Vertex mode used in PbPb23 causing some inefficiencies) may crash due to the high memory if there are 2 central collisions in the same ROF.
This is a protection to skip the TF if the combinatorics grows too high.
We should use this option if we go to per-ROF ITS processing.

chiarazampolli:
Hello @shahor02 , 
If this is only when Per-PV is not enabled, should we have an `if`?
Chiara

mconcas:
> Hello @shahor02 , If this is only when Per-PV is not enabled, should we have an `if`? Chiara

This is regardless the per-PV, we see that there are anyways crashes even with the perPV, much less but still.
So I would treat thise as a default option for PbPb in general. Not to be backported, as it affects only the % of success and not directly the physics.

chiarazampolli (approved):

chiarazampolli:
Thanks, merged.

PR #1850, opened 2024-12-13
dynamically manipulating always the local json file in GLO QC, for K0s

ehellbar:
and minor formatting changes (spaces instead of tabs)

this is required to make #1834 work also in online
- downloading the json from apricot to local and adding it to the `JSON_FILES`
  - when using local jsons to begin with, they are also added to `JSON_FILES` in the `add_QC_JSON` function
- manipulating the local json taken from `JSON_FILES` and store it in a temporary file
- replace the initial local json with the manipulated temporary file in `JSON_FILES`

**please do not merge yet, should be tested in online first**


chiarazampolli:
Hi @ehellbar , 

I assume the changes are needed since something in online did not work. I tested both FST and async reco, and there all was fine.

Chiara

shahor02:
In the online topology creation, the consul path was treated as a local file.

shahor02 (approved):

chiarazampolli:
Ok, thanks!

ehellbar:
tested in online on STG, works.

PR #1853, opened 2024-12-16
its-threshold-processing.sh: allow-empy-rofs for all calib scan

iravasen:


iravasen:
Hi @shahor02, a trivial change for ITS2 calibration scans.

shahor02 (approved):

PR #1854, opened 2024-12-17
Ability to treat holes in run + Smaller fixes

sawenzel:


sawenzel:
@mbroz84 @chiarazampolli @shahor02

chiarazampolli:
Hi, 

Thanks! Do I understand that only exclusion is foreseen (no white list)? 

Cheers,

Chiara

MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -295,12 +296,52 @@ def determine_timestamp(sor, eor, splitinfo, cycle, ntf, HBF_per_timeframe = 256
...
+    orbit is some orbit after the orbitreset of the run
+    """
+    if len(filename) == 0:
+       return False
+
+    if not os.path.isfile(filename):
+       return False
+
+    # read txt file into a pandas dataframe ---> if this fails catch exception and return
+    df = pd.read_csv(filename, header=None, names=["Run", "From", "To", "Message"])
```
shahor02, line 321: Note that the CTFReader accepts delimiters like `,`, `;`, space and tab.

sawenzel:
> Hi,
> 
> Thanks! Do I understand that only exclusion is foreseen (no white list)?
> 
> Cheers,
> 
> Chiara

Don't know. So far, we only talked about exclusion. Can add the other thing later.

MC/bin/o2dpg_sim_workflow_anchored.py
sawenzel: I did this based on the example file that was given to me. Wouldn't it be easier to have a simple rule for this file? 
And I am not sure that space as a delimiter is useful because the example file contained words like `FC unstable` which would probably be split incorrectly.

Will make the code more generic tomorrow. For now it works and I need to have it in the nightly tag for testing tomorrow/

PR #1859, opened 2024-12-19
Fix injection gap math

mpuccio:
The change introduced in #1431 had the side effect that no signal events were present in recent productions using this generator, as for operator precedence one could never satisfy the condition of the if regulating the generation of signal events.

@njacazio @ddobrigk


ddobrigk:
Thanks, nice catch!

skundu692 (approved):

sawenzel:
You may also take a look at the novel hybrid event generator (see for instance https://github.com/AliceO2Group/AliceO2/blob/dev/run/SimExamples/Hybrid/hybridconfig.json) which allows to pick events from multiple different generator instances in certain sequences.

PR #1861, opened 2024-12-24
add option to set EMC busy time to 0

jaimenorman:
this option is useful for simulation setups where we want the EMCal to be read out for every event, in order to increase the number of events that the EMCal is available, for statistics reasons


sawenzel:
@jaimenorman : Thanks for your PR. Could you check if the alternative proposal #1869 would work for you as well. This solution is slightly more generic and would work for other users as well (without the need to add detector specific command line options such as `--emc-remove-busy`).

sawenzel:
Closing this (in favour of #1869). Discussed with @jaimenorman in private.

jaimenorman:
Thanks @sawenzel, I tested that your PR https://github.com/AliceO2Group/O2DPG/pull/1869 works well for this goal, and can be merged

PR #1863, opened 2025-01-07
Added evtpool to hybrid template generator

jackal1-66:


MC/bin/o2_hybrid_gen.py
```diff
@@ -63,6 +63,18 @@ def main():
...
+                    'name': 'evtpool',
+                    'config': {
+                        "eventPoolPath": "/path/to/filename.root",
+                        "skipNonTrackable": True,
+                        "roundRobin": False,
+                        "randomize": True,
+                        "rngseed": 0,
+                        "randomphi": False,
+                    }
+                })
```
sawenzel, line 77: I guess this is fine for now. In future, we could deduce these fields dynamically using the ROOT dictionaries. Python can surely simply iterate over the class fields.

PR #1864, opened 2025-01-07
Other Jpsi decays and trigger for various particles in dpmjet

mbroz84:


mbroz84:
Dear @sawenzel, can you have a look and merge? Would be good if I can add the new processes defined here to the 10% run of the UPC production.

PR #1871, opened 2025-01-09
PWGHF: rename config of OmegaC->Omega in OmegaC NoDecay

fgrosa:
@mpuccio and I realised that 
```
4332:onIfAny = 3334
```
does not work, since in PYTHIA the OmegaC is always decayed to diquarks/quarks that then hadronise later on, so we cannot select the decays with a specific hadron in the final state. To avoid misunderstanding I rename the relative config, which produces events with OmegaC that do not decay. 

This can be used to produce pregenerated pools of events with undecayed OmegaC baryons, that can be decayed later on with GEANT using an external generator (the same pool can be used for several productions requiring different forced decays of OmegaC), provided this fix in `GENAT4_VMC` https://github.com/vmc-project/geant4_vmc/pull/70.


stefanopolitano (approved):
Hi @fgrosa! Thanks for spotting this. Approved

PR #1872, opened 2025-01-11
Adding option to save multiple MFT-MUON matching candidates

mcoquet642:
This PR allows to save multiple MFT-MUON matching candidates, which will be used to build machine learning training samples.


MC/bin/o2dpg_sim_workflow.py
```diff
@@ -151,6 +151,7 @@
 
 # Matching training for machine learning
 parser.add_argument('--fwdmatching-save-trainingdata', action='store_true', help='enables saving parameters at plane for matching training with machine learning')
+parser.add_argument('--fwdmatchingNcandidates', type=int, help='enables saving provided number of matching candidates', default=0)
```
sawenzel, line 154: I believe this change is no longer necessary with https://github.com/AliceO2Group/O2DPG/pull/1869

You can now simply pass configurables from an external file. Could you please check if this works for you?

I really would favour to avoid adding more and more detector specific options to o2dpg_sim_workflow (when they merely set a config-key).

PR #1876, opened 2025-01-16
Included trigger in template generator

jackal1-66:


jackal1-66:
PR linked to https://github.com/AliceO2Group/AliceO2/pull/13868

MC/bin/o2_hybrid_gen.py
```diff
@@ -69,6 +69,19 @@ def main():
     else:
         print(f"Running in {mode} mode")
 
+    # Available options for trigger are "off", "or", "and"
+    # in all the other cases the trigger is forced "off"
+
+    trgbase = {
```
sawenzel, line 75: we could print this only when the user asks for it. Otherwise it might bloat the JSON quickly.

jackal1-66: I will add a --trigger flag.

PR #1879, opened 2025-01-20
Add Omega_c decay table and G4 config to use external decayer

mpuccio:
Co-authored with Fabrizio Grosa <fabrizio.grosa@cern.ch>


fgrosa (approved):
Thanks!

PR #1883, opened 2025-01-23
Change bad-data-interval parsing to account for more requirements

sawenzel:
Apparently the separator can change each line and be part of the comment message. Hence, simple csv parsing does not work.


sawenzel:
+async-label async-2024-PbPb-apass1

PR #1885, opened 2025-01-24
External Pythia8+Powheg generator

jackal1-66:
The generator has been tested using both o2-sim and the normal workflow system. 
Nothing to report with normal o2-sim, however for the workflow system I noticed that it's not possible 
to configure properly the parameters in the funcName => There might be a small bug that doesn't allow \" to be used or that discards it. With the workflow framework both the configuration via iniFile and hybrid JSON worked flawlessly.

### DIFF
--- a/MC/config/PWGGAJE/external/generator/generator_pythia8_powheg.C
+++ b/MC/config/PWGGAJE/external/generator/generator_pythia8_powheg.C
@@ -1,0 +1,105 @@
+R__ADD_INCLUDE_PATH($O2DPG_MC_CONFIG_ROOT)
+///#include "FairGenerator.h"
+//#include "Generators/GeneratorPythia8.h"
+#include "Pythia8/Pythia.h"
+#include "Generators/GeneratorPythia8Param.h"
+#include "CommonUtils/FileSystemUtils.h"
+// Pythia8 generator with POWHEG
+//
+// Author: Marco Giacalone (marco.giacalone@cern.ch)
+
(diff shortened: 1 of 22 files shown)


vkucera:
Thanks a lot @jackal1-66 !

MC/config/PWGGAJE/pythia8/generator/pythia8_powheg.cfg
```diff
@@ -0,0 +1,79 @@
+Beams:eCM 13000. 		# GeV
```
vkucera, line 1:
```suggestion
Beams:eCM 13600. 		# GeV
```

jackal1-66:
The CI failing is fixed by https://github.com/alisw/alidist/pull/5745

MC/config/PWGGAJE/external/generator/generator_pythia8_powheg.C
```diff
@@ -0,0 +1,102 @@
...
+// o2-sim -g external --noGeant -n 2 -j 8 --configFile $O2DPG_MC_CONFIG_ROOT/MC/config/PWGGAJE/ini/GeneratorPythia8POWHEG.ini
+
+namespace o2
+{
+namespace eventgen
+{
+
+using namespace Pythia8;
+
+class GeneratorJEPythia8POWHEG : public o2::eventgen::GeneratorPythia8
```
sawenzel, line 24: I think it would be useful to make a short description/comment of what this generator class does.

jackal1-66:
All CI tests passing, can be merged

PR #1894, opened 2025-02-04
MID: check the filtered digits in the async QC

dstocco:
This PR allows for a better comparison between data and MC.
The masks are applied during reconstruction for both data and MC. But data are already masked since this is done at the electronics level, while MC is masked only during reconstruction.
So far the QC checked for digits **before** the masks are applied, which leads to a difference between data and MC.
This PR introduces a check on digits **after** the masks are applied, so that we can compare apples to apples.

### DIFF
--- a/DATA/production/qc-async/mid.json
+++ b/DATA/production/qc-async/mid.json
@@ -3,10 +3,9 @@
     "config": {
       "database": {
         "implementation": "CCDB",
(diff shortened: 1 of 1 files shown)


lmassacr:
Hi all, 
It seems the failure of the CI is unrelated to the code, but to a token issue. Can this PR be approved and merged?

PR #1908, opened 2025-02-13
set pw in test script + toggle ideal ITS in anchored simulation

maciacco:
@sawenzel @mconcas @alcaliva 

I opened a new PR on the addition of the parallel world feature to the steering scripts.

As suggested in the previous thread by Sandro, I removed the overriding `confKey` setting in the `anchorMC.sh` script. As an example, the activation of the parallel world feature is shown in the pp test script.

I still include the changes in `anchorMC.sh` that toggle the usage of the ideal geometry for ITS in anchored simulations. First, we avoid fetching the ideal alignment object for ITS from the ccdb in case we require the parallel world. Additionally, if the parallel world is requested, the ITS geometry snapshot is not added to the local ccdb cache, while this is always done for MFT.

### DIFF
--- a/MC/run/ANCHOR/anchorMC.sh
+++ b/MC/run/ANCHOR/anchorMC.sh
@@ -300,7 +300,7 @@ fi
 
 # -- Create aligned geometry using ITS ideal alignment to avoid overlaps in geant
 ENABLEPW = 0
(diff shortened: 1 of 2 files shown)


mconcas:
Thanks, @maciacco , approving to start CI, in the menawhile.

jackal1-66 (approved):

PR #1910, opened 2025-02-18
MUON: added option for storing nCandidates in the MFT-MCH matching

aferrero2707:
The number of candidates to be stored is specified in the `MUON_MATCHING_NCANDIDATES` environment variable.

By default the variable is not set, and no option is passed to the global forward matching workflow for storing multiple candidates.

Hence the default behavior of the forward matching is unchanged.

The new option also requires the changes in https://github.com/AliceO2Group/AliceO2/pull/13977.


aferrero2707:
@mcoquet642 could you please have a look? Thanks!

mcoquet642:
> @mcoquet642 could you please have a look? Thanks!

Thanks Andrea, this looks fine to me. Indeed it would be nice to have this merged for the upcoming recos.

PR #1914, opened 2025-02-21
MID calibration: parse qc file with apricot

dstocco:
This is needed to launch a calibration run at P2.


dstocco:
@chiarazampolli @shahor02 @davidrohr : can  this PR be merged if it is ok with you ?
Thanks!
Cheers,
Diego

PR #1919, opened 2025-02-25
PWGHF: muon production with gap3 and correction of typo in charm macros with gap5

NicoleBastid:
1- new ini files and macros for charm and beauty  production of muons  with gap3

2- correction of typos in macros for charm production with gap5


fgrosa (approved):
I am approving this PR, but since these are configs I am removing the labels since we can always use the `O2DPG_MC_CONFIG_ROOT` environmental variable to load them, independent of the tag used for the anchoring (so no need to port them).

PR #1925, opened 2025-02-28
MUON: add default values for number of MFT-MCH matching candidates

aferrero2707:
The default number of matching candidates to be saved in output AO2Ds is set to 5 for pp and 20 for Pb-Pb. For other beam types it is set to zero, which prevents from adding the `FwdMatching.saveMode` and `FwdMatching.nCandidates` options to the forward matching workflow.

Setting the `MUON_MATCHING_NCANDIDATES` environment variable overrides the default values.

### DIFF
--- a/DATA/common/setenv.sh
+++ b/DATA/common/setenv.sh
@@ -169,9 +169,6 @@ ITSMFT_STROBES=""
 [[ ! -z ${ITS_STROBE:-} ]] && ITSMFT_STROBES+="ITSAlpideParam.roFrameLengthInBC=$ITS_STROBE;"
 [[ ! -z ${MFT_STROBE:-} ]] && ITSMFT_STROBES+="MFTAlpideParam.roFrameLengthInBC=$MFT_STROBE;"
 
-MFTMCH_NCANDIDATES_OPT=
-[[ ! -z ${MUON_MATCHING_NCANDIDATES:-} ]] && MFTMCH_NCANDIDATES_OPT+="FwdMatching.saveMode=3;FwdMatching.nCandidates=${MUON_MATCHING_NCANDIDATES};"
-
 
 # Set active reconstruction steps (defaults added according to SYNCMODE)
 for i in `echo $LIST_OF_GLORECO | sed "s/,/ /g"`; do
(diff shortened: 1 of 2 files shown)


aferrero2707:
+async-label async-2024-PbPb-apass2, async-2023-PbPb-apass5

aferrero2707:
@mcoquet642 could you please have a look? Thanks a lot!

mcoquet642:
Thanks Andrea, this looks fine for me

mcoquet642:
Dear @sawenzel , @chiarazampolli, @shahor02 could you please see if this could be merged, so that it can be included in the software freeze ? Thanks

alcaliva (approved):

alcaliva:
@sawenzel, can you check and if all is fine merge this PR? We need this to produce the new tag for PbPb apass

sawenzel:
@davidrohr

alcaliva:
Dear @davidrohr , could you please check this PR? We need this to produce the new tag for PbPb

davidrohr:
sorry, didn't see this. Is this needed only for async? Then I would prefer to have it in setenv_extra.sh, not in the general setenv.sh.

aferrero2707:
@davidrohr ok, I will move the initializations in setenv_extra.sh and update the PR.

alcaliva:
Hi @aferrero2707, did you move the initializations to setenv_extra.sh?

DATA/common/setenv.sh
```diff
@@ -169,9 +169,6 @@ ITSMFT_STROBES=""
 [[ ! -z ${ITS_STROBE:-} ]] && ITSMFT_STROBES+="ITSAlpideParam.roFrameLengthInBC=$ITS_STROBE;"
 [[ ! -z ${MFT_STROBE:-} ]] && ITSMFT_STROBES+="MFTAlpideParam.roFrameLengthInBC=$MFT_STROBE;"
 
-MFTMCH_NCANDIDATES_OPT=
```
davidrohr, line 172: @aferrero2707 : Sorry, but this line must stay in setenv.sh, since this one is also used in online :(.

davidrohr (changes requested):

davidrohr:
I'll move it back myself to get this going.

DATA/common/setenv.sh
aferrero2707: @davidrohr why you say this is needed for online as well? AFAIK we are not using the option for multiple candidates in online.

In the previous PR I have put the lines in `setenv.sh` because I was not aware of the `setenv_extra.sh` specific for async processing...

davidrohr: Ah, ok, I didn't know that.
I thought also MFTMCH_NCANDIDATES_OPT must be set for sync, since it is used.
But it is actually used like `[[ ! -z ${MFTMCH_NCANDIDATES_OPT:-} ]] &&` , so ok if unset.
I'll revert my moving of the fail. Thx for checking and sorry for the confusion.

PR #1943, opened 2025-03-18
Fix path to file in consul

dstocco:
With the PR https://github.com/AliceO2Group/O2DPG/pull/1914 , the MID calibration run was changed in order to use QC files from apricot. However, this led to a bug since the path to consul had to be changed as well.
This PR should solve the issue, allowing to launch calibration runs at P2.

### DIFF
--- a/DATA/production/calib/mid-badchannels.sh
+++ b/DATA/production/calib/mid-badchannels.sh
@@ -44,7 +44,7 @@ workflow_has_parameter CTF && {
 add_W o2-mid-calibration-workflow "" "" 0
 add_W o2-calibration-ccdb-populator-workflow "--ccdb-path=\"$CCDB_POPULATOR_UPLOAD_PATH\" --sspec-min 0 --sspec-max 0"
 add_W o2-calibration-ccdb-populator-workflow "--ccdb-path=\"$CCDB_POPULATOR_UPLOAD_PATH_DCS\" --sspec-min 1 --sspec-max 1 --name-extention dcs"
-workflow_has_parameter QC && add_QC_from_apricot "/o2/components/qc/ANY/any/mid-calib-qcmn" "--local --host localhost"
+workflow_has_parameter QC && add_QC_from_apricot "components/qc/ANY/any/mid-calib-qcmn" "--local --host localhost"
 WORKFLOW+="o2-dpl-run $ARGS_ALL $GLOBALDPLOPT"
 
 if [ "$WORKFLOWMODE" == "print" ]; then


dstocco:
@shahor02 @chiarazampolli @davidrohr : can this be merged? The MID calibration is currently broken at P2 (sorry, my fault), so it would be good if this PR lands in P2 before the next flp-suite

shahor02:
I am not familiar with the consul path conventions, in the O2DPG I see as many scripts setting `QC_CONFIG="/o2/components/qc/ANY....` as `QC_CONFIG="components/qc/ANY/...`.

@davidrohr @chiarazampolli, do you know the rule ?

chiarazampolli:
I don't know but I assume that @dstocco  is right, looking at the TPC scripts at least I see none with the `/o2/` that Diego wants to remove. I also assume that he knows that this is the right way. I would approve.

dstocco:
Well, on my side it is just try and fail. I've launched the calibration run by pointing to my local O2DPG in `alihlt-gw-prod.cern.ch`. If I use "/o2/" it fails. But if I launch it with the current PR, it works.
Unfortunately I did not find any documentation on this point...
In my previous PR I forgot that I could test it this way and simply assumed that I just needed to change the helper function.

chiarazampolli (approved):

chiarazampolli:
Since Diego tested it, i merged it.

chiarazampolli:
@knopers8 , @Barthelemy , can you confirm that like this is correct and point us to a documentation, if it exists, for the future?

knopers8:
If the config file is retrieved from apricot by the QC executable, please refer to: https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#templating-config-files

The Apricot REST API, which appears to be directly used by O2DPG scripts, is documented at:
- https://github.com/AliceO2Group/Control/blob/master/apricot/docs/apricot_http_service.md
- `<aliecs_hostname>:32188/docs/index.html`

In particular, notice that when apricot is used directly, `o2` has to be removed. It should be present in URLs passed to QC. I agree it's confusing.

chiarazampolli:
Hi @knopers8 , 

Also to me it seems that the function uses the REST API (https://github.com/AliceO2Group/O2DPG/blob/691d4d5ba2616a4dc54b667fb9194503cfe9533a/DATA/common/gen_topo_helper_functions.sh#L280) 

From this page for the REST API (which I guess is what you refer to when you say "when apricot is used directly", right?):

https://github.com/AliceO2Group/Control/blob/master/apricot/docs/apricot_http_service.md

I'd deduce that the "o2" in front is not needed. 

@dstocco , please confirm that everything is fine now, as soon as you can test it.

Chiara

dstocco:
Thanks @knopers8 for the links. Indeed, in the previous PR I did not remove the `/o2/`, that is why it was failing.
@chiarazampolli : this time I managed in testing the modification at P2, and got no error. So, I'm pretty confident. Now the PR is merged, so, in principle, the fix should be included in the next FLP suite.

knopers8:
> From this page for the REST API (which I guess is what you refer to when you say "when apricot is used directly", right?):

Right.

PR #1952, opened 2025-03-26
Disable MCHMFT QC task

sawenzel:
until https://its.cern.ch/jira/browse/O2-5885 is fixed

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -1519,11 +1519,12 @@ for tf in range(1, NTIMEFRAMES + 1):
   
                 
      ### MCH && MFT
-     if isActive('MCH') and isActive('MFT') :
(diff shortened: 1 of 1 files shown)


sawenzel:
@lmassacr : I would disable the failing to have working nightlies. We can reactive when the issue is fixed. Would you agree?

lmassacr:
Hi @sawenzel , 

Sure you can disable the MCH-MFT aQC task in the meanwhile. Shouldn't you also comment the following lines in o2dpg_qc_finalization_workflow.py? :

if isActive('MCH') and isActive('MFT'):
     add_QC_finalization('MCHMFTTaskQC', 'json://${O2DPG_ROOT}/MC/config/QC/json/mftmch-tracks-task.json')
     
Cheers,
Laure

sawenzel:
Closing in favour of #1953

PR #1960, opened 2025-04-09
Remove sources and inputs in async json, and fix for sync

chiarazampolli:
@f3sch , as we discussed.
Fixed also the json in the qc_sync folder, even if I don't think it is used.


chiarazampolli:
Error unrelated, @ktf , Sergio (I don't know the guthub account, I will look for it), maybe you'd like to take a look.

Merging.

PR #1962, opened 2025-04-11
FIT: MC digitizer settings for 2024

andreasmolander:
Temporary gain settings for 2024 MC. Will be reverted once we can take this from CCDB.

### DIFF
--- a/MC/bin/o2dpg_sim_config.py
+++ b/MC/bin/o2dpg_sim_config.py
@@ -90,10 +90,21 @@ def create_sim_config(args):
     if args.col == "PbPb" or (args.embedding and args.colBkg == "PbPb"):
         add(config, {"ITSVertexerParam.lowMultBeamDistCut": "0."})
 
-    # FIT digitizer settings for 2023 PbPb
+    # FIT digitizer settings
+    # 2023 PbPb
     if 543437 <= int(args.run) and int(args.run) <= 545367:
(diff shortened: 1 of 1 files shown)


andreasmolander:
+async-label async-2024-pp-apass1, async-2024-pp-cpass0, async-2024-PbPb-apass1, async-2024-ppRef-apass1, async-2024-PbPb-apass2

sawenzel (approved):

sawenzel:
@andreasmolander : We probably do not need these back-portings for MCs using the 2-tag approach (https://aliceo2group.github.io/simulation/docs/o2dpgworkflow/anchored.html?highlight=anchorMC#the-2-software-tag-approach-for-anchored-mc-productions)

How urgent is this, and for whom? Can't we wait for this to come from CCDB?

andreasmolander:
Hi @sawenzel my guess is that this is not urgent at all to be honest. I would have to double check with the rest of the team. Actually I think we can adjust the labels so that you don't need to cherry pick or back port anything, it's fine if it's included naturally the next time you checkout from or rebase onto the master branch. So that means no labels?

Unfortunately I cannot give an estimate for when we can take these from CCDB, so I would still like to include this. Our JIRA ticket for getting these from CCDB is blocked by a ticket to revert these temporary solutions, so it should not be forgotten.

PR #1966, opened 2025-04-16
Refactor of anchor configuration logic

sawenzel:
This commit refactors the transfer of anchoring configuration from async-reco scripts into the MC pipeline, addressing O2-5011 by replacing the previous whitelisting mechanism with a blacklisting approach.

In addition, it introduces broader configuration/customization support: users can now inject arbitrary command-line options into any task via a `customize.json` file, reducing the need to edit workflow scripts.

For example:
{
  "ConfigParams": {
    "EMCSimParam": {
      "mBusyTime": "11.11"
    }
  },
  "Executables": {
    "o2-ft0-reco-workflow": {
      "filtered": {
        "--my-custom-option": "foo"
      }
    },
    "ft0fv0emcctp_digi": {
      "filtered": {
        "--another-custom-option": "baz"
      }
    }
  }
}

This allows for flexible injection of both configKeyValues and CLI options via `o2dpg_sim_workflow.py --overwrite-config customize.json`.


sawenzel:
@shahor02 : Proposal for the blacklisting refactor. We will now pickup things like `--trackqc-fraction` for the aod-converter from the async workflows and in addition we gain a much better configurability overall.

We still need to do some final adjustments, proper testing, and the treatment for TPC distortions. For now WIP.

shahor02:
@sawenzel thanks, looks nice! (did not dive into the implementation details)

MC/run/ANCHOR/anchorMC.sh
```diff
@@ -126,7 +126,6 @@ echo_info "Substituting ALIEN_JDL_LPMPRODUCTIONTAG=$ALIEN_JDL_LPMPRODUCTIONTAG w
 ALIEN_JDL_LPMPRODUCTIONTAG=$ALIEN_JDL_LPMANCHORPRODUCTION
 
 if [[ $ALIEN_JDL_ANCHOR_SIM_OPTIONS == *"--tpc-distortion-type 2"* ]]; then
-  export O2DPG_ENABLE_TPC_DISTORTIONS=ON
```
shahor02, line 129: @sawenzel suppression of this line leads to disabling corrections in the reconstruction even if distortions were used in the digitization (this happened in the recent test of Mesut LHC25b8a_default_v4, while in LHC25b8a_default_v3 they were OK). 
I remember you did something else to detect distortion used in the digi. but I don't remember what and it seems it did not work. Shall we simply restore this line?

PR #1969, opened 2025-04-22
Add Angantyr O-O cfg

ddobrigk:


jackal1-66:
Ciao @ddobrigk, an example of OO configuration has been merged already in AliceO2 via this PR: https://github.com/AliceO2Group/AliceO2/pull/14190 . It's the most basic configuration one could think of at 10.72 TeV. I'm not against your configuration as well, considering it contains also additional parameters and an eCM of 6.8 TeV, but wanted to understand if it has a specific purpose.

sawenzel:
@ddobrigk : Thanks for your initiative. We also started to prepare these scenarios alongside JIRA ticket https://its.cern.ch/jira/browse/O2-5914. @jackal1-66  made some configs available directly in O2, so that they can also be used in full-system testing on the EPN. Ideally we should avoid duplication.

For what concerns O2DPG MC (not hyperloop, not O2), the config can actually be generated using the tool `mkpy8cfg.py` based on input to the workflow generator and as a function of energy etc.

ddobrigk:
Hi @sawenzel , of course, I agree we should not duplicate, but I have to admit I was somewhat confused that generator configs are now first in O2 and not in O2DPG, where - at least to my knowledge - this was standard. Am I to understand that the O2 version is the 'master' one and the rest is to be generated with the python script you point out and then put (with a manual PR) into O2DPG? 

As for the duplication: I will close this soon then if the reply to the question above is yes :-) 

Note, though, that the p-O is likely still not correct in @jackal1-66 's initialization, but I can comment this directly to the JIRA ticket then.

ddobrigk:
Hi @sawenzel - alright, thanks a lot! Then maybe, provided we can cover the O-O cases right away with the tool you propose, I will close this PR :-)  

Just one very minor comment - for O-O in ALICE 3, we used the nominal-top energy (rigidity equivalent to pp 14TeV, which is then 7 TeV for O-O since mass number and atomic number are 2:1... and for 2025 O-O this won't be the right energy, so better have a separate cfg file.

PR #1971, opened 2025-04-23
Disable outdated tpc_spacecharge_downloader task

sawenzel:
This task seems no longer needed.


sawenzel:
@shahor02 @wiechula : I believe this is no longer needed. No one seems to access objects `TPC/Calib/CorrectionMap`, `TPC/Calib/CorrectionMapRef` anymore. Can you confirm?

shahor02:
@sawenzel indeed, they are not used unless very old (<20230817) O2 is deployed, which should never be a case.

PR #1977, opened 2025-04-25
MUON: added dataSampling policies for muon tracks in async QC configuration

aferrero2707:
The GLO-MUONTracksMFT QC task has been reported to use 100% CPU in some async reconstructions, thus reducing the overall CPU efficiency of the production.

The input data for this task is now sampled to 10%. A similar sampling is also introduced, but not enabled by default, for the GLO-MUONTracks task.

See https://its.cern.ch/jira/browse/O2-5931 for further details.


aferrero2707:
+async-label async-2023-PbPb-apass5

PR #1979, opened 2025-04-25
Do not unset ARGS_EXTRA_PROCESS_o2_ctf_reader_workflow

shahor02:
As it prevents passing extra options, like the limit on the N TFs to process


chiarazampolli:
Mmm... Setting the max number of TFs always worked for me locally.

shahor02:
Setting `--max-tf <N>` will work because the `dpl-workflow.sh` has a dedicated env.var. for that. But more selective options, like `--max-tf-per-file N` or `--select-ctf-ids <CSL>` will not. There is simply no reason to unset the env.var.

We should also eventually remove any option for the TF injection delay, as not all workflows support throttling.

PR #1982, opened 2025-04-28
Include special runs in wf creator

jackal1-66:
Generated dictionary with all available collision systems and added special runs


alcaliva (approved):

sawenzel:
This needs prior merge of https://github.com/alisw/alidist/pull/5870.

PR #1991, opened 2025-05-07
Common: add OTF inis for special 2025 runs

ddobrigk:
Adds `ini` and `cfg` files such that one can run hyperloop on-the-fly simulations with the 2025 special run configurations.


jackal1-66:
Ciao @ddobrigk, could you please define some easy test scripts for the ini files? Something that checks the collision systems and the number of events in the root file should be enough from my pov.

ddobrigk:
Ciao @jackal1-66, certainly! Now done. Please let me know if there's anything further to adjust. Thanks a lot!

PR #1995, opened 2025-05-10
Add injected nuclei generator for HMPID; extend long-lived generator with phi range selection

alcaliva:
…with phi range selection


alcaliva:
@njacazio and @gvolpe79, I modified the existing generator making it more flexible by allowing the selection of eta and phi ranges (default values are -1<eta<1 and 0<phi<pi) to inject particles into HMPID acceptance. I also added the ini file which generates PbPb events and injects 10 (anti)deuterons using the modified generator_pythia8_longlived.C version

alcaliva:
Hi @sawenzel, 

I get this error which I do not understand:

/O2DPG/MC/config/PWGLF/pythia8/generator_pythia8_longlived.C:46:28: error: only virtual member functions can be marked 'override'
  Bool_t importParticles() override


isn't this supposed to override a virtual method in the base class, as it used to do? Why all of a sudden it doesn't work here? Any clue? Thanks

njacazio (approved):
Thanks a lot Alberto

PR #2005, opened 2025-05-20
parse script: adding optional env var to disable MI100 serialization workaround in dpl-workflow.sh

ehellbar:


ehellbar:
related PR in O2: https://github.com/AliceO2Group/AliceO2/pull/14302

PR #2006, opened 2025-05-20
Fetch models from CCDB or AliEn

jackal1-66:
ONNX models are now fetched by default on CCDB, but the option to pick them from AliEn has been introduced.
For now each TF will download its own files (1.2MB in total), but in the future maybe these could be put directly in the ccdb folder.


MC/config/common/external/generator/TPCLoopers.C
```diff
@@ -390,6 +397,49 @@ FairGenerator *
     gauss = gSystem->ExpandPathName(gauss.c_str());
     scaler_pair = gSystem->ExpandPathName(scaler_pair.c_str());
     scaler_compton = gSystem->ExpandPathName(scaler_compton.c_str());
+    const std::array<std::string, 2> models = {model_pairs, model_compton};
+    const std::array<std::string, 2> local_names = {"WGANpair.onnx", "WGANcompton.onnx"};
+    const std::array<bool, 2> isAlien = {models[0].starts_with("alien://"), models[1].starts_with("alien://")};
+    const std::array<bool, 2> isCCDB = {models[0].starts_with("ccdb:"), models[1].starts_with("ccdb:")};
+    if (std::any_of(isAlien.begin(), isAlien.end(), [](bool v) { return v; }))
+    {
+        TGrid::Connect("alien://");
```
sawenzel, line 406: Maybe it's not necessary to always connect to the GRID. This should only be needed in case of files on AliEn, right?

MC/config/common/external/generator/TPCLoopers.C
```diff
@@ -3,6 +3,8 @@
 #include <vector>
 #include <fstream>
 #include <rapidjson/document.h>
+#include "CCDB/BasicCCDBManager.h"
```
sawenzel, line 6: I don't see BasicCCDBManager used anywhere.

MC/config/common/external/generator/TPCLoopers.C
jackal1-66: This is already the case. The connection is opened only if at least an alien path is passed as an argument

shahor02: @jackal1-66 please always check for the existing gGrid before connecting, see https://github.com/AliceO2Group/AliceO2/pull/14322

jackal1-66: Added check

MC/config/common/external/generator/TPCLoopers.C
jackal1-66: Indeed, I used it before for other tests. Now replaced with CCDBTimeStampUtils.h

MC/config/common/external/generator/TPCLoopers.C
```diff
@@ -377,6 +379,11 @@ class GenTPCLoopers : public Generator
 } // namespace eventgen
 } // namespace o2
 
+// ONNX model files can be local, on AliEn or in the ALICE CCDB.
+// For local and alien files it is mandatory to provide the filenames, for the CCDB instead the
+// path to the object in the CCDB is sufficient. The model files will be downloaded locally.
+// Example of CCDB path: "ccdb:Users/n/name/test"
```
sawenzel, line 385: I have some doubts about the prefix `ccdb:`. Is this already used like this somewhere else? I see your intention but maybe we could stick to the more common practice of having protocols be specified with `ccdb://` (similar to `alien://` or `http://`). I am also wondering if there are actually user paths like `Users/n/name` on our CCDB (so far I have not found any)?

PR #2008, opened 2025-05-21
Edited POWHEG using parallel generation and manyseeds

jackal1-66:
- POWHEG generation is now split into chunks (default 1e4 events). This should be helpful especially on hyperloop (where the generator was first requested). The program will be configured only once, creating the needed files which later will make the simulation faster at each iteration. The .lhe filename is updated at every step inside Pythia.
- A fix to improve macros changes detection linked to ini files has been pushed


jackal1-66:
After the fix, the ini files seems to be grepped correctly from local and lxplus tests. Waiting for new CI results

PR #2010, opened 2025-05-26
Dummy change for testing purpose

jackal1-66:


jackal1-66:
PR will be closed, but it proves that there's an issue in the current macros testing mechanism. Many more ini files depend on the gaptrigger generator, compared to what appears in the CI check. This will be fixed by https://github.com/AliceO2Group/O2DPG/pull/2008

PR #2013, opened 2025-05-27
PWGHF: Exclude parton event in Ds reso triggered config

fgrosa:


stefanopolitano (approved):
Hi @fgrosa ! Thanks for the update. No comments from my side, approved!

PR #2015, opened 2025-05-30
Async QC trending example restricted to one period

knopers8:
Async QC trending example is restricted to one period to avoid excessive calls to QCDB.


knopers8:
@davidrohr , @shahor02 , @chiarazampolli could one of you review this PR please?

shahor02 (approved):

PR #2016, opened 2025-05-30
EMC ClustersTrend should publish only complete trends

knopers8:
EMC ClustersTrend configuration is changed to publish trends only once all input objects are processed, there is no need for intermediate versions.


knopers8:
@davidrohr , @shahor02 , @chiarazampolli could one of you review this PR please?

JianLIUhep:
Hi @catalinristea, @alcaliva, the EMC trending task is publishing a large number of objects to QCDB, and we are running out of space on the QCDB server. 

Could you please take this PR for all the upcoming productions? 

Thank you.

PR #2017, opened 2025-06-01
Fix mother -> daughter indices in case of replacement in HF gerator

fgrosa:
This PR fixes the history in case of particle replacement.
In particular, the indices of the mothers of the replaced particles are not updated correctly after `remove` and it is necessary in case of `includePartonEvent=false` (while not important if we keep the parton event, since the new particle -> mothers indices are always OK).


stefanopolitano (approved):
Hi @fgrosa ! Thanks for the development. No comments form my side. Approved!

PR #2034, opened 2025-06-17
[PWGHF] Add the `.ini` file for pp reference without charm baryon pdg list

wuctlby:
Hi @fgrosa , there is the `.ini` file for the pp reference without the charm baryon pdg list.  

Since I looked at the `High priority` in HF MC production list, there are three items: two for the charm baryon and one for D2H. I thought the D2H one is you were asking, but there is no dedicated `.ini` for this. So I added this file and followed the suggestion from @stefanopolitano, and tested this `.ini` file successfully.

Please let me know if I missed anything. Thanks a lot!


sawenzel:
@wuctlby : Please note that we are not backporting generator setting to older tags. You can simply set the env `O2DPG_MC_CONFIG_ROOT` variable to the correct O2DPG daily cvmfs tag.

wuctlby:
> @wuctlby : Please note that we are not backporting generator setting to older tags. You can simply set the env `O2DPG_MC_CONFIG_ROOT` variable to the correct O2DPG daily cvmfs tag.

Hi @sawenzel , thanks for the kind reminder; I will take note next time.

fgrosa (approved):
Hi @wuctlby thanks!
Indeed the `ini` file is necessary, however I did not get the comment about the baryons (it seems to me that they are - correctly - included). 
Beware also that the test for the pp ref is giving an error:
```
[0;31mError found in log /sw/BUILD/0c97f7df52b9503889f263351e3d908a26a7a7f9/O2DPG-sim-tests/o2dpg-sim_tests/o2dpg_tests/generators/0_GeneratorHF_D2H_ccbar_and_bbbar_gap5_Mode2_pp_ref.ini_External_dir/o2dpg-test-kine.log[0m
```
Did you test it locally?

PR #2037, opened 2025-06-19
Apply keys from generic config if applicable

sawenzel:
Improvement for configuration: Apply the key-values from the generic config on top of the anchorConfig, if the anchorConfig does not have these keys themselves. Useful for key-values in the digitization space for instance since they are not mentioned in async reco workflows.

In addition, actually apply FT0DigParam and FV0DigParam to the digitization of FIT. This didn't have any effect so far.


sawenzel:
Fixes an issue where FT0,FV0 were never applied in digitization despite intentions in this PR https://github.com/AliceO2Group/O2DPG/pull/1962

PR #2045, opened 2025-07-03
Force TPC common mode correction

wiechula:
This is done to avoid issues the CMk values stored in the CCDB. Data before 30/06/2023, 16:41:42 (first good run should be 539008 LHZ23ze) are affected

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -994,6 +994,9 @@ for tf in range(1, NTIMEFRAMES + 1):
 
    tpcLocalCF={"DigiParams.maxOrbitsToDigitize" : str(orbitsPerTF), "DigiParams.seed" : str(TFSEED)}
 
+   # force TPC common mode correction in all cases to avoid issues the CMk values stored in the CCDB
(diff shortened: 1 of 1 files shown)


MC/bin/o2dpg_sim_workflow.py
```diff
@@ -994,6 +994,9 @@ def putConfigValues(listOfMainKeys=[], localCF = {}):
 
    tpcLocalCF={"DigiParams.maxOrbitsToDigitize" : str(orbitsPerTF), "DigiParams.seed" : str(TFSEED)}
 
+   # force TPC common mode correction in all cases to avoid issues the CMk values stored in the CCDB
+   tpcLocalCF['TPCEleParam.DigiMode'] = str(2) # 2 = o2::tpc::DigitzationMode::ZeroSuppressionCMCorr from TPCBase/ParameterElectronics.h
```
sawenzel, line 998: so we just force this all the time, not just for some runs that are buggy?

wiechula: Well, for all other runs this is done by the default auto mode anyhow ...

sawenzel: Did someone test that it actually works and fixes our problems? Then I can merge.

alcaliva:
Hi @sawenzel,
if this PR is important to fix issues in the TPC digitizer, I think we should bump the current stable release for MC (O2PDPSuite::MC-prod-2025-v2-1). Shall we?

PR #2050, opened 2025-07-11
Included light collision systems in anchored workflow script generator

jackal1-66:


MC/bin/o2dpg_sim_workflow_anchored.py
```diff
@@ -258,21 +258,45 @@ def retrieve_MinBias_CTPScaler_Rate(ctpscaler, finaltime, trig_eff_arg, NBunches
...
+            "default": 0.8222
+        },
+        "Op": {
+            "default": 0.8222
+        },
+        "OO": {
+            "default": 0.8677
+        },
+        "NeNe": {
+            "default": 0.9147
```
sawenzel, line 277: @shahor02 @chiarazampolli : Hardcoding and maintaining these values in our script seems error-prone. Could we not have the trigger effs and related numbers stored in some GRP like CCDB object?

PR #2070, opened 2025-07-24
ITSMFT: rerun clusterizer

f3sch:
needs https://github.com/AliceO2Group/AliceO2/pull/14541


davidrohr:
This PR is correct in combination with the corresponding O2 PR. However we will use old O2 tags in production for a while, so I would give it 3-4 months of lead time until we adapt O2DPG.

PR #2076, opened 2025-07-29
Add option to have AO2Ds with prepropagated tracks

ddobrigk:
Uses https://github.com/AliceO2Group/AliceO2/pull/14544 to allow for tracks that have IB contributions to be pre-propagated at AO2D creation time. Tracks that only have OB contributions are kept as they are to allow for maximum flexibility when handling those while still taking advantage of the pre-propagation for most primary track analyses.

### DIFF
--- a/DATA/production/configurations/asyncReco/setenv_extra.sh
+++ b/DATA/production/configurations/asyncReco/setenv_extra.sh
@@ -803,7 +803,7 @@ if [[ $ALIEN_JDL_THINAODS == "1" ]] ; then
 fi
 
 if [[ $ALIEN_JDL_PREPROPAGATE == "1" ]] ; then
-  export ARGS_EXTRA_PROCESS_o2_aod_producer_workflow+=" --propagate-tracks 1 --propagate-tracks-max-xiu 5"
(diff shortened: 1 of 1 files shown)


DATA/production/configurations/asyncReco/setenv_extra.sh
```diff
@@ -802,6 +802,10 @@ if [[ $ALIEN_JDL_THINAODS == "1" ]] ; then
   export ARGS_EXTRA_PROCESS_o2_aod_producer_workflow+=" --thin-tracks"
 fi
 
+if [[ $ALIEN_JDL_PREPROPAGATE == "1" ]] ; then
+  export ARGS_EXTRA_PROCESS_o2_aod_producer_workflow+=" --propagate-tracks 1 --propagate-tracks-max-xiu 5"
```
shahor02, line 806: The `propagate-tracks` is a boolean switch, so its presence in the options turns it `on` regardless of the following argument, better remove `1` to avoid confusion if somebody wants to modify this option.

shahor02 (changes requested):

alcaliva (approved):

PR #2081, opened 2025-08-05
Adaptable QED events based on INTRATE

jackal1-66:
Changed number of QED events based on https://its.cern.ch/jira/browse/O2-5861

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -522,7 +522,8 @@ precollneeds=[GRP_TASK['name']]
 # max number of QED events simulated per timeframe.
 # A large pool of QED events (0.6*INTRATE) is needed to avoid repetition of events in the same or
 # neighbouring ITS readout frames, which would fire already activated pixel, discarding the event.
-NEventsQED = 10000 if (INTRATE <= 16668) else int(INTRATE*0.6)
+# Discussed in detail in https://its.cern.ch/jira/browse/O2-5861
+NEventsQED = max(10000, int(INTRATE*0.6))
 # Hadronic cross section values are taken from Glauber MC
 XSecSys = {'PbPb': 8., 'OO': 1.273, 'NeNe': 1.736}
 # QED cross section values were calculated with TEPEMGEN
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -519,7 +519,10 @@ vtxmode_sgngen = 'kCollContext'
 
 # preproduce the collision context / timeframe structure for all timeframes at once
 precollneeds=[GRP_TASK['name']]
-NEventsQED=10000  # max number of QED events to simulate per timeframe
+# max number of QED events simulated per timeframe.
+# A large pool of QED events (0.6*INTRATE) is needed to avoid repetition of events in the same or
+# neighbouring ITS readout frames, which would fire already activated pixel, discarding the event.
+NEventsQED = 10000 if (INTRATE <= 16668) else int(INTRATE*0.6)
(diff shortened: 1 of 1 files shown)


shahor02 (commented):
Thanks @jackal1-66 , I would say factor 0.3 is a strict minimum, could you change the factor to 0.6 and cap the `NEventsQED` to e.g. 10000?

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -519,7 +519,7 @@ def getDPL_global_options(bigshm=False, ccdbbackend=True):
 
 # preproduce the collision context / timeframe structure for all timeframes at once
 precollneeds=[GRP_TASK['name']]
-NEventsQED=10000  # max number of QED events to simulate per timeframe
+NEventsQED=int(INTRATE*0.3)  # max number of QED events to simulate per timeframe
```
sawenzel, line 522: So at 50kHz, this would be 15000 which seems manageable. 
Could we have an explanation added as comment where this comes from?

sawenzel:
> Thanks @jackal1-66 , I would say factor 0.3 is a strict minimum, could you change the factor to 0.6 and cap the `NEventsQED` to e.g. 10000?

@shahor02 : But if we cap to 10000, no improvement will be made. We are already at 10000 standard.

MC/bin/o2dpg_sim_workflow.py
jackal1-66: Is the comment enough? Otherwise I can go more in detail

jackal1-66:
> Thanks @jackal1-66 , I would say factor 0.3 is a strict minimum, could you change the factor to 0.6 and cap the `NEventsQED` to e.g. 10000?

Done and added the floor to 10000

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -519,7 +519,10 @@ def getDPL_global_options(bigshm=False, ccdbbackend=True):
 
 # preproduce the collision context / timeframe structure for all timeframes at once
 precollneeds=[GRP_TASK['name']]
-NEventsQED=10000  # max number of QED events to simulate per timeframe
+# max number of QED events simulated per timeframe.
+# A large pool of QED events (0.6*INTRATE) is needed to avoid repetition of events in the same or
+# neighbouring ITS readout frames, which would fire already activated pixel, discarding the event.
+NEventsQED = 10000 if (INTRATE <= 16668) else int(INTRATE*0.6)
```
sawenzel, line 525: I think that `NEventsQED = max(10000, int(INTRATE*0.6))` is clearer and avoids putting hard-coded constants. Also, maybe add link to JIRA ticket where this is discussed on more detail.

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -519,7 +519,11 @@ def getDPL_global_options(bigshm=False, ccdbbackend=True):
 
 # preproduce the collision context / timeframe structure for all timeframes at once
 precollneeds=[GRP_TASK['name']]
-NEventsQED=10000  # max number of QED events to simulate per timeframe
+# max number of QED events simulated per timeframe.
+# A large pool of QED events (0.6*INTRATE) is needed to avoid repetition of events in the same or
+# neighbouring ITS readout frames, which would fire already activated pixel, discarding the event.
+# Discussed in detail in https://its.cern.ch/jira/browse/O2-5861
+NEventsQED = max(10000, int(INTRATE*0.6))
```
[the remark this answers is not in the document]
jackal1-66: Ciao Ruben, it's floored no? If the interaction rate is above 16668 the QED events will be calculated from it, otherwise 1e4 will be used as value

PR #2086, opened 2025-08-06
Allow TPC residual extraction in MC and data corr maps in the reco

shahor02:
1) The TPC residuals extraction/aggregation is triggered by `ALIEN_JDL_DOTPCRESIDUALEXTRACTION=1` env.var, as in the standard cpass. The only difference wrt cpass processing is that the calibration slot duration for the `o2-tpc-scdcalib-interpolation-workflow` is set to 1s (`--sec-per-slot 1 `) to maximise the yield of tracks with residuals (virtually all tracks will be used).
In opposite, the `o2-calibration-residual-aggregator` uses `--sec-per-slot` as in the original anchored workflow.

The output is `o2tpc_residuals_<tsmin>_<tsmax>_<tfmin>_<tfmax>.root` files.

2) if `ALIEN_JDL_REMAPPINGS` is defined, it will be applied as `--condition-remap $ALIEN_JDL_REMAPPINGS` added to all workflows supporting this option.

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -153,6 +153,7 @@ parser.add_argument('--mft-assessment-full', action='store_true', help='enables 
 
 # TPC options
 parser.add_argument('--tpc-distortion-type', default=0, type=int, help='Simulate distortions in the TPC (0=no distortions, 1=distortions without scaling, 2=distortions with CTP scaling)')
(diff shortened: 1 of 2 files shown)


shahor02:
I've already tested it with the attached script, but want also to add/test a possibility to remap some CCDB objects to custom repositories (as for the planned test we will nee custom TPC correction maps).
[xxx.tar.gz](https://github.com/user-attachments/files/21663942/xxx.tar.gz)

PR #2090, opened 2025-08-11
Store CTP lumi in CTP digits in anchored sim.

shahor02:


shahor02:
Not tested yet, don't merge.

shahor02:
tested locally, CI fails due to running against previous O2 daily.

PR #2099, opened 2025-08-18
Update the injector to use rapidity instead of eta

BongHwi:
We're suspecting the strange injection results may come from the injection scheme we're using.

So we would like to try to produce some injected MCs from the local grid submission with the rapidity property.

Accordingly, the injection script has been introduced and updated all related files.


ktf:
No, I have not seen it. Also keep in mind that the new version of ROOT is not merged, so effectively nothing changed there.

jackal1-66:
CI is now green, it will be merged when someone from LF approves it

njacazio (approved):

PR #2106, opened 2025-08-28
[PWGHF] new files for production of electrons from heavy-flavour hadron decays

jpxrk:
First commit for production of electrons from heavy-flavour hadron decays.

### DIFF
--- a/MC/config/PWGHF/ini/GeneratorHF_HFe_ccbar_and_bbar_gap5_Mode2.ini
+++ b/MC/config/PWGHF/ini/GeneratorHF_HFe_ccbar_and_bbar_gap5_Mode2.ini
@@ -1,0 +1,8 @@
+### The external generator derives from GeneratorPythia8.
+[GeneratorExternal]
+fileName=${O2DPG_MC_CONFIG_ROOT}/MC/config/PWGHF/external/generator/generator_pythia8_gaptriggered_hf.C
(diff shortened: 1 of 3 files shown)


stefanopolitano:
Hi @jpxrk! Thanks a lot for your contribution. I don't have major comments, just a question for myself as I am not an HFe expert. By setting the `onIfAny` you enable all possible HF decays containing at least a e-,+ that might have a resonant state (like K*, phi, rho0, omega, etc.) for which you do not set any requirement on the decay as you do for eta and pi0, +. Is this intended? Thanks in advance for the clarification!

jpxrk:
Dear @stefanopolitano, I intended to keep only electrons from heavy-flavour hadron semileptonic decays and electrons from primary pi0 and eta decays. Thank you for poining out that I don't want electrons from intermediate resonances. I will modify the cfg file and push it again.

stefanopolitano:
Hi @jpxrk! Thanks! However, I still have a couple of questions. I thought the main motivation behind having the pi and eta was related to their large contribution as sources of background in your sample. Isn't that the case? Also, in the latest version of the config the decay are enabled only when electrons or neutrinos are present, but I thought you wanted to store decays with either electrons or positrons. Can you please clarify? Thanks a lot!

stefanopolitano:
Hi @jpxrk ! Thanks a lot for the clarification. Good for me!

PR #2113, opened 2025-09-05
Take out CPV, PHS, ZDC from workflow when they are not in the run

sawenzel:
This commit removes CPV, PHS and ZDC from the simulation workflow when these detectors are not part of the data taking.

Fixes https://its.cern.ch/jira/browse/O2-6145

Previusly we treated this scenario with zero-hit propagation, but this could still lead to noise-digits leading to some spurios AOD entries.


sawenzel:
@shahor02 : Could you please check if this is ok? I am sanitizing some detector lists getting from async-reco. PHS is still in there although it probably shouldn't.

PR #2129, opened 2025-09-24
Add Dzero meson in QC

fgrosa:
Hi @catalinristea this PR adds the D0 meson in the QC. I however noticed that the configuration for the other workflows is pretty old, how old is the O2Physics tag used in the QC? If very old, we might need to readjust a few configurations for the D0. Thanks!


catalinristea:
Hi! 
PR was tested and found fine for async reco Pb-Pb. It can be merged and put it in production. 
Thanks a lot, Fabrizio!

Yours, Catalin.

PR #2134, opened 2025-09-26
PWGHF: Smoothen number of injected signal events vs impact parameter for Pb-Pb collision MC

fchinu:
After discussing with @fgrosa we decided to replace the discontinuous function that determines the number of injected signal events to be embedded in a background Pb-Pb event with a smooth function. We also changed the asymptotic behaviour in peripheral events, increasing the number of injected events from 1 to 5, to increase the amount of signal in these collisions. 

Here you can find the comparison between the old (green) and new (red) function:
image

Tagging also @mfaggin for information


mfaggin:
ciao @fchinu , many thanks! It looks reasonable to me

PR #2141, opened 2025-09-30
TPC-time-series: dynamic track-types

sawenzel:


sawenzel:
@shahor02, @miranov25  : This should make the TPC time series task react on a disabled TOF detector. However, the workflow then actually crashes: It still searches for TOF. Could you maybe take a look when you have moment?

PR #2143, opened 2025-10-02
fix detectorList argument for Pb-Pb

f3sch:
In #2128, I introduced a bug where the detectorList argument did not have a separating space after it.
This leads to an incorrect construction of the cmd string `--detectorList ALICE2--configKeyValues` for the qed background sim. I only tested my changes in pp not in Pb-Pb...sorry


sawenzel:
No worries. We should improve the CI and test various different workflows.

PR #2324, opened 2026-04-16
Add Parameters for CMV Readout

rmunzer:


DATA/common/setenv_calib.sh
```diff
@@ -178,6 +178,9 @@ if [[ $CAN_DO_CALIB_TPC_SAC == 1 ]]; then
     fi
   fi
 fi
+if [[ -z ${CALIB_TPC_CMV} ]]; then
```
ehellbar: we don't allow unbound variables. That's why the CI crashes. Please use
```
if [[ -z ${CALIB_TPC_CMV+x} ]]; then
```

ehellbar (commented):
Please be aware that like this, the CMV workflows will be disabled by default. If this is what you want, please also make sure that this is the case for the FLP workflow, too. Can you confirm, please, that this is what you want? Then it's fine with me, modulo the one change requested. Also, make sure to enable it in the ECS config that you plan to use.

If you would like to enable it by default, please follow the logic that we use for CALIB_TPC_IDC.

DATA/common/setenv_calib.sh
rmunzer: Thanks @ehellbar,

I changed the variable. For the moment I would keep it disabled per default until we are sure, that it is running fine. Like this the change will be transparent.

PR #2326, opened 2026-04-16
Set proper settings for ITS vertexer for sim config

f3sch:
Sets now proper pp vertexer configs and removes non-existent option for Pb-Pb


shahor02 (approved):

shahor02:
couple to a change in the dpl-workflow.sh, merging

PR #2336, opened 2026-04-27
Update BTOPSIJPSITODIELECTRON.DEC to add chiC_1 and chiC_2 channels

shreyasiacharya:


shreyasiacharya:
adding new file for chiC

PR #2349, opened 2026-05-07
production.desc: add TPC CMV processing as calib collection

ehellbar:


ehellbar:
@rmunzer this is the last piece for the CMV processing. With this and https://github.com/AliceO2Group/AliceO2/pull/15370, it will automatically run on half of a calib node whenever you provide `CALIB_TPC_CMV=1` in pdp_extra_env_vars. Obviously, the workflow which runs on the FLPs should then match whatever we do on the EPNs.

PR #2355, opened 2026-05-11
TPC: add tpc-scalers workflow to standalone workflows

ehellbar:


ehellbar:
tested at P2, topology generation successful and QC plots look good. Merging.

PR #2358, opened 2026-05-19
workflow-multiplicities.sh: adjusting default ITS multiplicity cuts for PbPb

ehellbar:


ehellbar:
we set defaults twice, once here and once in https://github.com/AliceO2Group/AliceO2/blob/9ad6dd4b29fdf5749b138ac8bf4ea62c49405378/prodtests/full-system-test/dpl-workflow.sh#L179.

I changed it last year in dpl-workflow.sh, but just noticed that it is still overridden by workflow-multiplicities.sh.

PR #2370, opened 2026-06-02
AODBcRewriter: regroup tracks by collision so the -1 group stays contiguous

ktf:


ktf:
@jgrosseo @jackal1-66 @sawenzel would this make sense? I assume this is the same as sort + uniq if the two collections of BCs are one a subset of the other.

ktf:
I am actually not convinced this changes anything, actually...

ktf:
this one seems to produce a valid file...

PR #2378, opened 2026-06-10
[PWGHF] Add possibility to use EVTGEN in HF generator + config for B2Jpsi

fgrosa:
@fchinu this PR adds the config for the B->JPsi(->MuMu) + X MC using the EVTGEN decayer. Only the decays of B hadrons to charmonia are kept, and for the main ones the BR is multiplied x10 (to be discussed if we have to keep those with high mass state charmonia or not). The JPsi is forced to decay 70% of the times into MuMu and 30% of the times in EE, to evaluate residual contaminations. Can you please check the decays?

The test for the ini file has still to be added.


fchinu (approved):
Looks good to me thanks!

stefanopolitano (approved):

PR #2382, opened 2026-06-12
setenv_calib.sh: add new ErrorInfo ITS deadmap builder input to CALIBDATASPEC_BARREL_TF

ehellbar:


ehellbar:
the extra input is introduced here https://github.com/AliceO2Group/AliceO2/pull/15468, should be added to the proxies dataspec as well

PR #2393, opened 2026-06-30
Add diffraction-tuned pythia8 inel generator

ddobrigk:
This PR adds PYTHIA8 settings in which the single / double / non-diffractive components of inelastic pp events have their relative abundances tuned to match the corresponding fractions obtained previously from 7 TeV data. More details can be found in the JIRA ticket [PWDUD-65](https://its.cern.ch/jira/browse/PWGUD-65); thanks also to @romainschotter for testing this configuration prior to opening the PR and helping me fix some minor issues.

Once merged, it would be good to proceed to a full reconstruction test to calculate a revised trigger efficiency. Given the increased diffractive component, it's likely that the trigger efficiency will go down - if all goes well, it should be much closer to the one obtained in the pp 13.6 TeV vdM analysis.

### DIFF
--- a/MC/config/common/ini/tests/pythia8_inel_difftuned.C
+++ b/MC/config/common/ini/tests/pythia8_inel_difftuned.C
@@ -1,0 +1,24 @@
+int External() {
+  std::string path{"o2sim_Kine.root"};
+
+  TFile file(path.c_str(), "READ");
+  if (file.IsZombie()) {
+    std::cerr << "Cannot open ROOT file " << path << "\n";
+    return 1;
+  }
+
+  auto tree = (TTree *)file.Get("o2sim");
+  if (!tree) {
+    std::cerr << "Cannot find tree o2sim in file " << path << "\n";
+    return 1;
+  }
+  std::vector<o2::MCTrack> *tracks{};
+  tree->SetBranchAddress("MCTrack", &tracks);
+
+  auto nEvents = tree->GetEntries();
+  if (nEvents == 0) {
+    std::cerr << "No event of interest\n";
+    return 1;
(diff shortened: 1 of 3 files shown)


alcaliva (approved):

MC/config/common/ini/tests/pythia8_inel_difftuned.C
```diff
@@ -0,0 +1,24 @@
+int External() {
```
jackal1-66: Please replace `External` with `pythia8`, otherwise the test will not be fetched correctly in the CI.

jackal1-66 (changes requested):

MC/config/common/ini/tests/pythia8_inel_difftuned.C
ddobrigk: Thanks a lot for taking a look! Changed now!

MC/config/common/pythia8/generator/pythia8_inel_difftuned.cfg
```diff
@@ -0,0 +1,25 @@
...
+### decays
+ParticleDecays:limitTau0 on	
+ParticleDecays:tau0Max 10.	
+
+### adjust diffraction
+### note that PYTHIA calculates the ND sigma via a
+### 'remaining from total' formula: no need to put
+### that in by hand
+
+SigmaTot:mode 0
```
jackal1-66: This is probably `SigmaTotal`

jackal1-66 (changes requested):

jackal1-66 (approved):

PR #2398, opened 2026-07-02
Add triggerLc function call to makeStarlightConfig.py

Rrantu:
Added **if 'Lc' in args.process** in makeStarlightConfig.py


Rrantu:
+async-label async-2023-PbPb-apass5

mbroz84 (approved):

PR #2406, opened 2026-07-14
Quick bugfix for photon and jet hooks wrt to phi angle in acceptence.

smaff92:


smaff92:
Hi,

Apologies, it seems that the commit tree became a bit messy.... still should be fine though, this PR only changes things in the two files for the hooks.

One thing to flag; I did not actually test the changes in the jet hook. It compiles, and should in all probability still work, but it would be nice that the review just gives it a quick check as well, so I did not miss some obvious typo.

Photon hook was tested, and I can confirm that it now interacts correctly with the detector_acceptance.C code.

smaff92:
Another PR was opened here, with a cleaner commit history. 
https://github.com/AliceO2Group/O2DPG/pull/2408

PR #2409, opened 2026-07-15
Disable CTP lumi in TPC reco if not active

jackal1-66:
Just a possible easy solution to the issue reported in O2-7050, however more work is probably required.


sawenzel (approved):
I think this should do it according to the full-system-test setup.

PR #2416, opened 2026-07-23
Fix seg fault caused by stale particle indices after coalescence from HF decays

alcaliva:
Fix a segmentation fault in the coalescence afterburner by returning immediately after successful coalescence in a heavy-hadron decay. When coalescence occurs, one or more nucleons are removed from the event, invalidating the stored particle indices. If the loops continue, the code may attempt to access an invalid index and crash.

This PR proposes a minimal fix for the analysis of deuterons from Λb decays, but the same fix should probably be applied to all occurrences.

@fgrosa , @MRazza879

### DIFF
--- a/MC/config/common/external/generator/CoalescencePythia8.h
+++ b/MC/config/common/external/generator/CoalescencePythia8.h
@@ -152,6 +152,9 @@ bool CoalescencePythia8(Pythia8::Event& event, std::vector<unsigned int> inputPd
       for (int iN{0}; iN < neutrons[iC].size(); ++iN) {
         if (nuclearMask & (1 << kDeuteron)) {
           coalHappened |= doCoal(event, iC, pdgList[kDeuteron], massList[kDeuteron], trivialCoal, coalescenceRadius, nuclFromDecay, protons[iC][iP], neutrons[iC][iN]);
+          if (nuclFromDecay && coalHappened) {
+            return true;
+          }
         }
         if (nuclearMask & (1 << kTriton)) {
           for (int iN2{iN + 1}; iN2 < neutrons[iC].size(); ++iN2) {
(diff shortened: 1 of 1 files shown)


alcaliva:
Hi @sawenzel , @jackal1-66 , @fgrosa , 
did you have a look at this PR? Do you agree with the proposed change?

alcaliva:
I think I found the reason why the code crashes only for deuterons, while it works fine for ^3He. In the case of deuteron production, the same nucleon can satisfy the coalescence condition with more than one partner. Since, at every successful coalescence, doCoal() replaces one particle and removes the remaining nucleons from the event record, the precomputed particle index lists become invalid for the subsequent loops. I added a return immediately after a successful coalescence in the decay mode.

The standard event-wide coalescence workflow remains unchanged.

fgrosa:
> I think I found the reason why the code crashes only for deuterons, while it works fine for ^3He. In the case of deuteron production, the same nucleon can satisfy the coalescence condition with more than one partner. Since, at every successful coalescence, doCoal() replaces one particle and removes the remaining nucleons from the event record, the precomputed particle index lists become invalid for the subsequent loops. I added a return immediately after a successful coalescence in the decay mode.
> 
> The standard event-wide coalescence workflow remains unchanged.

Hi @alcaliva it looks good to me, but maybe @fmazzasc is the one who should have a look, since he implemented the coalescence for nuclei from decays

alcaliva:
> > I think I found the reason why the code crashes only for deuterons, while it works fine for ^3He. In the case of deuteron production, the same nucleon can satisfy the coalescence condition with more than one partner. Since, at every successful coalescence, doCoal() replaces one particle and removes the remaining nucleons from the event record, the precomputed particle index lists become invalid for the subsequent loops. I added a return immediately after a successful coalescence in the decay mode.
> > The standard event-wide coalescence workflow remains unchanged.
> 
> Hi @alcaliva it looks good to me, but maybe @fmazzasc is the one who should have a look, since he implemented the coalescence for nuclei from decays

Hi Fabrizo, thanks. Indeed, I've discussed also with @fmazzasc but maybe, could you have another look?

fmazzasc (approved):
Thanks @alcaliva for spotting the issue. The proposed fix looks good to me

sawenzel (approved):

PR #2418, opened 2026-07-27
AODBcRewriter : remapping matching indices in O2fwdtrack table

mcoquet642:
The reordering of O2fwdtrack table in AODBcRewriter Stage 1b does not remap the indices fIndexMFTTracks and fIndexFwdTracks_MatchMCHTrack. This PR defer writing O2fwdtrack and remap mathcing indices using MFT/Fwd PermMaps.


sawenzel:
Changes incorporated in larger fix. Thank you !

PR #2427, opened 2026-07-31
Pin reproducer CCDB queries to the original job's run time

sawenzel:
Reproducers resolved CCDB against the present, so any object re-uploaded since the job ran silently changed the outcome -- e.g. FT0/Calib/EventsPerBc, whose late arrival flips o2-steer-colcontexttool to the non-uniform sampler and gives a different collision context at the same fixed seed.

Inject ALICEO2_CCDB_CONDITION_NOT_AFTER, which BasicCCDBManager picks up, so it covers the anchoring stage as well as the workflow tasks.

Only pin when the moment is actually known: $2 if given, else "now" when run inside the job itself (the anchorMC.sh path). Otherwise skip and say so -- a wrong pin is worse than none. Skipped too if the JDL already sets it.

### DIFF
--- a/GRID/utils/getReproducerScript.sh
+++ b/GRID/utils/getReproducerScript.sh
@@ -1,6 +1,30 @@
 #!/bin/bash
 ALIEN_PID=$1
 
+# Epoch-ms the generated reproducer will pin its CCDB queries to (see the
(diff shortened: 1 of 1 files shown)


sawenzel:
Companion to https://github.com/AliceO2Group/AliceO2/pull/15653

PR #2435, opened 2026-08-20
tunable ratio of ccbar and bbbar for embedding

wuctlby:
Hi @xinyepeng , @gluparel , @stefanopolitano , @zhangbiao-phy , and @mfaggin ,

As discussed in the previous D2H meeting, I implemented a new generator with a tunable ratio of ccbar and bbbar for embedding the HF event into the signal event and the corresponding `ini` file, as well as the modification of `0.36000 0 2212 311 ### Λc+ -> p K0S 1.59%` ==> `1 0.36000 0 2212 310 ### Λc+ -> p K0S 1.59%` in the Pythia config file.

Specifically, 
- generator_pythia8_embed_hf.C
-- new `hf_generators` enum `GapHFRatio` to trigger `GeneratorPythia8GapTriggeredHF` with a given `quarkPdgList`
-- `BuildQuarkListFromBOverC`, which builds the `quarkPdgList` from `bOverCRatio`
-- `GeneratorPythia8EmbedHFRatio`, which accepts the new parameter `bOverCRatio` for generating the `quarkPdgList`. For example, if `bOverCRatio=3`, the `BuildQuarkListFromBOverC` will generate a list `[5,5,5,4]`, which will embed the ccbar and bbbar  in the order of this list

Please let me know if you have any comments or suggestions. Thanks a lot!

### DIFF
--- a/MC/config/PWGHF/external/generator/generator_pythia8_embed_hf.C
+++ b/MC/config/PWGHF/external/generator/generator_pythia8_embed_hf.C
@@ -18,6 +18,7 @@ namespace hf_generators
         GapTriggeredBeauty,         // --> GeneratorPythia8GapTriggeredBeauty: beauty enriched
         GapTriggeredCharmAndBeauty, // --> GeneratorPythia8GapTriggeredCharmAndBeauty: charm and beauty enriched (with same ratio)
         GapHF,                      // --> GeneratorPythia8GapHF
+        GapHFRatio,                 // --> GeneratorPythia8GapHF, quark list built from b/c ratio
         NGenType
     };
 }
@@ -82,6 +83,12 @@ public:
             LOG(info) << "********** [GeneratorPythia8EmbedHF] configuring GeneratorPythia8GapHF **********";
             LOG(info) << "**********                           Default number of HF signal events to be merged (updated by notifyEmbedding): " << mNumSigEvs;
             mGeneratorEvHF = dynamic_cast<GeneratorPythia8GapTriggeredHF*>(GeneratorPythia8GapTriggeredBeauty(/*no gap trigger*/1, yQuarkMin, yQuarkMax, yHadronMin, yHadronMax, hadronPdgList, partPdgToReplaceList, freqReplaceList));
+            break;
+
+        case hf_generators::GapHFRatio:
+            LOG(info) << "********** [GeneratorPythia8EmbedHF] configuring GapHFRatio (custom b/c ratio) **********";
+            LOG(info) << "**********                           Default number of HF signal events to be merged (updated by notifyEmbedding): " << mNumSigEvs;
+            mGeneratorEvHF = dynamic_cast<GeneratorPythia8GapTriggeredHF*>(GeneratorPythia8GapHF(/*no gap trigger*/1, yQuarkMin, yQuarkMax, yHadronMin, yHadronMax, quarkPdgList, hadronPdgList, partPdgToReplaceList, freqReplaceList));
             break;
         default:
             LOG(fatal) << "********** [GeneratorPythia8EmbedHF] bad configuration, fix it! **********";
@@ -387,6 +394,44 @@ private:
 
 };
 
+// Helper: build quarkPdgList from bOverCRatio (= N(beauty)/N(charm), default 1 = 1:1)
+// Integer b/c: taken directly; fractional b/c: reduced to nearest nB:nC with nC+nB <= 20
+static std::vector<int> BuildQuarkListFromBOverC(float bOverCRatio)
+{
+    if (bOverCRatio <= 0.f) { 
+        LOG(fatal) << "bOverCRatio (b/c) must be > 0";
+    }
+    if (bOverCRatio > 19.f) {
+        bOverCRatio = 19.f;
+        LOG(warn) << "bOverCRatio (b/c) too large, using 19:1";
+    }else if (bOverCRatio < 1.f/19.f) {
+        bOverCRatio = 1.f/19.f;
+        LOG(warn) << "bOverCRatio (b/c) too small, using 1:19";
+    }
+
+    int nC = 1, nB = 1;
+    float bestErr = 1e9f;
+    for (int c = 1; c <= 19; ++c) {
+        int b = (int)std::lround(bOverCRatio * c);
+        if (b < 1 || c + b > 20) {
+            continue;
+        } 
+        float err = std::fabs((float)b / c - bOverCRatio);
+        if (err < bestErr) {
+            bestErr = err;
+            nC = c;
+            nB = b;
+        }
+    }
+
+    std::vector<int> list;
+    // Bresenham interleaving
+    int len = nC + nB;
+    for (int k = 0; k < len; ++k)
+        list.push_back((((k + 1) * nC) / len > (k * nC) / len) ? 4 : 5);
+    return list;
+}
+
 // Charm enriched
 FairGenerator * GeneratorPythia8EmbedHFCharm(bool usePtHardBins = false, float yQuarkMin = -1.5, float yQuarkMax = 1.5, float yHadronMin = -1.5, float yHadronMax = 1.5, std::vector<int> quarkPdgList = {}, std::vector<int> hadronPdgList = {}, std::vector<std::array<int,2>> partPdgToReplaceList = {}, std::vector<float> freqReplaceList = {})
 {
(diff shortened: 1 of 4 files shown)


MC/config/PWGHF/external/generator/generator_pythia8_embed_hf.C
```diff
@@ -387,6 +394,44 @@ private:
...
+        if (b < 1 || c + b > 20) {
+            continue;
+        } 
+        float err = std::fabs((float)b / c - bOverCRatio);
+        if (err < bestErr) {
+            bestErr = err;
+            nC = c;
+            nB = b;
+        }
+    }
```
mfaggin:
```suggestion
const int iterNMax = 19; 
 if (bOverCRatio <= 0.f) { 
     LOG(fatal) << "bOverCRatio (b/c) must be > 0";
 }
 if (bOverCRatio > iterNMax*1.f) {
     bOverCRatio = iterNMax*1.f;
     LOG(warn) << "bOverCRatio (b/c) too large, using 19:1";
 }else if (bOverCRatio < 1.f/iterNMax) {
     bOverCRatio = 1.f/iterNMax;
     LOG(warn) << "bOverCRatio (b/c) too small, using 1:19";
 }

 int nC = 1, nB = 1;
 float bestErr = 1e9f;
 for (int c = 1; c <= iterNMax; ++c) {
     int b = static_cast<int>(std::lround(bOverCRatio * c));
     if (b < 1 || c + b > 20) {
         continue;
     }
     float err = std::fabs(static_cast<float>(b) / c - bOverCRatio);
     if (err < bestErr) {    
         /// This check here is needed in case bOverCRatio*c is not integer (it can happen with e.g. bOverCRatio=4./9.)
         /// In this case, b is its truncation and the desired ratio is not obtained
         /// It means that one needs to continue iterating until bOverCRatio*c is integer, i.e. b not truncated
         ///
         /// Possible cases:
         ///  1. we want nB = nC*R, with R integer
         ///     -> we enter here in the first loop iteration
         ///  2. we want more charm than beauty by an integer amount, i.e. nB = nC*R with R=1./N, with N integer
         ///     -> we enter here after N iterations, when c=N
         ///  3. we want either more charm or beauty, but with a factor R that is not integer, as well as its inverse (e.g. bOverCRatio=4./9.)
         ///     -> In this case, bOverCRatio*c is not integer, namely b is its truncation and the desired ratio is not obtained
         ///        The code iterates at most until c becomes equal to the denominator of the fraction
         ///  4. we want one of the previous cases, but we assign to bOverCRatio a value that is not rational
         ///     or such as we do not enter here within 19 iterations
         ///     -> nC and nB, are not touched, therefore we do not have the desired fraction. We'll need to throw a fatal (*)
         ///
         bestErr = err;
         nC = c;
         nB = b;

         /// If we are at this point, we reached already the desired ratio between b and c.
         /// Let's break the loop
         break;
     }
 }
 
 // (*) check if we have the desired fraction
 bool isRatioUnity = std::fabs(bOverCRatio-1) < 1e-05;
 if (!isRatioUnity && nC==1 && nB==1) {
  LOG(fatal) << "nC=" << nC << ", nB=" << nB << " but you ask bOverCRatio to be " << bOverCRatio <<", which is different from nB/nC. It means either that bOverCRatio is not rational, or that you need more than " << iterNMax << " iterations. Change it!";
 }
```

mfaggin: this approach here is nice, but in my opinion as written here it can be improved.

1. The loop can be stopped the first time we have `c` and `b` such that their ratio corresponds to the desired one, namely we need to `break;` the loop to avoid useless iterations.
In fact, with the current implementation always 19 iterations are done, even if only 1 is needed.

Your code gives what follows, assuming `bOverCRatio=3.f`:
```
iter
1          c=1, b=3   OK -> nC=1, nB=3    <-- we can stop here already
2          c=2, b=6   OK -> nC=2, nB=6
3          c=3, b=9   OK -> nC=3, nB=9
4          c=4, b=12  OK -> nC=4, nB=12
5          c=5, b=15  OK -> nC=5, nB=15   <-- this is the last time nC and nB are updated
6          c=6, b=18  NOT OK (c + b > 20)
7          c=7, b=21  NOT OK (c + b > 20)
.
.
.
19         c=19, b=57  NOT OK (c + b > 20)
```
After this loop you have then `nC=5` and `nB=15`, therefore you create a `vector` of 20 elements equal to `{5,5,5,4,5,5,5,4,5,5,5,4,5,5,5,4}`.
I propose to add a `break;` to solve this, avoiding the useless iterations and reducing in this case to have the `vector` `{5,5,5,4}`, which is enough for our purpose.

2. The code does not crash in case the the desired ratio is not guaranteed. This can happen in case as input one puts a value that is not rational, and for (very) particular values of this ratio that do not allow to converge within 19 iterations (I do not dislike the idea to have a max. number of iterations, to keep the running time limited). In this case we need the code to crash and ask the user to change the value, since it's very likely that the input value is not correct. I propose to add this check after the loop.

3. I propose to add some comments to make the code more readable

I propose these code changes, it would be good if you can give a few trials and let me know what you think

MC/config/PWGHF/pythia8/generator/pythia8_charmhadronic_with_decays_Mode2_hardQCD_5TeV.cfg
```diff
@@ -107,7 +107,7 @@ BeamRemnants:saturation 5
 4122:addChannel = 1 0.04500 100 2224 -321         ### Λc+ -> Delta++ K- 1.08%
 4122:addChannel = 1 0.09000 100 102134 211        ### Λc+ -> Lambda(1520) K- 2.20e-3
 ### Λc+ -> p K0S (36%)
-4122:addChannel = 1 0.36000 0 2212 311            ### Λc+ -> p K0S 1.59%
+4122:addChannel = 1 0.36000 0 2212 310            ### Λc+ -> p K0S 1.59%
```
mfaggin, line 110: In principle this should work in pythia, but did you check that this works?
The physics wants that only `4122:addChannel = 1 0.36000 0 2212 311` is correct, since K0, K0bar are the only states with a well-defined strangeness content, that materialize as K0s or K0L only in decay.
Did you check that the correct and expected decays are present, i.e. the fact that we have Lc decaying into p and K0s with K0s decaying in 2 pions?

mfaggin:
thanks @wuctlby , I left a couple of comments with a few suggestions and a question. Thank you!

MC/config/PWGHF/external/generator/generator_pythia8_embed_hf.C
wuctlby: Hi @mfaggin , thanks a lot for the comments. Initially, I was implementing the approach with any precision and any length quark pdg list. Then, I thought we would have several instances of `generator_pythia8_embed_hf` in production.  In this case, I was worried that the global event counter `mGeneratedEvents` in `gaptriggered_hf` could be reset, and we would embed only the `bbbar` in the peripheral events if the ratio is too large. So I implemented this approach and limited the quark pdg list.
1. But I don't think we would tune the ratio too much. So I simplified the approach and wrongly removed the `break`. Indeed, you are right. We should stop at the first case matching the requirements.
2. I further checked the logic of the process. It seems there is only one instance of `generator_pythia8_embed_hf` created. So, in the future we could further improve this approach to accept the ratio of  ccbar and bbbar with any precision,
3. Thanks!
I will directly apply your suggestion, it's good for me.

mfaggin: thanks Chuntai, I agree

MC/config/PWGHF/pythia8/generator/pythia8_charmhadronic_with_decays_Mode2_hardQCD_5TeV.cfg
wuctlby: Yes, I did a private test, and it works.
I checked the decay channels produced in one of the kinematic files, and they look good. 
ScreenShot_2026-08-21_184841_592
Then I further checked the decay of produced K0s, as follows:
Found K(S)0 daughters for Lambdac+ decay: 111 (#pi^{0}) 111 (#pi^{0}) 
Found K(S)0 daughters for Lambdac+ decay: 211 (#pi^{+}) -211 (#pi^{-})

wuctlby:
Hi @mfaggin and all, if there are no further comments, I would approve this PR first, @xinyepeng and @gluparel could you please approve this PR? I will then submit a new test with today's tag and open a JIRA for 2025 PbPb production if everything is okay. What do you think?

xinyepeng (approved):

PR #2442, opened 2026-08-25
Add a CCDB TimeMachine timestamp harvester for anchored MC

sawenzel:
This commit provides a new tool GRID/utils/getCCDBTimeMachineTimestamp.py that estimates a CCDB_TIMEMACHINE_TIMESTAMP for a given run and reco pass, for use as ALIEN_JDL_CCDB_CONDITION_NOT_AFTER in anchorMC.sh.

- Derives it from AliEn catalog timestamps of the pass's own per-timeframe reco output for that run
- Cross-checks the result against CCDB's own object history for a configurable calibration path.

Usage: GRID/utils/getCCDBTimeMachineTimestamp.py --year 2023 --period LHC23s --run 536683 --pass apass4

Related to https://its.cern.ch/jira/browse/O2-6543

### DIFF
--- a/GRID/utils/getCCDBTimeMachineTimestamp.py
+++ b/GRID/utils/getCCDBTimeMachineTimestamp.py
@@ -1,0 +1,175 @@
+#!/usr/bin/env python3
+
+# Estimates a CCDB TimeMachine "not-after" timestamp for a given run + reco pass,
+# for use as ALIEN_JDL_CCDB_CONDITION_NOT_AFTER in anchored MC.
+#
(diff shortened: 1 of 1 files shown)


shahor02 (commented):
Looks good! 
A possible caveat: there are a handful of CCDB objects used only for the MC (like TPC distortions/corrections, MCH masks etc.) which are usually prepared quite late and, if used with the time machine, should be uploaded with fake creation time.

Perhaps we should add an option to pass a list of such objects and check/report their presence for the selected timestamp.

sawenzel:
@shahor02 : Good point ! Thx.

PR #2445, opened 2026-08-28
o2dpg_sim_workflow.py: only add reco tasks of smaller detectors if they are active

ehellbar:


ehellbar:
this fixes the second issue in https://its.cern.ch/jira/browse/O2-7142?focusedId=7427580&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-7427580

sawenzel (approved):

ehellbar:
for some of the matching tasks, the new logic might not be correct, still checking

MC/bin/o2dpg_sim_workflow.py
```diff
@@ -1658,7 +1663,7 @@ def getDigiTaskName(det):
 
    #<--------- MFT-MCH forward matching
    forwardmatchneeds = [MCHRECOtask['name'],
-                        MFTRECOtask['name'] if isActive("MFT") else None,
+                        MFTRECOtask['name'],
```
ehellbar, line 1666: changed the MFT reco task to an unconditional requirement for the forward matching task

PR #2449, opened 2026-09-02
Keep the ZEM geometry instead of skipping the whole ZDC module

sawenzel:
This changes the simulation default from skipping the ZDC module to dropping only its +-113 m beam line, so the ZEM material stays in the transport geometry.

- The ZDC geometry is the +-113 m beam line, its magnets, ZN, ZP and ZEM; the transport cost is the beam line, while ZEM sits at z = 7.6 m.
- ZEM carries 217 kg of material between z = 745 and 862 cm, spanning eta 2.8 to 5.6, which is inside the FIT and FDD acceptance.
- --skipModules no longer defaults to ZDC; create_geant_config() sets ZDCSimParam.buildBeamLine=false instead, unless --with-ZDC is given.
- An explicit --skipModules ZDC still removes the module altogether.
- ZDC stays out of digitisation and reconstruction as before, which the existing deactivate_detector('ZDC') already handles.
- The MODULES variable removed from the two anchor scripts was never referenced; it read as if anchored MC opted out of ZDC, which the workflow default decided.
- Requires ZDCSimParam.buildBeamLine in AliceO2.

https://its.cern.ch/jira/browse/O2-7158

### DIFF
--- a/MC/bin/o2dpg_sim_config.py
+++ b/MC/bin/o2dpg_sim_config.py
@@ -183,6 +183,12 @@ def create_geant_config(args, externalConfigString):
     # ----- add default settings -----
 
     add(config, {"MFTBase.buildAlignment" : "true"})
+
+    # Keep the ZEM calorimeters, drop the +-113 m beam line that makes the ZDC
+    # expensive to transport. --with-ZDC asks for the whole thing, and an explicit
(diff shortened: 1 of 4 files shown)


sawenzel:
Needs https://github.com/AliceO2Group/AliceO2/pull/15752 be merged beforehand

PR #2450, opened 2026-09-03
add different gap configs for 13 TeV

jaimenorman:


marcovanleeuwen:
Hi @nzardosh,

Could you have a look at this pull request? It is needed for an MC production: https://its.cern.ch/jira/browse/O2-7162 .

Once it is merged, we should also ask for a tag. Do we do that here as well, or separately?

Marco.

nzardosh (approved):

PR #2451, opened 2026-09-03
Adapt o2dpg_sim_workflow.py to new TPC corr.maps

shahor02:
Given that the new maps are provided by the tpc-scaler device and the latter is added in the MC by the o2-tpc-reco workflow, the corrections and CTP input request should be disable when the o2-tpc-reco is used for clusterization only (correction not used). In general, all settings of o2-tpc-scaler from the anchoring workflowconfig.log should be added to options of all MC reco workflows requiring TPC corrections (except the MC-specific settings like imposing --lumi-type 1 --corrmap-lumi-mode 2), but since at the moment MC does not simulate sector edge fluctuations, we simply add --disable-sec-edge-fluc-correction.

### DIFF
--- a/MC/bin/o2dpg_sim_workflow.py
+++ b/MC/bin/o2dpg_sim_workflow.py
@@ -1280,6 +1280,7 @@ for tf in range(1, NTIMEFRAMES + 1):
    if not args.combine_tpc_clusterization:
      # We treat TPC clusterization in multiple (sector) steps in order to
      # stay within the memory limit or to parallelize over sector from outside (not yet supported within cluster algo)
+     # For the clusterization we disable the TPC SC corrections
      tpcclustertasks=[]
      sectorpertask=18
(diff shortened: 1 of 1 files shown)


sawenzel:
I verified the clusterization part on `async-2026-PbPb-apass1-v7`: I reproduce the
`No matching output found for CTP/LUMI/0` fatal there, and `--corrmap-lumi-mode 3
--disable-ctp-lumi-request` fix it, with no effect on the older tags I tried. LGTM for that.

I see that no async tag supports `--disable-sec-edge-fluc-correction` yet, so we need to
query its availability via `option_if_available`:

```python
tpc_corr_scaling_options += option_if_available('o2-tpc-reco-workflow', '--disable-sec-edge-fluc-correction', envfile=async_envfile)
```

shahor02:
Thanks for spotting, indeed, I was testing with O2 dev. Added `option_if_available...`

PR #2452, opened 2026-09-03
[PWGGAJE] jet param model: add bkg-only, fix eta

aimeric-landou:


aimeric-landou:
How many events does a test run through? the two tests that fail are testing the _Sgl generators, that can have events with 0 particles quite often, so if the number of events generated for the test is small, the test that checks that at least one particle has been produced might fail.

nzardosh (approved):

nzardosh:
Its approved. Do you want to tag it with any of the reconstruction passes above before I merge it?

aimeric-landou:
Hi @nzardosh, this one is a generator-only MC that is not anchored to any dataset, so no need to tag it with a reconstruction pass.

PR #2461, opened 2026-09-14
Added runtime user hook for triggered events to LambdaB gen

jackal1-66:
GeneratorHFTrigger_LambdaBToDeuteron.ini ran usually for >24h on
the GRID and got killed on TTL.
`generator_pythia8_hfhadron_to_nuclei.C` trigger branch retries Pythia8
event generation indefinitely until the requested b-hadron turns up at midrapidity. The underlying .cfg
uses plain SoftQCD:inelastic with no production bias, so on unbiased
minimum-bias QCD the number of full hadronizations needed per trigger
event was measured at mean ~800 retries, up to 2928 observed in a 10-event sample.
A production job needs thousands of trigger events, so any single one
drawing a bad statistical fluctuation can stall the job for hours with
no error and no log output - looking indistinguishable from a hang.
So, this solution biases parton-level generation towards bbbar production at midrapidity
via a Pythia8 UserHooks veto (reusing UserHooks_qqbar, already used by
the standard beauty gap-triggered HF generator), so non-bbbar attempts
are rejected before the hadronization step runs. This drops
the mean retries needed per trigger event by ~2 orders of magnitude.
The bias is  toggled via the new UserHooks_qqbar::setActive() around generateEvent()'s retry loop, since it
must not run during the minimum-bias branch, but only for the triggered events.
The issue was spotted in https://its.cern.ch/jira/browse/O2-7169


jackal1-66:
Tagging @MRazza879 for a review on this as well

alcaliva (approved):

PR #2464, opened 2026-09-16
Add DittoMC

njacazio:
Needs https://github.com/alisw/alidist/pull/6333


MC/config/ALICE3/ini/ditto_pp_136tev.ini
```diff
@@ -0,0 +1,6 @@
+[Diamond]
+width[2]=6.0
```
sawenzel: do you need this vertex setting? I would prefer not having as part of generator config or is it tightly coupled to this generator?

njacazio: Not needed! Removed!

sawenzel:
@njacazio : It appears that there are some "relocation problems". I see hard-coded header include paths: `#include "/local/workspace/DailyBuilds/DailyO2Physics-slc9/daily-tags.FGEPWk1CYf/SOURCES/DittoMC/v1.0.0/v1.0.0/Tuning/DittoTune.h"` which survive in the cvmfs installation. Most likely, we need to go one step back and fix the Ditto sources.

njacazio:
Hi @sawenzel indeed, thanks for the suggestion! I verified locally via lxplus and added a new release with the fix that we can import with https://github.com/alisw/alidist/pull/6343

